PR #14, opened 2017-05-01
Run all BUILD/skylark files through buildifier

ixdy:
I'm working on adding some functionality to the rules, and my vim auto-runs buildifier, which was causing a bunch of unrelated diffs.

This change is the effect of running `github.com/bazelbuild/buildifier` on all `BUILD` and `bzl` files in the repo. It should otherwise be a no-op.

(It'd be nice to enable buildifier in travis or something, but I'm not sure how to get it installed there.)


mattmoor:
Yeah, I'd love a buildifier presubmit check. :-/

Issue #3, opened 2017-04-12
support build stamp with docker_push


mikedanese:
I should be able to use a variable from my workspace status to tag a docker image when running docker_push.

mattmoor:
`docker_push` supports make-style expansion for the components:

```
vardef("MY_TAG", "latest")

docker_push(
    name = "foo_push",
    image = ":foo_image",
    registry = "gcr.io",
    repository = "foo/bar/baz",
    tag = "$(MY_TAG)",
)
```

Or is that not what you want?

mattmoor:
@damienmg How can I access [this](https://github.com/kubernetes/kubernetes/blob/master/.bazelrc#L6) in skylark?

mattmoor:
Ok, blocked by: https://github.com/bazelbuild/bazel/issues/1054

ixdy:
Similar / clarifying request: I'd like to be able to tag an image automatically using the git sha of the workspace. We use `--workspace_status_command` to set this for go rules, but there's currently no way to use that in docker rules.

Make-style expansion isn't ideal, since that requires adding `--define`s to the bazel command line.

mattmoor:
In other places we allow `@foo` to load the value from a file.  I wonder if this would be sufficient for your purposes?

ixdy:
That might work. I implemented a hack (https://github.com/ixdy/bazel/commit/3b29803eb528ff525c7024190ffbf4b08c598cf2) to do that a few months ago, and it's what the kubernetes bazel build is currently using, though it still feels pretty hacky

(I got preoccupied with other things and didn't follow up on some review feedback, and now things have moved around a bit so I'd need to rebase my change.)

The go rules got around waiting for Skylark support (https://github.com/bazelbuild/bazel/issues/1054) by implementing stamping through reading the `ctx.info_file` and `ctx.version_file` directly.

Since IIRC most of the actual docker image/layer-building implementation happens in python, I wonder if we could do something similar and have the info/version files passed through to the scripts on `stamp=1`? 
I haven't thought through all of the implications yet, or if the rules will work correctly with unexpanded variables in skylark.

mikedanese:
Is this fixed?

mattmoor:
`docker_push` doesn't support it, but `docker_bundle` does.

PR #17, opened 2017-05-01
Add stamp option to docker_bundle rule

ixdy:
This allows using workspace status information (such as git describe) in image tags. To differentiate from make variables (and laziness), I'm using python formatting syntax, e.g. `{STABLE_BUILD_USER}` or `{MY_CUSTOM_VAR}`.

This is basically an implementation of what I was describing in https://github.com/bazelbuild/rules_docker/issues/3#issuecomment-297563224.

I think it'd work for what I need for kubernetes right now, though I'm not sure if this is the correct implementation, e.g. does it makes sense to do the stamping in `docker_bundle`? Should it also be supported in `docker_build` and/or `docker_push`? What things should be "stampable"? Do we want to be able to embed stamping info in image labels? (I might in the future, e.g. embed revision/other build data.)

(`docker_push` actually looks harder, since it's basically just a shell script template wrapper to the code in containerregistry.)

The proper way to do all of this would be entirely in skylark, though that depends on  https://github.com/bazelbuild/bazel/issues/1054 being resolved.

### DIFF
--- a/docker/join_layers.py
+++ b/docker/join_layers.py
@@ -43,6 +43,9 @@ parser.add_argument('--tags', action='append', required=True,
                           'and the layer they tag. '
                           'e.g. ubuntu=deadbeef,gcr.io/blah/debian=baadf00d'))
 
+parser.add_argument('--stamp-info-file', action='append', required=False,
+                    help=('If stamping these layers, the list of files from '
+                          'which to obtain workspace information'))
 
 def create_image(output, layers, tag_to_layer=None, layer_to_tags=None):
   """Creates a Docker image from a list of layers.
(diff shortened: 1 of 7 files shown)


docker/join_layers.py
```diff
@@ -87,13 +90,24 @@ def main():
...
   tag_to_layer = {}
   layer_to_tags = {}
+  stamp_info = {}
+
+  if args.stamp_info_file:
+    for infofile in args.stamp_info_file:
+      with open(infofile) as info:
+        for line in info:
+          line = line.strip("\n")
+          key, value = line.split(" ")
```
mattmoor: you probably want `line.split(" ", 1)` (or maybe 2, I can never remember the sense of this)

docker/join_layers.py
```diff
@@ -87,13 +90,24 @@ def main():
...
+          stamp_info[key] = value
```
mattmoor: Can you print a warning when `key` exists in the map?

ixdy:
Thanks for the quick review! I've addressed your comments in a follow-up commit and can squash if desired.

mattmoor:
GitHub will let me squash and merge, but I'm happy to leave the history there.  thanks for the change.

PR #22, opened 2017-05-05
Remove verbosity on tar in build_{oci_}test.sh

mattmoor:
This verbosity leads to additional complexity to strip it out and is (I believe) causing these tests to fail on OSX because the output is differently shaped.


damienmg:
test this please

mattmoor:
Ok, with this all of the builds are green :)

PR #30, opened 2017-05-13
Add a rule for generating passwd files.

dlorenc:
Ref https://github.com/GoogleCloudPlatform/distroless/pull/16


bazel-io:

docker/build_test.sh
```diff
@@ -547,6 +547,13 @@ function test_build_with_tag() {
     "[\"gcr.io/build/with:tag\"]"
 }
 
+function test_build_with_passwd() {
+  # We should have a tag in our manifest containing the name
+  # specified via the tag kwarg.
+  check_layers "with_passwd" \
+    "621fe329a78d65d90d34d6dc277ccac2249bba4c8228222271418bf6b07c4dec"
+}
```
mattmoor: You should add an additional check that uses `get_layer_listing` to check that the `passwd` file is in the new layer.

mattmoor:
test this please

damienmg:
I would suggest we move the documentation to use https://github.com/bazelbuild/skydoc and we generate the doc on the gh-pages. That way we would cover the contrib part too. Pulling in @davidzchen about that.

PR #36, opened 2017-05-24
Switch sample to use distroless base.

mattmoor:


mattmoor:
@dlorenc PTAL

README.md
```diff
@@ -50,8 +50,7 @@ docker_repositories()
 docker_pull(
   name = "java_base",
   registry = "gcr.io",
-  # TODO(mattmoor): Replace this.
-  repository = "google-appengine/java",
+  repository = "distroless/java",
   # 'tag' is also supported, but digest is encouraged for reproducibility.
   digest = "sha256:deadbeef",
```
dlorenc: Want to use a real digest here?

mattmoor: No, because people copy and paste :)

pcj: Can we leave a note as to the recommended way to obtain the sha256 of the pulled image?

mattmoor: You should be able to put `gcr.io/foo/bar:blah` into your browser, and then clicking "Show Pull Command" will contain the digest.

pcj: Thanks.  Attaching screenshot for someone in the future.  Might be nice to include this info in the README.

screen shot 2017-08-26 at 3 58 25 pm

Issue #40, opened 2017-06-02
bazel explodes on multiple docker_build rules in same package with different external base images


ixdy:
Simple example given in https://github.com/ixdy/bazel-rules_docker/commit/8125a4cb8e74a8e7785aa4235df3a80389da2f7c.

I have two `docker_build` rules in the same BUILD file, each one depending on an different external image (pulled in through `docker_pull` in my `WORKSPACE`).

When I try to build both, bazel explodes, I guess because they are both trying to create the same artifact, `image.tar.id`:
```console
$ bazel build foo.tar bar.tar
.
ERROR: file 'image.tar.id' is generated by these conflicting actions:
Label: //:foo, //:bar
RuleClass: docker_build_ rule
Configuration: 32a994dd81b79c0621b68fa6003f0e25
Mnemonic: ExtractID
Action key: c0d8c6c6b75540d6052f398dcbf6fc31, fdc7414328621a989a7e2ce8c931afc8
Progress message: ExtractID image.tar.id
PrimaryInput: File:[/usr/local/google/home/jgrafton/.cache/bazel/_bazel_jgrafton/1c344542258c3021719db9a8e772b13f[source]]external/official_python/image/image.tar, File:[/usr/local/google/home/jgrafton/.cache/baz
el/_bazel_jgrafton/1c344542258c3021719db9a8e772b13f[source]]external/official_busybox/image/image.tar
PrimaryOutput: File:[[/usr/local/google/home/jgrafton/.cache/bazel/_bazel_jgrafton/1c344542258c3021719db9a8e772b13f/execroot/rules_docker]bazel-out/local-fastbuild/bin]image.tar.id
.
ERROR: Analysis of target '//:foo' failed; build aborted.
INFO: Elapsed time: 35.226s
```

mattmoor:
Thanks for the repro, I will try to TAL tomorrow.

@damienmg FYI, and in case he has any thoughts on this.

mattmoor:
@damienmg It's essentially [this](https://github.com/bazelbuild/rules_docker/blob/52d9faf209ff6d16eb850b6b66d03483735e0633/docker/layers.bzl#L24) I think.  LMK if there is something better we should name these...

damienmg:
Right, question: why docker_pull does not do a docker_build we can depend on?

Anyway the underlying problem can be fixed by  using https://www.bazel.build/versions/master/docs/skylark/lib/ctx.html#new_file(sibling_file, basename)

[...]

PR #44, opened 2017-06-05
Incorporate label name into the ID field when extracting base image ids.

mattmoor:
Bazel puts the base's id/name files alongside the consumer, not the producer, so multiple base images with identical basename portions will conflict with each other.  Using the `sibling_file` first argument to `ctx.new_file` doesn't work because these can straddle package boundaries, which Bazel doesn't like.

Instead, we simply incorporate our label's name as a prefix of the new output file.

Fixes https://github.com/bazelbuild/rules_docker/issues/40


mattmoor:
@ixdy If you want to try this out and review in @damienmg absence so I can resolve your issue.

mattmoor:
Wanted to bump this, since I've started to run into merge conflicts.

WORKSPACE
```diff
@@ -13,8 +13,21 @@
...
+    name = "distroless_base",
+    registry = "gcr.io",
+    repository = "distroless/base",
+)
+
+docker_pull(
+    name = "distroless_cc",
+    registry = "gcr.io",
+    repository = "distroless/cc",
+)
```
damienmg: nit: add newline at the end of the file

ixdy:
apologies for not following up on this. thanks for the fix!

Issue #35, opened 2017-05-24
bazel run docker_build doesn't work when using stamps


mikedanese:
cc @ixdy 

repro

```console
$ bazel run //build:kube-controller-manager                                                                                                                                                                                                                                                           
INFO: Found 1 target...
Target //build:kube-controller-manager up-to-date (nothing to build)
INFO: Elapsed time: 7.461s, Critical Path: 5.67s

INFO: Running command line: bazel-bin/build/kube-controller-manager
Loading d10f0ce649a0386cfc6df629ba86375210548a6d4cf62ca52e722213f4d95eda...
Loaded image: bazel/build:busybox
Loading 9bd778a641e0bbd2a59d3a0d11cfa670aeb0b73800db6321491ada6022f86519...
Loaded image: bazel/build:kube-controller-manager-internal
Tagging 9bd778a641e0bbd2a59d3a0d11cfa670aeb0b73800db6321491ada6022f86519 as gcr.io/google_containers/kube-controller-manager:{STABLE_DOCKER_TAG}
Error parsing reference: "gcr.io/google_containers/kube-controller-manager:{STABLE_DOCKER_TAG}" is not a valid repository/tag
```

mattmoor:
I think for `run` we want to resolve things [here](https://github.com/bazelbuild/rules_docker/blob/master/docker/bundle.bzl#L53)

mattmoor:
I really don't want to pass this into the incremental load template and write a version of [this](https://github.com/bazelbuild/rules_docker/blob/master/docker/join_layers.py#L131-L140) in bash...  @damienmg ideas?

mattmoor:
Just to write it down, you can reproduce this with the checked in unit test:
```
bazel run docker/testdata:stamped_bundle_test
```

mikedanese:
@ixdy recently wrote a version of this in bash. It would be great to standardize this across rules (maybe a binary in tools that handled this).

mattmoor:
@ixdy Do you have a pointer?

ixdy:
yes, https://github.com/bazelbuild/rules_go/pull/505 is where I implemented similar functionality for rules_go using skylark + shell utilities.

PR #45, opened 2017-06-05
Support stamp variable expansion with 'bazel run'

mattmoor:
Fixes https://github.com/bazelbuild/rules_docker/issues/35

### DIFF
--- a/docker/incremental_load.sh.tpl
+++ b/docker/incremental_load.sh.tpl
@@ -72,6 +72,21 @@ function tag_layer() {
   fi
 }
 
+function read_variables() {
+  local file=${RUNFILES}/$1
+  local new_file=$(mktemp)
+  trap "rm -f ${new_file}" EXIT
+
+  # Rewrite the file from Bazel for the form FOO=...
+  # to a form suitable for sourcing into bash to expose
+  # these variables as substitutions in the tag statements.
+  sed -E "s/^([^ ]+) (.*)\$/export \\1='\\2'/g" < ${file} > ${new_file}
+  source ${new_file}
+}
+
+# Statements initializing stamp variables.
+%{stamp_statements}
+
 # List of 'incr_load' statements for all layers.
 # This generated and injected by docker_build.
 %{load_statements}
--- a/docker/BUILD
+++ b/docker/BUILD
@@ -57,6 +57,7 @@ TEST_DATA = [
     # "//docker/testdata:extras_with_deb.tar",
     "//docker/testdata:bundle_test.tar",
     "//docker/testdata:stamped_bundle_test.tar",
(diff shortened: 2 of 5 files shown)


mattmoor:
@mikedanese @ixdy FYI

mattmoor:
@mikedanese @ixdy Can one of you TAL at this PR?  IIUC @damienmg is OOTO for the next month, and I believe you both have familiarity with the pieces here.

mattmoor:
Wanted to bump this, since I've started to run into merge conflicts.

docker/incremental_load.sh.tpl
```diff
@@ -72,6 +72,17 @@ function tag_layer() {
   fi
 }
 
+function read_variables() {
+  local file=${RUNFILES}/$1
+  local new_file=$(mktemp)
```
damienmg: Delete the file at exit?

docker/incremental_load.sh.tpl
```diff
@@ -72,6 +72,17 @@ function tag_layer() {
...
+
+  sed -E "s/^([^ ]+) (.*)\$/export \\1='\\2'/g" < ${file} > ${new_file}
```
damienmg: Instead of that tricky sed, you could read the file and emit export.
```bash
while IFS='' read -r line || [[ -n "$line" ]]; do
    export "${line%% *}=${line#* }"
done < "${file}"
```

See https://stackoverflow.com/questions/10929453/read-a-file-line-by-line-assigning-the-value-to-a-variable

mattmoor: Not sure either of these is better than adding a comment, we'd be trading "tricky sed" for "tricky bash".  Your suggestion is also buggy, since it doesn't quote the RHS. :)

I'm adding a descriptive comment.

mattmoor:
Should be RFAL.

docker/incremental_load.sh.tpl
```diff
@@ -72,6 +72,21 @@ function tag_layer() {
...
+function read_variables() {
+  local file=${RUNFILES}/$1
+  local new_file=$(mktemp)
+
+  # Rewrite the file from Bazel for the form FOO=...
+  # to a form suitable for sourcing into bash to expose
+  # these variables as substitutions in the tag statements.
+  sed -E "s/^([^ ]+) (.*)\$/export \\1='\\2'/g" < ${file} > ${new_file}
+  source ${new_file}
+  rm ${new_file}
```
damienmg: Add a trap instead so a failure remove the file?

`trap "rm -f ${new_file}" EXIT`

(side-note: the bash magic avoids the creation of a temporary file, that's why I proposed it).

Issue #54, opened 2017-06-16
Stamp for docker_push?


samuraisam:
In my docker/BUILD file I have something like this to create docker images. Currently this requires `bazel clean` between builds for these services because I haven't figured out how to get the stamps to update every run (which is besides the point, there must be a way).

```python
package(default_visibility = ["//visibility:public"])

load("@io_bazel_rules_docker//docker:docker.bzl", "docker_bundle", "docker_push")

DOCKER_BUILDS = {
    "assets":       "//assets/assets:docker", # etc...
}

[docker_bundle(
    name = binary,
    images = {"gcr.io/{GCLOUD_PROJECT_ID}/%s:{BRANCH_NAME}-{BUILD_NUMBER}" % binary: docker_image},
    stamp = True
 ) for binary, docker_image in DOCKER_BUILDS.items()]
```

My desire is to define matching docker_push targets:

```python
[docker_push(
    name = "push_" + binary,
    image = ":" + binary,
    registry = "gcr.io",
    repository = "{GCLOUD_PROJECT_ID}/%s" % binary,
    tag = "{BRANCH_NAME}-{BUILD_NUMBER}",
    stamp = true,
 ) for binary in DOCKER_BUILDS.keys()]
```

However this doesn't appear to be supported. I assume there is a reason why that is escaping me. Or maybe I should just be doing this in an external tool.

mattmoor:
This is the final bit remaining here: https://github.com/bazelbuild/rules_docker/issues/3

I just need to incorporate a trick like what @ixdy did for docker_bundle into the `docker_pusher` tool.

I'll investigate whether this is as simple as I think today, and post back if there are any unforeseen problems.

mattmoor:
@samuraisam Hopefully my PR addresses what you're looking for?

samuraisam:
Based on the test in your PR this looks like it would work for my use case above other than still needing to hardcode the repository (eg `{GCLOUD_PROJECT_ID}`) but maybe that could be incorporated into the tag?

mattmoor:
@ixdy is a bit more familiar than I am with manipulating the workspace status stuff, so perhaps he has a suggestion?

ixdy:
the test in the PR didn't cover it, but I just tested manually where I used `{BUILD_USER}` in the `repository` and `{STABLE_DOCKER_TAG}` (a custom value) in the `tag` and it worked as expected.

samuraisam:
[...]

PR #56, opened 2017-06-16
Add support for stamping docker_push

mattmoor:
Fixes: #3
Fixes: #54

### DIFF
--- a/docker/testdata/BUILD
+++ b/docker/testdata/BUILD
@@ -13,7 +13,12 @@
 # limitations under the License.
 package(default_visibility = ["//docker:__subpackages__"])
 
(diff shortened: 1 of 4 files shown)


docker/testdata/BUILD
```diff
@@ -246,6 +251,15 @@ docker_build(
     },
 )
 
+docker_push(
+    name = "push_stamp",
+    image = ":link_with_files_base",
+    registry = "gcr.io",
+    repository = "convoy-adapter/bazel-test",
```
ixdy: @samuraisam's note from #54: it might be worth putting a stamped value in here too. (it seems to work, but it'd be good to test.)

PR #70, opened 2017-06-26
This implements a dramatic simplification of docker_build

mattmoor:
In addition to simplifying things, this will dramatically improve their performance, which previously entailed a lot of superfluous seeking through tarballs, g[un]zipping, and converting between various Docker image formats.

On several pathological internal images, this improves the build times 4x or more.  The only docker_* action remaining on their critical path is now ExtractConfig on their large checked-in base image tarball.

This commit only changes the behavior of `docker_build`, and it doesn't clean up a lot of the vestigial utilities (e.g. extract_id, create_image).  A subsequent change will clean these up.  After that, I will spend some time enabling `docker_pull` to produce the efficient representation and `docker_push` to consume it (vs. going through the linked tarball).


mattmoor:
@dlorenc @damienmg This should be RFAL.

Even if I have to rework things for `docker_pull` and `docker_push` integration this should be a large step in the right direction.

docker/incremental_load.sh.tpl
```diff
@@ -22,38 +22,96 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
...
+IMAGES=$(${DOCKER} inspect $(${DOCKER} images -aq) |& \
+  grep sha256 | grep -v Id | cut -d'"' -f 2 | cut -d':' -f 2 | \
+  sort | uniq)
 IMAGE_LEN=$(for i in $IMAGES; do echo -n $i | wc -c; done | sort -g | head -1 | xargs)
 
 [ -n "$IMAGE_LEN" ] || IMAGE_LEN=64
 
-function incr_load() {
+# Create temporary files in which to record things to clean up.
+TEMP_FILES=$(mktemp)
```
damienmg: here and below, quote `"$(mktemp)"`

docker/incremental_load.sh.tpl
```diff
@@ -22,38 +22,96 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
...
+  # docker load has elision of preloaded layers built in.
+  echo "Loading legacy tarball base $1..."
+  "${DOCKER}" load -i "${tarball}"
+}
+
+function import_layer() {
   # Load a layer if and only if the layer is not in "$IMAGES", that is
   # in the local docker registry.
-  name=$(cat ${RUNFILES}/$1)
+  local name=$(cat ${RUNFILES}/$1)
```
damienmg: quote filename `$(cat "${RUNFILES}/$1")`

docker/incremental_load.sh.tpl
```diff
@@ -22,38 +22,96 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
...
 
   if (echo "$IMAGES" | grep -q ^${name:0:$IMAGE_LEN}$); then
-    echo "Skipping $name, already loaded."
+    echo "Skipping ${name}, already loaded."
   else
-    echo "Loading $name..."
-    "${DOCKER}" load -i ${RUNFILES}/$2
+    echo "Importing ${name}..."
+    TEMP_IMAGE=$(cat ${RUNFILES}/$2 | "${DOCKER}" import -)
```
damienmg: quote

docker/incremental_load.sh.tpl
```diff
@@ -22,38 +22,96 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
```
damienmg: This script has overgrown in comlexity, I wonder if we should not replace it with a python script

docker/incremental_load.sh.tpl
```diff
@@ -22,38 +22,96 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
...
-    "${DOCKER}" load -i ${RUNFILES}/$2
+    echo "Importing ${name}..."
+    TEMP_IMAGE=$(cat ${RUNFILES}/$2 | "${DOCKER}" import -)
+    echo "${TEMP_IMAGE}" >> "${TEMP_IMAGES}"
   fi
 }
 
+function import_config() {
+  # Create an image from the image configuration file.
+  local name=${RUNFILES}/$1
```
damienmg: Quote filenmae here and below.

docker/incremental_load.sh.tpl
```diff
@@ -22,38 +22,96 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
...
+  local diff_id=$(cat ${RUNFILES}/$2)
+  local layer=${RUNFILES}/$3
+
+  local tmp_dir=$(mktemp -d)
+  echo "${tmp_dir}" >> "${TEMP_FILES}"
+
+  cd "${tmp_dir}"
+  cp "${name}" the-image.json
+  cp "${layer}" "${diff_id}.tar"
+  local parent_layers=$(cat the-image.json | python -mjson.tool | \
```
damienmg: Add ( ) to create a bash array?

docker/join_layers.py
```diff
@@ -35,65 +36,105 @@
...
-def create_image(output, layers, tag_to_layer=None, layer_to_tags=None):
+def create_bundle(output, tag_to_config, diffid_to_blobsum,
+                  blobsum_to_unzipped, blobsum_to_zipped, blobsum_to_legacy):
   """Creates a Docker image from a list of layers.
 
   Args:
     output: the name of the docker image file to create.
     layers: the layers (tar files) to join to the image.
-    tag_to_layer: a map from docker_name.Tag to the layer id it references.
+    tag_to_layer: a map from google3.third_party.bazel_rules.rules_docker.docker_name.Tag to the layer id it references.
```
damienmg: google3? :)

docker/incremental_load.sh.tpl
```diff
@@ -22,38 +22,96 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
...
+  cat "${TEMP_FILES}" | xargs rm -rf> /dev/null 2>&1 || true
+  cat "${TEMP_IMAGES}" | xargs "${DOCKER}" rmi > /dev/null 2>&1 || true
+
+  rm -rf "${TEMP_FILES}"
+  rm -rf "${TEMP_IMAGES}"
+}
+trap cleanup EXIT
+
+
+function load_legacy() {
```
damienmg: This operation will be really slow if the base image is say ubuntu compressed, and we redo it at every run.

mattmoor: This path will be exceedingly rare when I'm done, so I'm not going to optimize it.  I already have a change branched off this than can leverage new pull/push utilities to perform a full roundtrip without ever dealing with the full tarball.  For folks in the mono-repo that can't benefit from `docker_pull` it introduces a new `docker_import` that they can use (which it uses under the hood).

This won't help your `docker_pull`, but consider the future change a carrot for switching :)

docker/incremental_load.sh.tpl
mattmoor: Done everywhere I could find

docker/incremental_load.sh.tpl
mattmoor: Done

docker/incremental_load.sh.tpl
mattmoor: Done

docker/join_layers.py
mattmoor: Clearly I built this elsewhere.  The funniest bit is that this rewrite (by the import tool) isn't even correct :)

docker/incremental_load.sh.tpl
mattmoor: Is there a reason a bash array is better?

mattmoor:
The reason I originally built this in the mono-repo is that we have *much* larger tarballs checked in that you guys use.  These speed up orders of magnitude and the only slow action is extracting the config from the tarball.

You guys won't even have that slow action once I'm done and you've switched to this `docker_pull` rule.  Frankly, that slow action replaces one that was 3-5x slower, so it is still an improvement (I mention it only because it's all that shows up).

As part of implementing the follow-on improvements, I've introduced `docker_import`, which Googlers can use in the mono-repo to avoid the slow path.  Outside of the mono-repo, folks should never deal with tarballs (and I plan to have a few paths emit warnings when the slow path is invoked).

docker/zip.bzl
```diff
@@ -0,0 +1,26 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
dlorenc: I wish this existed somewhere more central than this repo :(

docker/incremental_load.sh.tpl
```diff
@@ -22,38 +22,96 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
...
+  # with the final layer.
+  echo -n "\"$x.tar\","
+done
+echo -n "\"${diff_id}.tar\""
+)
+   ],
+   "RepoTags": []
+}]
+EOF
+  tar cf image.tar the-image.json "${diff_id}.tar" manifest.json
```
dlorenc: Any reason to write image.tar to disk instead of piping it right into docker load?

docker/join_layers.py
```diff
@@ -35,48 +36,96 @@
...
+                    help='A list of tarballs from which our images may derive.')
+
 parser.add_argument('--stamp-info-file', action='append', required=False,
                     help=('If stamping these layers, the list of files from '
                           'which to obtain workspace information'))
 
 
-class ImageConfig(v2_2_image.DockerImage):
-  """Create an image that only exposes config_file."""
+class FromParts(v2_2_image.DockerImage):
```
dlorenc: Any reason not to put this right into the containerregistry.client library?

docker/join_layers.py
```diff
@@ -143,21 +175,45 @@ def main():
...
+    tag_to_config[tag] = config_file
+
+  # Do this first so that if there is overlap with the loop below it wins.
+  blobsum_to_legacy = {}
+  for tar in args.legacy or []:
+    with v2_2_image.FromTarball(tar) as legacy_image:
+      config_file = legacy_image.config_file()
+      cfg = json.loads(config_file)
+      fs_layers = list(reversed(legacy_image.fs_layers()))
+      for i in xrange(len(cfg['rootfs']['diff_ids'])):
```
dlorenc: `for i, diff_id in enumerate(cfg['rootfs']['diff_ids'])` ?

docker/zip.bzl
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Functions for producing the gzip of an artifact."""
+
+def gzip(ctx, artifact):
+  """Create an action to compute the gzipped artifact."""
+  out = ctx.new_file(artifact.basename + ".gz")
+  ctx.action(
+      command = 'gzip < %s > %s' % (artifact.path, out.path),
```
damienmg: TIL: gzip header includes timestamp (bytes 5-8), see: https://tools.ietf.org/html/rfc1952#section-2.3

add `-n` option to gzip

docker/incremental_load.sh.tpl
```diff
@@ -22,38 +22,96 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
...
+IMAGES=$(${DOCKER} inspect $(${DOCKER} images -aq) |& \
+  grep sha256 | grep -v Id | cut -d'"' -f 2 | cut -d':' -f 2 | \
+  sort | uniq)
 IMAGE_LEN=$(for i in $IMAGES; do echo -n $i | wc -c; done | sort -g | head -1 | xargs)
 
 [ -n "$IMAGE_LEN" ] || IMAGE_LEN=64
 
-function incr_load() {
+# Create temporary files in which to record things to clean up.
+TEMP_FILES="$(mktemp)"
```
damienmg: TIL: `mktemp` does not work out of the box in yosemite. use `-t` option

docker/incremental_load.sh.tpl
damienmg: I wasn't suggesting to do this here, new issue: #72

docker/layers.bzl
```diff
@@ -71,11 +89,51 @@ def assemble(ctx, layers, tags_to_names, output, stamp=False):
...
   )
 
-def incremental_load(ctx, layers, images, output, stamp=False):
+def incremental_load(ctx, images, output, stamp=False):
   """Generate the incremental load statement."""
   stamp_files = []
   if stamp:
     stamp_files = [ctx.info_file, ctx.version_file]
+
+  load_statements = []
```
dlorenc: +1 on rewriting the incr_load script in python. Python generating bash makes me sad.

damienmg:
The problem on continuous-integration is that we need to have some good images and we use `docker build` for them. We used to use Bazel but we had to use deb package from the web but they were falling out all the time, then we switched to a rule calling `docker build`

Here the one that use the most package we need to build:
```
FROM ubuntu:wily
RUN apt-get update
RUN apt-get install -y zip g++ openjdk-8-jdk openjdk-8-source wget git unzip python python3 curl dpkg-dev
ENV JAVA_HOME /usr/lib/jvm/java-8-openjdk-amd64
RUN apt-get install -y golang make jekyll reprepro devscripts
RUN apt-get install -y ssh
```

Of course we could upload those image on some container registry but it would be better if we do not have to maintain custom images.

docker/join_layers.py
mattmoor: I don't like the signature :(

docker/zip.bzl
mattmoor: me too

docker/incremental_load.sh.tpl
mattmoor: Done

docker/incremental_load.sh.tpl
mattmoor: Done

mattmoor:
FML, I made those edits on my other PR's branch.  redoing now...

mattmoor:
@damienmg I assume you mean they were failing because the packages source drifted?  You should look at the debian packages rules @dlorenc's team built in distroless.

mattmoor:
This should be RFAL, I was able to cherry pick the commit :)

mattmoor:
I am hitting a weird bug on the incremental load path, I've added some new e2e testing locally that will basically `bazel run` all of the `docker_build` or `docker_bundle` targets under `docker/testdata`, and was able to repro it with a small example.  I'll post back here once I have more information.

mattmoor:
The problem is `docker import`.  It gets the right `diff id` into the daemon, but the daemon [seems to rely](https://github.com/moby/moby/blob/254fc83cba90ed79c78f4cb0cb33aeeaff492798/image/tarexport/load.go#L98) on the chain id produced by sequencing the diff ids (effectively the context-sensitive diff id), so I think `docker import`ing more than the base layer is a non-starter.  I think by some miracle I've avoided this path through development.

I'll rejigger things until I can get my new e2e test that loads the world working.

mattmoor:
This should be RFAL.

docker/zip.bzl
damienmg: We could put that in bazel directly but that would be the wrong direction. We could also create a new repo that would have pkg_tar, gzip and some more.

mattmoor: WTB skylib (before I build my own)... :)

mattmoor:
Whatever it is needs to be decoupled from the Bazel release cycle.  e.g. I had to add `sha256` back because Bazel still didn't have it when I broke these rules out (it should have it now).

mattmoor:
@dlorenc LMK if you had anything further, if not I'll submit and get the next piece into a more readily reviewable state.

mattmoor:
Thanks guys, I'm very excited to see this going in :D

PR #78, opened 2017-06-29
Introduce a docker_import rule.

mattmoor:
`docker_import` imports the constituent elements of an image into our intermediate form.

This will shortly become an implementation detail of `docker_pull`, but in mono-repo scenarios it can be used to more efficiently commit an image to avoid expensive legacy `ExtractConfig` actions.

### DIFF
--- a/docker/testdata/BUILD
+++ b/docker/testdata/BUILD
@@ -17,6 +17,7 @@ load(
     "//docker:docker.bzl",
     "docker_build",
     "docker_bundle",
+    "docker_import",
     "docker_push",
 )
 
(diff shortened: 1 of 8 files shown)


docker/testdata/BUILD
```diff
@@ -345,3 +346,33 @@ docker_build(
...
 )
+
+PAUSE_CONFIG = "2b58359142b094165abfc9ad9c327c453a3eaa624d0c5ee8f0ed1ab50483603b.json"
+
+genrule(
+    name = "pause_config",
+    cmd = ("tar xOf $(location :pause.tar) %s " +
+           "> $(location :pause_config.json)") % PAUSE_CONFIG,
+    outs = ["pause_config.json"],
+    srcs = [":pause.tar"],
```
dlorenc: Did you mean to check in a pause.tar? I don't see the base file anywhere.

dlorenc: D'oh, nevermind. The file is already in here. Disregard :)

Issue #79, opened 2017-06-30
docker_import may result in layer mismatch


mattmoor:
Resulting in:
```
invalid manifest, layers length mismatch: expected 6, got 4
```

The bug is here:
```
  TOTAL_DIFF_IDS=($(cat "${name}" | python -mjson.tool | \
      grep sha256 | cut -d'"' -f 2 | cut -d':' -f 2))
```

Essentially, this doesn't filter things enough.  The config may contain lines like:
```
        "Image": "sha256:3fdb6bcbf1c1e852e13c0defe66743162cd07bee677bdbdd82db7f3f395349e4",
```

Which results in the `manifest.json` containing additional entries like `"Image.tar"`.

The fix is to expand the filter to something like:  `grep -E '^ +"sha256:'`

PR #80, opened 2017-06-30
Fix incremental loading of `docker_import`.

mattmoor:
Fixes https://github.com/bazelbuild/rules_docker/issues/79

### DIFF
--- a/docker/incremental_load.sh.tpl
+++ b/docker/incremental_load.sh.tpl
@@ -22,10 +22,18 @@ RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
 
 DOCKER="${DOCKER:-docker}"
 
+function list_diffids() {
+    for image in $("${DOCKER}" images -aq 2> /dev/null);
+    do
+	for entry in $("${DOCKER}" inspect "${image}" -f '{{json .RootFS.Layers}}');
+	do
+	    echo -n $entry | python -mjson.tool | grep sha256 | cut -d'"' -f 2 | cut -d':' -f 2
+	done
+    done
+}
+
 # Fetch the diff ids of the layers loaded in the docker daemon already
-IMAGES=$(${DOCKER} inspect $(${DOCKER} images -aq) |& \
-  grep sha256 | grep -v Id | cut -d'"' -f 2 | cut -d':' -f 2 | \
-  sort | uniq)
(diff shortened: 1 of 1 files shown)


docker/incremental_load.sh.tpl
```diff
@@ -24,7 +24,7 @@ DOCKER="${DOCKER:-docker}"
 
 # Fetch the diff ids of the layers loaded in the docker daemon already
 IMAGES=$(${DOCKER} inspect $(${DOCKER} images -aq) |& \
-  grep sha256 | grep -v Id | cut -d'"' -f 2 | cut -d':' -f 2 | \
+  grep -E '^ +"sha256:' | grep -v Id | cut -d'"' -f 2 | cut -d':' -f 2 | \
```
promiseofcake: In lieu of requiring `jq` to explicitly select the `Layers` values, this looks like it will work 👍

mattmoor: Yeah, I'd love to depend on `jq`, but I think that's a non-starter for some of the environments this needs to work in :(

dlorenc: What about something like:

```shell
docker inspect $(docker images -aq) -f '{{range .RootFS.Layers}} {{.}} {{end}}' | sort | uniq
```

mattmoor: neat, I'll do that

mattmoor: So that doesn't exactly work as is, but solves our filtering problem if we use:
```
for entry in $(docker inspect $(docker images -aq) -f '{{json .RootFS.Layers}}'); do echo -n $entry | python -mjson.tool; done | grep sha256 | cut -d'"' -f 2 | cut -d':' -f 2 | sort | uniq
```

mattmoor: ok, I updated the PR to use the inspect filtering, nice trick.

mattmoor:
lol, I changed the wrong line moving my fix into the Git repo...  As I said in the linked issue, this is relevant to the line:
```
  TOTAL_DIFF_IDS=($(cat "${name}" | python -mjson.tool | \
      grep sha256 | cut -d'"' -f 2 | cut -d':' -f 2))
```

I changed the `docker inspect` line by mistake, oops.

mattmoor:
Sadly this means we can't use `inspect` formatting tricks.

Issue #77, opened 2017-06-29
docker_repositories() fails if user's workspace already contains "subpar"


dmitry-xnor:
We use `subpar` for packaging Python. When trying to introduce Docker support using WORKSPACE file (following the example in README.md), I get the following:

```
	File "/home/dmitryb/.cache/bazel/_bazel_dmitryb/8e618220280fa417619d9462a0a41d8d/external/io_bazel_rules_docker/docker/docker.bzl", line 128, in docker_repositories
		native.git_repository(name = "subpar", remote = "https:/...", ...")
Cannot redefine repository after any load statement in the WORKSPACE file (for repository 'subpar').
ERROR: Error evaluating WORKSPACE file.
ERROR: error loading package 'external': Package 'external' contains errors.
```
Seems like you're loading `subpar` as a part of docker_repositories(), and it's conflicting with the existing `subpar` target we already have.

Maybe namespacing it in some way (e.g. `rules_docker_subpar`) could prevent this?

promiseofcake:
Related: https://github.com/bazelbuild/rules_docker/issues/55

mattmoor:
Yeah I think #55 is what we want, I just haven't had a moment to do it amidst some improvements I've been working on.  If the workaround in #55 doesn't work for you, then I'll try to TAL at it sooner than later.

Issue #55, opened 2017-06-16
Allow Ignoring of pre-loaded packages.


promiseofcake:
Via the `docker_repositories()` initialization command it would be great if we had a way to re-use existing dependencies. 

Let's say for example I was attempting to build something in the `tensorflow/tensorflow` `tf_workspace`, that we have loaded up. It turns out they have already imported `six`:

https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace.bzl#L302
```
native.bind(
    name = "six",
    actual = "@six_archive//:six",
)
```

Without being able to override in some manner we are unable to use the `docker_build` rules in the same workspace. [pubref/rules_protobuf](https://github.com/pubref/rules_protobuf) an interesting way to handle these sorts of dependencies:
https://github.com/pubref/rules_protobuf#overriding-or-excluding-workspace-dependencies

```
go_proto_repositories(
  excludes = [
    "com_github_golang_glog",
  ]
)
```

Which then uses some `require` logic under the hood to load / not load the repository functions.

In interim I suppose it's possible to manually include the dependencies _other_ than the ones we have already defined and go from there. This seems like something that _might_ get solved as we move towards: https://github.com/bazelbuild/bazel/issues/1943

mattmoor:
Yeah, I think we can safely add a kwarg for that.  I hadn't seen the rules_protobuf stuff, that has a number of interesting features.

promiseofcake:
Determined that if I changed the ordering of my linked workspaces I circumvented this issue so this is merely a nice to have at this point. Thanks!

promiseofcake:
Thank you @mattmoor!

mattmoor:
@promiseofcake My pleasure, sorry this was a little slow.

promiseofcake:
No worries, thanks also for `native.existing_rules().keys()`, that is a neat trick.

promiseofcake:
Same issue when attempting to build containers in a repository that has a [tensorflow](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/workspace.bzl#L255) dependency (which is also providing that bind for `com_google_protobuf`).

Here is a related issue from the parent `containerregistry` where this dependency originates: https://github.com/google/containerregistry/issues/36. The problem is that we don't have a way to choose which `@six` we are providing due to the bind, and there doesn't seem to be a way to programmatically use `//external:six or @six//:six`.

[...]

PR #84, opened 2017-06-30
Add limited support for overriding and excluding dependencies.

mattmoor:
Fixes: https://github.com/bazelbuild/rules_docker/issues/55
Fixes: https://github.com/bazelbuild/rules_docker/issues/77

### DIFF
--- a/docker/docker.bzl
+++ b/docker/docker.bzl
@@ -32,53 +32,62 @@ CONTAINERREGISTRY_RELEASE = "v0.0.12"
 
 def docker_repositories():
   """Download dependencies of docker rules."""
-  native.http_file(
-    name = "puller",
-    url = ("https://storage.googleapis.com/containerregistry-releases/" +
-           CONTAINERREGISTRY_RELEASE + "/puller.par"),
-    sha256 = "7ebdefbb8a98c5bf8b737a7891a8860ed2f22ae149d3dba200bc104305b6e1d6",
-    executable = True,
-  )
+  excludes = native.existing_rules().keys()
 
-  native.http_file(
-    name = "pusher",
-    url = ("https://storage.googleapis.com/containerregistry-releases/" +
-           CONTAINERREGISTRY_RELEASE + "/pusher.par"),
-    sha256 = "a8e48dd99b3e85ac5fe30e1daec327ee509540ff18f5ad7f4724d84ccfd3d150",
-    executable = True,
-  )
+  if "puller" not in excludes:
+    native.http_file(
+      name = "puller",
+      url = ("https://storage.googleapis.com/containerregistry-releases/" +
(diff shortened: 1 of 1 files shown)


docker/docker.bzl
```diff
@@ -30,55 +30,65 @@ load(":push.bzl", "docker_push")
 # The release of the github.com/google/containerregistry to consume.
 CONTAINERREGISTRY_RELEASE = "v0.0.11"
 
-def docker_repositories():
+def docker_repositories(excludes=[], overrides={}):
```
dlorenc: In python there are issues with using mutable objects as default vals in functions, does that apply to skylark?

docker/docker.bzl
```diff
@@ -30,55 +30,65 @@ load(":push.bzl", "docker_push")
...
-    sha256 = "90a76d01ee57a5df7353a533c96966822c9efda61636c11eac11344171556017",
-    executable = True,
-  )
+  gcr_release = overrides.get("containerregistry", CONTAINERREGISTRY_RELEASE)
+  if "puller" not in excludes:
+    native.http_file(
+      name = "puller",
+      url = ("https://storage.googleapis.com/containerregistry-releases/" +
+             gcr_release + "/puller.par"),
+      sha256 = overrides.get(
```
dlorenc: If you're overriding it might be nice to not have to specify a SHA, wdyt?

mattmoor: I thought about it, but let's see if someone complains.  I was debating exactly this for the large list of TODOs below.  Ultimately I didn't do this because AFAIK there is no way of doing this without duplicating the rule.  @damienmg thoughts?

docker/docker.bzl
mattmoor: @damienmg ?

mattmoor: I fixed regardless.

docker/docker.bzl
damienmg: You could make it optional but generally this go against reproducibility which is going to get back at you.

BTW, maybe use native.existing_rule to make sure you do not add an existing deps instead, that would better address #55.

mattmoor:
@damienmg I believe what you are suggesting is that `native. existing_rules`s keys replaces excludes and if a user wants to override, they should just import it themselves and then `existing_rules` will do the right thing.  Is that a fair interpretation?

damienmg:
Yes it is.

mattmoor:
@damienmg then it should be RFAL, that's what I pushed :)

mattmoor:
@dlorenc any problems with the new approach?

mattmoor:
rebased after the fast push/pull stuff was merged through.

mattmoor:
@dlorenc Ping

Issue #72, opened 2017-06-28
Rewrite incremental_load.sh.template in python


damienmg:
The shell script has grown in complexity and is hard to port to new architecture, we should probably rewrite it in python at some point.

pcj:
Closing due to age/inactivity.

PR #87, opened 2017-07-04
Improve mac compatibility/portability of incremental_load script

cwoodcock:
Tried running on my mac and ran into a few issues, these are the fixes for them.

Have kept things separate in case you only wanted parts.

`27c3d4c` - Feels like a bit of a hack but seemed the easiest approach was to ask the user to install GNU tar rather than doing something more complex.  Makes the assumption that GNU tar won't change its version output.

`46d0b43`- `&|` is not portable and was giving me an unexpected token error.  Couldn't see why stdout+stderr needed to be piped though, so changed that up to be just a pipe of stdout.  If there's something I haven't considered here, let me know and I'll change it to the unabbreviated `2>&1 |`.

### DIFF
--- a/docker/incremental_load.sh.tpl
+++ b/docker/incremental_load.sh.tpl
@@ -107,9 +107,7 @@ EOF
 EOF
 
   set -o pipefail
-  tar c config.json manifest.json |& \
-      "${DOCKER}" load |& \
-      cut -d':' -f 2- >> "${TEMP_IMAGES}" 2> /dev/null
+  tar c config.json manifest.json | "${DOCKER}" load | cut -d':' -f 2- >> "${TEMP_IMAGES}" 2>/dev/null
 }
 
 function find_diffbase() {
--- a/docker/incremental_load.sh.tpl
+++ b/docker/incremental_load.sh.tpl
@@ -35,7 +35,7 @@ DOCKER="${DOCKER:-docker}"
 function list_diffids() {
   for image in $("${DOCKER}" images -aq 2> /dev/null);
   do
-    for entry in $("${DOCKER}" inspect "${image}" -f '{{json .RootFS.Layers}}');
+    for entry in $("${DOCKER}" inspect -f '{{json .RootFS.Layers}}' "${image}");
     do
       echo -n $entry | python -mjson.tool | grep sha256 | cut -d'"' -f 2 | cut -d':' -f 2
     done
--- a/docker/incremental_load.sh.tpl
+++ b/docker/incremental_load.sh.tpl
@@ -49,8 +49,8 @@ IMAGE_LEN=$(for i in $IMAGES; do echo -n $i | wc -c; done | sort -g | head -1 | 
 [ -n "$IMAGE_LEN" ] || IMAGE_LEN=64
 
 # Create temporary files in which to record things to clean up.
-TEMP_FILES="$(mktemp -t)"
-TEMP_IMAGES="$(mktemp -t)"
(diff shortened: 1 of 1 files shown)


bazel-io:

docker/incremental_load.sh.tpl
```diff
@@ -18,14 +18,24 @@ set -eu
...
 
+# Check we are using GNU tar as this script relies on some impl specifics
+# e.g. on a Mac the default is a BSD variant which will bork
+tar --version | grep 'GNU' >/dev/null 2>&1
+if [ ! $? -eq 0 ]; then
+  echo "Error: GNU tar needs to be installed."
+  echo "If you are on a Mac, install Homebrew then..."
+  echo "  brew install gnu-tar --with-default-names"
+  exit 1
+fi
```
mattmoor: I assume this is because of:
```
tar --create --absolute-names --dereference \
...
```
?

docker/incremental_load.sh.tpl
```diff
@@ -39,8 +49,8 @@ IMAGE_LEN=$(for i in $IMAGES; do echo -n $i | wc -c; done | sort -g | head -1 |
 [ -n "$IMAGE_LEN" ] || IMAGE_LEN=64
 
 # Create temporary files in which to record things to clean up.
-TEMP_FILES="$(mktemp -t)"
-TEMP_IMAGES="$(mktemp -t)"
+TEMP_FILES="$(mktemp -t 2>/dev/null || mktemp -t 'rules_docker_files')"
```
mattmoor: Can you hoist a new `temp_file()` function to encapsulate this?

docker/incremental_load.sh.tpl
mattmoor: I'd like to remove this requirement if possible, so can we at least open an issue to track that?

mattmoor:
Test this please

mattmoor:
@damienmg FYI the bot doesn't seem to like the magic phrase in a review comment :-/

docker/incremental_load.sh.tpl
cwoodcock: That's the one.

100% with you that it should go.  However I saw #72 and figured that was the desired mid/long term approach.  If that ticket happens then this will become mute.  Will create an issue in a minute anyway so it what we've discussed here doesn't get lost to the sands of time.

docker/incremental_load.sh.tpl
cwoodcock: Will be a bit of a delay before I can look at that, should be able to get to it Fri or Sat.  Either hold the merge until then or I'll send in another PR.

mattmoor: Ok, the Bazel CI setup seems a bit wedged at the moment, so we'll see which happens first.

damienmg:
The CI is down with slaves offline since yesterday. I guess the team has
been unsuccessful in fixing the issue so far. I was not able to investigate
so far, I'll be back in muc next week and investigate then if the issue is
still there.

damienmg:
Ci is back up: test this please

docker/incremental_load.sh.tpl
damienmg: Why not just `mktemp -t rules_docker_images`?

mattmoor:
I'm going to merge as-is, as I'd rather unblock Mac usage than wait for this largely cosmetic piece of feedback.  thanks again for the PR.

PR #89, opened 2017-07-06
incremental_load.sh.tpl: Use standard tar syntax.

### LINKED ISSUE #88: Remove requirement on GNU tar
A patch in #87 explicitly made GNU tar a requirement.

The root cause of the issue that prompted the change is tar usage within [incremental_load.sh.tpl](https://github.com/bazelbuild/rules_docker/blob/master/docker/incremental_load.sh.tpl).  `tar --create --absolute-names --dereference ...` causes errors when the BSD variant is used (default on Macs).

It would be good to remove the prerequisite on GNU tar within the script so other implementations (especially BSD) could be used without users having to do anything.

Implementing #72 will most likely supersede this.

ohler:
Fixes #88.

### DIFF
--- a/docker/incremental_load.sh.tpl
+++ b/docker/incremental_load.sh.tpl
@@ -18,16 +18,6 @@ set -eu
 
 # This is a generated file that loads all docker layers built by "docker_build".
 
-# Check we are using GNU tar as this script relies on some impl specifics
-# e.g. on a Mac the default is a BSD variant which will bork
-tar --version | grep 'GNU' >/dev/null 2>&1
-if [ ! $? -eq 0 ]; then
-  echo "Error: GNU tar needs to be installed."
-  echo "If you are on a Mac, install Homebrew then..."
-  echo "  brew install gnu-tar --with-default-names"
-  exit 1
-fi
-
 RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
 
 DOCKER="${DOCKER:-docker}"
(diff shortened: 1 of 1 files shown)


bazel-io:

docker/incremental_load.sh.tpl
```diff
@@ -214,7 +204,7 @@ EOF
   # We minimize reads / writes by symlinking the layers above
   # and then streaming exactly the layers we've established are
   # needed into the Docker daemon.
-  tar --create --absolute-names --dereference \
+  tar cPh \
```
mattmoor: nit: this will fit on a single line now.

mattmoor:
Test this please

(For Bazel CI)

mattmoor:
Hmm, that was easier than I thought it would be.  @cwoodcock Can you confirm this works for you?

My bad for going with the longer flag names, I mostly did it for an abundance of clarity, but who knew it would have compat issues :(

ohler:
My only testing of this was manual, with a few docker rules I happened to have, on Mac OS Sierra.  Is this code covered by the Travis tests?

mattmoor:
No, I'm not aware of a good way of testing this, which is why I was hoping @cwoodcock might weigh in because he added some of this logic.

cwoodcock:
Sorry for the delay, been out and about.

Nice fix @ohler.  I should have dug deeper into how to rewrite the flags.

Not covered at the moment afaik.  But Travis does have an [OSX environment](https://docs.travis-ci.com/user/osx-ci-environment/) so hopefully it will be possible to reconfig the tests to allow that to happen.  Will do some discovery and feedback onto an issue rather than here.

Will manually test+confirm in the morning.

cwoodcock:
LGTM

mattmoor:
Thanks, guys!

PR #98, opened 2017-07-21
First cut at a foo_image rule

mattmoor:
This change is a first-cut at attempting a `foo_image` rule that directly assembles a Docker image compatible with `foo_binary`, but leveraging a sharded archive format (vs. a mono-archive format like [PAR](https://github.com/google/subpar) or "deploy JAR")

The advantage of the sharded archive format is simple:  **We want to translate build-time incrementality into redeployment incrementality.**

By enabling the partitioning of dependencies into their own layers, in a binary agnostic fashion, we enable:
1. sharing of those dependencies across different images that depend on them
1. no republishing of those dependencies' layers when they haven't changed
1. no repulling unchanged dependencies' layers

In this model, what gets repushed/repulled can be made a direct function of Bazel's library-level incrementality.

### DIFF
--- a/docker/build.bzl
+++ b/docker/build.bzl
@@ -11,7 +11,35 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""Rule for building a Docker image."""
+"""Rule for building a Docker image.
+
+In addition to the base docker_build rule, we expose its constituents
+(attr, outputs, implementation) directly so that others may expose a
+more specialized build leveraging the same implementation.  The
+expectation in such cases is that users will write something like:
+
+  load(
+    "@io_bazel_rules_docker//docker:build.bzl",
+    _build_attrs="attrs",
+    _build_outputs="outputs",
+    _build_implementation="implementation",
+  )
+
+  def _impl(ctx):
+    ...
+    return _build_implementation(ctx, ... kwarg overrides ...)
+
+  _foo_image = rule(
+      attrs = _build_attrs + {
+         # My attributes, or overrides of _build_attrs defaults.
+         ...
+      },
+      executable = True,
+      outputs = _build_outputs,
+      implementation = _impl,
+  )
+
+"""
 
 load(
     ":filetype.bzl",
@@ -51,14 +79,14 @@ load(
     _serialize_dict = "dict_to_associative_list",
 )
 
-def _build_layer(ctx):
+def _build_layer(ctx, files=None, directory=None, symlinks=None):
   """Build the current layer for appending it the base layer."""
 
   layer = ctx.outputs.layer
   build_layer = ctx.executable.build_layer
   args = [
       "--output=" + layer.path,
-      "--directory=" + ctx.attr.directory,
+      "--directory=" + directory,
       "--mode=" + ctx.attr.mode,
   ]
 
@@ -68,23 +96,23 @@ def _build_layer(ctx):
         dirname(ctx.outputs.out.short_path),
         _canonicalize_path(ctx.attr.data_path))
     args += ["--file=%s=%s" % (f.path, strip_prefix(f.short_path, data_path))
-             for f in ctx.files.files]
+             for f in files]
   else:
     # Otherwise, files are added without a directory prefix at all.
     args += ["--file=%s=%s" % (f.path, f.basename)
-             for f in ctx.files.files]
+             for f in files]
 
   args += ["--tar=" + f.path for f in ctx.files.tars]
   args += ["--deb=" + f.path for f in ctx.files.debs if f.path.endswith(".deb")]
-  args += ["--link=%s:%s" % (k, ctx.attr.symlinks[k])
-           for k in ctx.attr.symlinks]
+  args += ["--link=%s:%s" % (k, symlinks[k])
+           for k in symlinks]
   arg_file = ctx.new_file(ctx.label.name + ".layer.args")
   ctx.file_action(arg_file, "\n".join(args))
 
   ctx.action(
       executable = build_layer,
       arguments = ["--flagfile=" + arg_file.path],
-      inputs = ctx.files.files + ctx.files.tars + ctx.files.debs + [arg_file],
+      inputs = files + ctx.files.tars + ctx.files.debs + [arg_file],
       outputs = [layer],
       use_default_shell_env=True,
       mnemonic="DockerLayer"
@@ -101,7 +129,7 @@ def _get_base_config(ctx):
     l = _get_layers(ctx, ctx.attr.base, ctx.files.base)
     return l.get("config")
 
-def _image_config(ctx, layer_name):
+def _image_config(ctx, layer_name, entrypoint=None, cmd=None):
   """Create the configuration for a new docker image."""
   config = ctx.new_file(ctx.label.name + ".config")
 
@@ -119,9 +147,9 @@ def _image_config(ctx, layer_name):
   args = [
       "--output=%s" % config.path,
   ] + [
-      "--entrypoint=%s" % x for x in ctx.attr.entrypoint
+      "--entrypoint=%s" % x for x in entrypoint
   ] + [
-      "--command=%s" % x for x in ctx.attr.cmd
+      "--command=%s" % x for x in cmd
   ] + [
       "--ports=%s" % x for x in ctx.attr.ports
   ] + [
@@ -168,17 +196,26 @@ def _repository_name(ctx):
   # the v2 registry specification.
   return _join_path(ctx.attr.repository, ctx.label.package)
 
-def _docker_build_impl(ctx):
+def implementation(ctx, files=None, directory=None,
+                   entrypoint=None, cmd=None, symlinks=None):
   """Implementation for the docker_build rule."""
 
+  files = files or ctx.files.files
+  directory = directory or ctx.attr.directory
+  entrypoint = entrypoint or ctx.attr.entrypoint
+  cmd = cmd or ctx.attr.cmd
+  symlinks = symlinks or ctx.attr.symlinks
+
   # Generate the unzipped filesystem layer, and its sha256 (aka diff_id).
-  unzipped_layer, diff_id = _build_layer(ctx)
+  unzipped_layer, diff_id = _build_layer(ctx, files=files,
+                                         directory=directory, symlinks=symlinks)
 
   # Generate the zipped filesystem layer, and its sha256 (aka blob sum)
   zipped_layer, blob_sum = _zip_layer(ctx, unzipped_layer)
 
   # Generate the new config using the attributes specified and the diff_id
-  config_file, config_digest = _image_config(ctx, diff_id)
+  config_file, config_digest = _image_config(
+    ctx, diff_id, entrypoint=entrypoint, cmd=cmd)
 
   # Construct a temporary name based on the build target.
   tag_name = _repository_name(ctx) + ":" + ctx.label.name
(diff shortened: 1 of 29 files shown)


mattmoor:
@duggelz I can't make you a reviewer (thanks, GitHub), but I'd very much appreciate your feedback on this.  I'm sure I'm missing tons of stuff (e.g. the `empty_filenames` fun) that you've surely dealt with in `subpar`.

mattmoor:
@damienmg I have no idea why the Bazel presubmit is unhappy, when I click through the `!`s everything is green...

damienmg:
build_test is failing on all the config:

http://ci.bazel.io/blue/rest/organizations/jenkins/pipelines/PR/rules_docker/runs/66/nodes/14/log/?start=0

mattmoor:
Thanks @damienmg.  Gotta say, I kinda hate the new Jenkins UI :-/  Don't show me green for the step that failed, so I need to expand every stupid thing. </grumble>

mattmoor:
I added `cc_image` too.

I debated adding `go_image` too, but I think adding tests would and up requiring the `go_prefix` stuff, which might make `rules_go` a harder dependency than I'd like.

I'll experiment with JAR / WAR rules once I'm back from vacation.

mattmoor:
@damienmg thoughts?

docker/testdata/BUILD
```diff
@@ -344,13 +344,13 @@ load(
...
-   name = "bundle_to_push",
-   images = {
-     "us.gcr.io/convoy-adapter/bazel-test/{BUILD_USER}:foo": ":with_env",
-     "eu.gcr.io/convoy-adapter/bazel-test/{BUILD_USER}:bar": ":with_double_env",
-     "asia.gcr.io/convoy-adapter/bazel-test/{BUILD_USER}:baz": ":with_user",
-   },
-   stamp = True
+    name = "bundle_to_push",
+    images = {
+        "us.gcr.io/convoy-adapter/bazel-test/{BUILD_USER}:foo": ":with_env",
```
duggelz: It looks like this is both a functionality test, and implicitly a system test of the gcr regions (or whatever they're called).  Can you explicitly document that this is "a docker image from each GCR region" so that it's clear that we need to add a "moon.gcr.io" when we open the lunar base?

docker/testdata/py_image.py
```diff
@@ -0,0 +1,25 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from docker.testdata import py_image_library
+
+def main():
+  print 'First: %d' % py_image_library.fn(1)
```
duggelz: Can you make these print statements Python3-compatible?

testing/e2e.sh
```diff
@@ -62,6 +62,77 @@ EOF
...
+  "@io_bazel_rules_docker//docker:docker.bzl",
+  "docker_repositories",
+)
+docker_repositories()
+
+# We must load these before the go_image rule.
+git_repository(
+    name = "io_bazel_rules_go",
+    remote = "https://github.com/bazelbuild/rules_go.git",
+    tag = "0.4.4",
```
duggelz: We should either collect all the pinned versions/hashes in a document somewhere, or have a janitor-bot remind us to update them on a regular basis.

docker/contrib/python/image.bzl
```diff
@@ -0,0 +1,171 @@
...
+      # Handle empty files by linking to /dev/null
+      "/app/" + empty: "/dev/null"
+      for empty in ctx.attr.dep.default_runfiles.empty_filenames
+    }
+  )
+
+_dep_layer = rule(
+    attrs = _build_attrs + {
+        # The base image on which to overlay the dependency layers.
+        "base": attr.label(default = Label("@py_image_base//image")),
```
mattmoor: That's a good question.

For the latter, I think this is just up to the toolchain that `rules_go` pulls in, and I punt on importing the Go rules in this repo (they must be imported before `contrib/go/image.bzl` or it will fail).

For the former, I haven't done anything specifically to address this, but I think the story should be one of:
1. Change the `base` to an appropriate image (requires exposing `base`, which has a TODO now)
1. Make the rules smart about `PY2`, `PY3`, `PY2AND3` in its choice of the distroless base (and perhaps override-able by the above) .

I would love to hear your thoughts and/or clever ideas here.

docker/contrib/cc/image.bzl
```diff
@@ -0,0 +1,173 @@
...
+    ctx,
+    # We put the files from dependency layers into a binary-agnostic
+    # path to increase the likelihood of layer sharing across images,
+    # then we symlink them into the appropriate place in the app layer.
+    # This references the binary package because the file paths are
+    # relative to it, and normalized by the tarball package.
+    directory="/app/" + ctx.label.package,
+    files=list(ctx.attr.dep.default_runfiles.files),
+    symlinks={
+      # Handle empty files by linking to /dev/null
```
mattmoor: I added a comment for now linking to your issues, this is a great catch thanks.

docker/contrib/cc/image.bzl
```diff
@@ -0,0 +1,173 @@
...
+  # the application layer, link to their binary-agnostic
+  # location from the runfiles path.
+  symlinks = {
+    binary_name: directory + "/" + basename
+  } + {
+    directory + "/" + input: "/app/" + input
+    for input in available
+  } + {
+    # Handle empty files by linking to /dev/null
+    base_directory + "/" + empty: "/dev/null"
```
mattmoor: Ack

mattmoor:
I pushed support for `war_image`, which is mildly different.  I am going to return to experimenting with `java_image`, for which I may end up refactoring some of this new code to share.

This is largely based on [this](https://github.com/bazelbuild/rules_appengine/blob/master/appengine/appengine.bzl#L130-L139), since I'm no WAR expert...

docker/contrib/python/image.bzl
```diff
@@ -0,0 +1,171 @@
...
+    for empty in empty_files
+  }
+
+  return _build_implementation(
+    ctx, files=files,
+    # Use entrypoint so we can easily add arguments when the resulting
+    # image is `docker run ...`.
+    # Per: https://docs.docker.com/engine/reference/builder/#entrypoint
+    # we should use the "exec" (list) form of entrypoint.
+    entrypoint=['/usr/bin/python', binary_name],
```
duggelz: You're right on both counts, I wasn't thinking.

docker/build.bzl
```diff
@@ -68,23 +102,26 @@ def _build_layer(ctx):
...
-             for f in ctx.files.files]
+             for f in files]
 
   args += ["--tar=" + f.path for f in ctx.files.tars]
   args += ["--deb=" + f.path for f in ctx.files.debs if f.path.endswith(".deb")]
-  args += ["--link=%s:%s" % (k, ctx.attr.symlinks[k])
-           for k in ctx.attr.symlinks]
+  for k in symlinks:
+    if ':' in k:
+      fail("The source of a symlink cannot contain ':', got: %s" % k)
```
damienmg: why?

docker/contrib/cc/image.bzl
```diff
@@ -0,0 +1,60 @@
...
+  if "cc_image_base" not in excludes:
+    docker_pull(
+      name = "cc_image_base",
+      registry = "gcr.io",
+      repository = "distroless/cc",
+      # 'latest' circa 2017-07-21
+      digest = "sha256:942eb947818e7e32200950b600cc94d5477b03e0b99bf732b4c1e2bba6eec717",
+    )
+
+def cc_image(name, base=None, deps=[], layers=[], **kwargs):
```
damienmg: cc_docker_build to be consistent in naming?

docker/contrib/go/image.bzl
```diff
@@ -0,0 +1,64 @@
...
+  if "go_image_base" not in excludes:
+    docker_pull(
+      name = "go_image_base",
+      registry = "gcr.io",
+      repository = "distroless/base",
+      # 'latest' circa 2017-07-21
+      digest = "sha256:06fcd3edcfeefe13b82fa8bdb9e3f4fa3bf4c7e8fe997bee0230e392f77d0e04",
+    )
+
+def go_image(name, base=None, deps=[], layers=[], **kwargs):
```
damienmg: idem go_docker_build

testing/e2e.sh
damienmg: I know :( unfortunately it is not on our roadmap yet. I think we should build some kind of tools that would update those versions callable from bazel e.g. `bazel tools bump-dep-versions` o

docker/build.bzl
```diff
@@ -11,7 +11,35 @@
...
+"""Rule for building a Docker image.
+
+In addition to the base docker_build rule, we expose its constituents
+(attr, outputs, implementation) directly so that others may expose a
+more specialized build leveraging the same implementation.  The
+expectation in such cases is that users will write something like:
+
+  load(
+    "@io_bazel_rules_docker//docker:build.bzl",
+    _build_attrs="attrs",
```
damienmg: That would be good to respect the naming we do native sandwich.

E.g. that would be:
_docker="docker"

docker would then be a struct inside build.bzl:
docker = struct(
  build = struct(
      impl=_implementation,
      attrs=_attrs,
      outputs=_outputs
  )
)

And you can call _docker.build.impl directly. Also we might want to switch to [named provider](https://bazel.build/designs/skylark/declared-providers.html) if we start offering interaction API

docker/build.bzl
mattmoor: Because of how we pass it to the underlying binary.
  "foo:bar": "baz"

Would be passed indistinguishably from:
  "foo": "bar:baz"

docker/build.bzl
mattmoor: It took me a little while to understand what you were suggesting with the structs, but I still have no idea what you are suggesting with the providers.  It looks to me like Bazel has reinvented proto options, but I'm not sure how that helps here.

docker/contrib/cc/image.bzl
mattmoor: I think I prefer the current naming given it's current home under `rules_docker`.  `docker_image` probably would have been a better name :)

docker/build.bzl
damienmg: There is an internal doc (/cc @laurentlb for making that doc public) about cleaning up namespace of native symbols. A consequence would be that we would write java.merge instead of java_common.merge, with `java` being a symbol loaded from an extension. That's the reason for struct.

For named providers this is just something to consider since this should be the new way.

mattmoor:
@damienmg Any suggestions for how I add empty files at arbitrary paths?

duggelz:
Re: symlinks to empty files, the Java code inside Bazel handles this with a Map where entries have the "symlink target" key as whatever, and the "symlink source" value is NULL.  So maybe try constructing the symlink dictionary with entries like `'myfilename': None`, and see if that None survives all the way from Skylark to Bazel to the sandboxer.

https://github.com/bazelbuild/bazel/commit/738e5727bab23961c7cc423606f38f8b97a8150e#diff-d336e8f8c32246f283d163ad11f0c40dR106

mattmoor:
@duggelz the symlink dictionary we use directs a custom script (`build_tar.py`, currently forked from `pkg_tar`) on how to create symlinks within the resulting tarball (it isn't for Bazel's symlink forest).

That said, we can probably plumb through a list of empty filenames for it to create.  I'll look at that next.


FWIW, I just fixed a bug in `java_image` and modified the test to use an external dependency to cover it.

mattmoor:
Ok, we no longer symlink `/dev/null` for empty files.

mattmoor:
I'm going to merge this to checkpoint, as I already have at least one follow on change, and I expect to find bugs as folks start to use this.  Feel free to continue to comment if you find any issues and I'll send PRs to fix.

docker/contrib/go/image.bzl
```diff
@@ -0,0 +1,64 @@
...
+  """Constructs a Docker image wrapping a go_binary target.
+
+  Args:
+    layers: Augments "deps" with dependencies that should be put into their own layers.
+    **kwargs: See go_binary.
+  """
+  binary_name = name + ".binary"
+
+  if layers:
+    print("go_image does not benefit from layers=[], got: %s" % layers)
```
mmikulicic: @mattmoor I'm confused about this warning. I'm I not supposed to pass layers when using go_image?

PR #101, opened 2017-07-26
Rename set to depset

vladmos:
`set` is a deprecated alias of `depset` and will be removed soon.

Also, they are not hashsets, if you need a hashset you can use a dict with empty values.


vladmos:
Updated the description.

mattmoor:
Perhaps this would make a good addition to an OSS skylib?
```python
def set(my_list):
    return {key: None for key in my_list}.keys()
```

vladmos:
Sure, there will be such a function. We may introduce an actual hashset data type in the future, that's another reason to rename all existing `set`s to `depset`s and to deprecate `set`s (at least temporarily).

PR #112, opened 2017-08-10
Run the container when `bazel run` on an image.

mattmoor:
This makes the `bazel run` behavior on `docker_build` configurable and opts `foo_image` into running the container when the target is `bazel run`.

This does not change the default for `docker_build`, but users can now opt into it via `legacy_run_behavior = False`.

### DIFF
--- a/docker/incremental_load.sh.tpl
+++ b/docker/incremental_load.sh.tpl
@@ -239,5 +239,8 @@ function read_variables() {
 %{tag_statements}
 
 # An optional "docker run" statement for invoking a loaded container.
-# This generated and injected by docker_*.
-%{run_statements}
+# This is not executed if the single argument --norun is passed.
+if [ "a$@" != "a--norun" ]; then
+  # This generated and injected by docker_*.
+  %{run_statements}
+fi
--- a/docker/incremental_load.sh.tpl
+++ b/docker/incremental_load.sh.tpl
@@ -243,4 +243,6 @@ function read_variables() {
 if [ "a$@" != "a--norun" ]; then
   # This generated and injected by docker_*.
   %{run_statements}
+  # Empty if blocks can be problematic.
+  echo > /dev/null
 fi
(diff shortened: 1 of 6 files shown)


docker/incremental_load.sh.tpl
```diff
@@ -237,3 +237,7 @@ function read_variables() {
 # List of 'tag_layer' statements for all tags.
 # This generated and injected by docker_*.
 %{tag_statements}
+
+# An optional "docker run" statement for invoking a loaded container.
+# This generated and injected by docker_*.
+%{run_statements}
```
damienmg: Maybe make the run optional via an argument of the runner instead?

So I can activate the new behavior with `bazel run //my:target -- --run` and there is no need for the legacy_XXX attr in the rule.

mattmoor: The arguments pass through, and so I don't want to wade into the morass of distinguishing between user arguments and script arguments (or worse: wrapping script arguments).  My ultimate goal would be to have `legacy_XXX` opt you OUT (opt in by default).  I have considered adding back the `:foo.load` target, but Bazel's story around `runfiles` for runnable implicit targets leaves a lot to be desired, which is why I didn't do both here.

mattmoor:
I updated the PR to accept a single `--norun` argument to disable execution when it has been enabled.  This naked argument lets me skirt parsing out an argument from a set of arguments intended for the user app.  Should be RFAL.

mattmoor:
```shell
/home/mattmoor/.cache/bazel/_bazel_mattmoor/5c340bb9d399314dffbfbc3c347e19fb/execroot/io_bazel_rules_docker/bazel-out/local-fastbuild/bin/docker/testdata/workdir_with_tar_base: line 246: syntax error near unexpected token `;'
```

mattmoor:
Changed to use `/dev/null`

PR #115, opened 2017-08-11
Avoid passing an empty list for deps, when there are no deps.

mattmoor:
This is to avoid an overly strict check in Bazel when a JAR built with java_library is being turned into a binary via runtime_deps + main_class, when there should be no `srcs` or `deps` kwargs.


mattmoor:
I updated this to include properly handling the `runtime_deps` attr, without which we currently drop JARs on the floor.

mattmoor:
@dlorenc Damien seems out, can you TAL?

PR #117, opened 2017-08-14
Add a `precompress_layers` option to `docker_build()`.

jmillikin-stripe:
This is used to move the layer compression from the push stage to the
build stage, where machines usually have more CPU available.

Depends on https://github.com/google/containerregistry/pull/13 because the layer joining is implemented on top of containerregistry's DockerImage logic.

### DIFF
--- a/docker/build.bzl
+++ b/docker/build.bzl
@@ -283,7 +283,8 @@ def _impl(ctx, files=None, empty_files=None, directory=None,
   }
 
   _incr_load(ctx, images, ctx.outputs.executable)
-  _assemble_image(ctx, images, ctx.outputs.out)
+  _assemble_image(ctx, images, ctx.outputs.out,
+                  precompress_layers=ctx.attr.precompress_layers)
 
   runfiles = ctx.runfiles(
       files = unzipped_layers + diff_ids + [config_file, config_digest] +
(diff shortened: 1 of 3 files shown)


bazel-io:

mattmoor:
@jmillikin-stripe I'd love to hear more about how you guys are using things.  I'd guess from this change that you have some sort of infrastructure you are building around this?

If you are open/able to discuss, I'd love to hear more.  Are you planning to attend the Bazel conference in November?

jmillikin-stripe:
> I'd love to hear more about how you guys are using things. I'd guess from this change that you have some sort of infrastructure you are building around this? If you are open/able to discuss, I'd love to hear more.

I'm not sure "infrastructure" is the right word, it's nothing that fancy. We're using Docker image tarballs as an inert transport format between our build servers (which run Docker and Bazel) and a separate VM that has permission to push to our Docker repositories.

We don't have any way to grant scoped push permission to particular builds, those permissions are scoped to the entire VM role (think MDB machine owner). So we don't want semi-trusted code in the build tree to be able to push images directly, it would be able to push to any repo it wanted to.

Some of our images are big and the pusher VMs don't have much CPU, so often a "push" would take 10-20 minutes gzipping everything before deciding that it only needed to upload a few megs of changed top layer.

> Are you planning to attend the Bazel conference in November?

This is the first I've heard of it. I'll see about signing up.

damienmg:
jenkins: test this please

PR #130, opened 2017-09-04
Link to docs at top of README

drigz:
Currently, these links go to the examples, and you have to scroll down to find the full docs.


mattmoor:
Only buildifier failed, which is red at head, so merging this.

Issue #136, opened 2017-09-08
Windows support


jolarsso:
I'm working on a project where we are evaluating using bazel for building and also running some of our services in the cloud (Docker/Kubernetes). The problem is that enough of our developers are running Windows that we want that to work also before deciding on a transition. I've been tasked with trying it out on Windows 7 and possibly also Windows 10 down the road.

I have followed the installation instructions for bazel on docs.bazel.build and it's dependecies for Windows and made sure I have 64-bit versions of everything.

I initially thought my problems where related to my enviroment and how it all was installed so I have tried everything including reinstalling everything by downloading all dependencies by hand and installing (not using choclatey like I did on the first try) and the exact same problem still remains. I narrowed the problem down to the rules_docker rule and tried to run the simplest possible example I could find to use it.

When I run the example borrowed from the issue reported [here](https://github.com/bazelbuild/rules_docker/issues/116). I run just the first part of the several steps process by executing:

`bazel build //:hello.tar`

And I get this output:

____Loading

 package: @local_config_cc//
____Loading complete.  Analyzing...
____Loading package: @io_bazel_rules_docker//docker
ERROR: E:/projects/rules_docker/BUILD:4:1: no such package '@helloworld//image': Pull command failed: 
java.io.IOException: CreateProcess(): %1 is not a valid Win32 application.
 (C:/msys64/tmp/_bazel_jonas.larsson/vs4nb3vm/external/puller/file/puller.par --directory C:/msys64/tmp/_bazel_jonas.larsson/vs4nb3vm/external/helloworld/image --name index.docker.io/library/hello-world:latest) and referenced by '//:hello'.
ERROR: Analysis of target '//:hello.tar' failed; build aborted.
____Elapsed time: 22,834s

It seems to me that the rule code expects the "puller.par" file to be a Win32 executable when it is in fact a bundle of code that needs to be executed with Python? In fact when I call the "puller.par" file with the python executable manually (and with the same parameters as in the output above) it executes. Silently with no output. But it executes.

bazel installed on Ubuntu and MacOS runs this same project just fine.

Any ideas on what my problem is? I've tried it all. No Python installed. Python for 64-bit Windows in the path. pacman installed python2 or python3 in a msys2 bash shell. All with the same result. It does not matter if I run from a Windows CMD prompt or a msys2 mintty bash terminal. The problem remains the same.

[...]

PR #142, opened 2017-09-15
Python all the things

mattmoor:
Switch to explicitly invoking Python when invoking PAR files in WORKSPACE phase.

Switch to consuming `py_binary` targets when invoking Python tooling in BUILD phase.

This is all working towards making rules_docker more portable, but doesn't completely resolve [this](https://github.com/bazelbuild/rules_docker/issues/136)

### DIFF
--- a/docker/pull.bzl
+++ b/docker/pull.bzl
@@ -18,6 +18,19 @@ Bazel rule for downloading base images without a Docker client to
 construct new images with docker_build.
 """
 
+def _python(repository_ctx):
+  if "BAZEL_PYTHON" in repository_ctx.os.environ:
+    return repository_ctx.os.environ.get("BAZEL_PYTHON")
+
+  python_path = repository_ctx.which("python")
+  if not python_path:
+    python_path = repository_ctx.which("python.exe")
+  if python_path:
+    return python_path
+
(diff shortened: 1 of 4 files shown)


mattmoor:
@dslomov I believe this is what we discussed, but feedback is ofc welcome.

/cc @laszlocsomor @meteorcloudy

docker/pull.bzl
```diff
@@ -18,6 +18,19 @@ Bazel rule for downloading base images without a Docker client to
...
+  if "BAZEL_PYTHON" in repository_ctx.os.environ:
+    return repository_ctx.os.environ.get("BAZEL_PYTHON")
+
+  python_path = repository_ctx.which("python")
+  if not python_path:
+    python_path = repository_ctx.which("python.exe")
+  if python_path:
+    return python_path
+
+  fail("rules_docker requires a python interpreter installed. " +
```
dlorenc: Do we actually ever want to use system python? What if that's a python 3 executable, for example?

mattmoor: I believe `repository_ctx.which` will return the system Python.  I think the answer is that we'll blow up (there is another issue on this).

Issue #140, opened 2017-09-14
Labels are not available on the loaded image or during a push. 


hsyed:
For all I know labels are currently not implemented. I'm using the latest commit on os x.

I don't see them when I load the target image into docker, they also do not show up as a parameter to the pusher when I try to pusher.par -- I can't verify the push at the moment (quay.io).

mattmoor:
I see a `labels` kwarg [here](https://github.com/bazelbuild/rules_docker/blob/master/docker/build.bzl#L491-L494), and we test them [here](https://github.com/bazelbuild/rules_docker/blob/master/docker/testdata/BUILD#L214-L222).

I will TAL at these to see if there is a bug.

mattmoor:
You are definitely correct.  If I `bazel run docker/testdata:with_label` and then `docker inspect bazel/docker/testdata:with_label` I get:
```json
[
    {
        "Id": "sha256:83c007425faff33ac421329af9f6444b7250abfc12c28f188b47e97fb715c006",
        "RepoTags": [
            "bazel/docker/testdata:with_label"
        ],
        ...
        "Author": "Bazel",
        "Config": {
            ...
            "Labels": null       <-- BAD
        },
        "RootFS": {
            "Type": "layers",
            "Layers": [
                "sha256:2e79ed5944783867c78cb6870d8b8bb7e68857cbc0894d79119d786d93bc09f7",
                "sha256:84ff92691f909a05b224e1c56abb4864f01b4f8e3c854e4bb4c7baf1d3f6d652"
            ]
        }
    }
]
```

mattmoor:
This is definitely in the v1 json:
```shell
$ tar xOf bazel-bin/docker/testdata/with_label.tar 57438bbe887e61de6b535eda83047c45ee1f98e2bf50071358e617067ccb04aa/json | python -mjson.tool
{
    "architecture": "amd64",
    "config": {
        "Label": [
            "com.example.bar={\"name\": \"blah\"}",
            "com.example.baz=qux",
            "com.example.foo={\"name\": \"blah\"}"
        ],
        "Volumes": {
            "/logs": {}
        }
    },
    "created": "0001-01-01T00:00:00Z",
    "id": "57438bbe887e61de6b535eda83047c45ee1f98e2bf50071358e617067ccb04aa",
    "os": "linux",
    "parent": "5e01cffa06e7ab7f23f1e46fe2e567084477d03e35692f20b6ad0ea2e51fe24e"
}
```

[...]

PR #144, opened 2017-09-16
Fix the Labels field in the image metadata for v2.2

mattmoor:
Fixes: https://github.com/bazelbuild/rules_docker/issues/140

### DIFF
--- a/docker/build_oci_test.sh
+++ b/docker/build_oci_test.sh
@@ -91,7 +91,7 @@ function check_env() {
 function check_label() {
   input="$1"
   shift
-  check_property Label "${input}" "${@}"
+  check_property Labels "${input}" "${@}"
 }
 
 function check_workdir() {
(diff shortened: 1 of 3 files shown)


mattmoor:
This shouldn't be merged yet because it depends on my `google/containerregistry` PR, but I wanted to publish the change for testing, and to unblock @hsyed 

With this change, when I `bazel run docker/testdata:with_label` and `docker inspect bazel/docker/testdata:with_label` I see:
```json
            "Labels": {
                "com.example.bar": "{\"name\": \"blah\"}",
                "com.example.baz": "qux",
                "com.example.foo": "{\"name\": \"blah\"}"
            }
```

So the daemon now recognizes this metadata.

mattmoor:
Ok, the `google/containerregistry` change was merged, I cut a release and pulled it into this change, so it should be RFAL.

Issue #139, opened 2017-09-14
py_image doesn't properly package external deps


mattmoor:
The sample I'm working on runs fine as a `py_binary` target, but when wrapped in `py_image` it produces:
```shell
Traceback (most recent call last):
  File "/app/examples/hellogrpc/py/server/server.binary.runfiles/io_bazel_rules_k8s/examples/hellogrpc/py/server/server.py", line 16,
 in <module>
    import grpc
ImportError: No module named grpc
```

mattmoor:
It looks like this has to do with more of the same from https://github.com/bazelbuild/rules_docker/issues/106

Looking at some of the paths that come through in the `py_image`'s application layer, things from the external deps are being strangely placed.  Here is a single file's full path (clearly wrong):
```shell
./app/examples/hellogrpc/py/server/server.binary.runfiles/ <break long line>
       io_bazel_rules_k8s/examples/hellogrpc/py/ <break long line>
       examples_helloworld_pip_grpcio/grpc/_utilities.py
```

IIUC this should just be:
```shell
./app/examples/hellogrpc/py/server/server.binary.runfiles/ <break long line>
       examples_helloworld_pip_grpcio/grpc/_utilities.py
```

I believe this is because the `.short_path` of this file starts with `../` and it is added in the context of  `io_bazel_rules_k8s/examples/hellogrpc/py/server`, thus resulting in `server/` being stripped and the path above.

Issue #106, opened 2017-08-07
files added from repository rule have impossible paths


mattmoor:
If I add the following to `docker/testdata/BUILD`:
```python
docker_build(
    name = "data_path_image_external",
    files = ["@com_google_guava_guava//jar"],
    mode = "0644",
    data_path = "/",
)
```

It produces an image with:
```shell
$ tar tvf ../../bazel-bin/docker/testdata/data_path_image_external-layer.tar
drwxr-xr-x 0/0               0 1970-01-01 00:00 ./
tar: Removing leading `./..' from member names
drwxr-xr-x 0/0               0 1970-01-01 00:00 ./../
tar: Removing leading `./../' from member names
drwxr-xr-x 0/0               0 1970-01-01 00:00 ./../com_google_guava_guava/
drwxr-xr-x 0/0               0 1970-01-01 00:00 ./../com_google_guava_guava/jar/
-rw-r--r-- 0/0         2256213 1970-01-01 00:00 ./../com_google_guava_guava/jar/guava-18.0.jar
```

mattmoor:
[Related](https://groups.google.com/forum/#!topic/bazel-discuss/1lX3aiTZX3Y)

ulfjack:
The title is a bit confusing. What's impossible about them?

mattmoor:
@ulfjack In the tarballs we build for Docker images, the expansion is implicitly done at `/`, so there is no `..`

mering:
There is still a `magic_path` in https://github.com/bazelbuild/rules_docker/blob/8e70c6bcb584a15a8fd061ea489b933c0ff344ca/container/layer.bzl#L49-L68

This should probably be replaced with something like https://github.com/bazelbuild/rules_docker/blob/8e70c6bcb584a15a8fd061ea489b933c0ff344ca/lang/image.bzl#L108-L109 as done for the language images.

PR #146, opened 2017-09-16
Refactor the way we pass files to the core docker_build for foo_image

mattmoor:
This introduces a new (internal-only for now) argument to the `docker_build` implementation that allows rules to pass a dictionary mapping from the path in the filesystem to a Bazel `File` object to place there.

The common `foo_image` implementation now consumes this in the refactored version to fix how files are being laid out.

I [first noticed](https://github.com/bazelbuild/rules_docker/issues/106) this with `java_image`, which still has this problem, but because we can just construct the `--classpath` argument appropriately it isn't really a correctness issue there (so I've punted on fixing this for now, since it's cosmetic).

With these changes, I am finally able to put together a layered Python GRPC example for `rules_k8s`!

Fixes: https://github.com/bazelbuild/rules_docker/issues/139

### DIFF
--- a/docker/build.bzl
+++ b/docker/build.bzl
@@ -93,7 +93,7 @@ def magic_path(ctx, f):
     # Otherwise, files are added without a directory prefix at all.
     return f.basename
 
-def _build_layer(ctx, files=None, empty_files=None,
+def _build_layer(ctx, files=None, file_map=None, empty_files=None,
                  directory=None, symlinks=None):
   """Build the current layer for appending it the base layer.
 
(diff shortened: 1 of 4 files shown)


damienmg:
You have my preemptive LGTM. I haven't reviewed the code fully yet but I am surprised we don't have that yet :) Could we use the same API as pkg_tar?

mattmoor:
@damienmg It seems like `pkg_tar` reverses the dictionary (Label -> Path), which gives up two properties I want:
1. The ability to add a file multiple places (key conflict)
1. Enforcement that filesystem paths are unique (keys are nice for this, values are not).

Bear in mind, this isn't exposed through `docker_build` at all, just in the implementation function such that the pieces of (e.g.) `py_image) can do interesting things to lay out the `runfiles`.

Issue #148, opened 2017-09-18
Make the rules easily reusable in workflows.


hsyed:
We should be able to reuse the docker rules logic in other rules. Unless I am missing something we can't `bazel run` from within a `bazel run`. A lot of this work will be done for rules_k8s it would be great if the problem is addressed in a general way. 

## Use cases:

* Minikube environment variables injection pre incremental loading bazelbuild/rules_k8s#9.
* Authentication #147 (although this should probably have support in the rules).
* Building CI systems.

## What I tried:

Tried to prototype a higher level rule that would get a "handle" to a docker_build runnable and then do something before or after incrementally loading the image.

```
ci_environment = rule(
  attrs = {
    "platform": attr.label(single_file = True ,executable=True, cfg = "host"),  # <-- docker_build target
  },
  executable=True,
  implementation = _impl
)
```

Somewhere in the rule

```
ctx.file_action(
  output=ctx.outputs.executable,
  content="./%s" % ctx.executable.platform.short_path,
  executable=True
)
```

## Problems
It's not straight forward to setup the runtime context of the target rules. I tried to repurpose the logic by studying the internals and [this](https://github.com/bazelbuild/rules_k8s/blob/master/k8s/object.bzl) but even if I had managed to get it to work it would not be the way forward as it depends on the internals.

## What I hope for

A simple way to reuse the logic in the rules defined here. What would be nicest is a simple handle to the target that I can execute at some arbitrary point in a rule / genrule. If this is not possible some utility functions with examples would be usefull. --e.g.,

```
def _my_rule_impl(ctx):
  info = DefaultInfo()
  run_docker_build(ctx, ctx.attr.my_image, info)
  // my actions
  return info
```

mattmoor:
@hsyed There is clearly a bug today where:
```shell
  # Works
  bazel run path/to:docker_build
  # Does not work
  bazel build path/to:docker_build
  bazel-bin/path/to/docker_build     <-- This is essentially what you want to include into other rules.
```

I bet fixing this would unblock you (if you "hold it" right, which is often harder than it looks with Bazel/Skylark), so I will poke at this a little to see if I can figure out how **I'm** "holding it wrong". :)

For your first two scenarios, I think we have a tentative plan or a solution (I've updated the linked issues).

The CI system bit is a little vague, but I'm happy to discuss (and/or pull folks in to help).


[...]

PR #151, opened 2017-09-18
Make sure we always use an absolute path for ${RUNFILES}

mattmoor:
Related to: https://github.com/bazelbuild/rules_docker/issues/148

### DIFF
--- a/docker/incremental_load.sh.tpl
+++ b/docker/incremental_load.sh.tpl
@@ -18,7 +18,13 @@ set -eu
 
 # This is a generated file that loads all docker layers built by "docker_build".
 
-RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
+function guess_runfiles() {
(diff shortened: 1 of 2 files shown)


docker/incremental_load.sh.tpl
```diff
@@ -18,7 +18,7 @@ set -eu
 
 # This is a generated file that loads all docker layers built by "docker_build".
 
-RUNFILES="${PYTHON_RUNFILES:-${BASH_SOURCE[0]}.runfiles}"
+RUNFILES="${PYTHON_RUNFILES:-${PWD}/${BASH_SOURCE[0]}.runfiles}"
```
damienmg: Humm this might not work. Let say I invoke it by absolute path?

Maybe instead use some kind of magic to get the good path like realpath? maybe simply:

`$(cd ${BASH_SOURCE[0]}.runfiles && pwd -P)`?

mattmoor: I put it in a function, should be RFAL.

mattmoor:
This should be RFAL

Issue #145, opened 2017-09-16
OSS the docker_flatten rule


mattmoor:
Now that the flatten tool is in a `google/containerregistry` release, add the `docker_flatten` rule here.  This is potentially interesting to the effort moving `structure_test` here from the `googlecloudplatform/runtimes-common` repo (e.g. the filesystem tests).

mattmoor:
@nkubala @dlorenc FYI

mattmoor:
Still needs: https://github.com/bazelbuild/rules_docker/pull/144

PR #152, opened 2017-09-18
Upstream the docker_flatten rule.

mattmoor:
This adds a new rule that takes a `docker_build` image and produces a flattened tarball of the filesystem (`:foo.tar`) and extracts the image's metadata (`:foo.json`).

Fixes: https://github.com/bazelbuild/rules_docker/issues/145


mattmoor:
With this it should be possible to write simple filesystem and configuration tests for images, akin to what `structure_test` does (at least for the FS), without requiring a Docker daemon.

@nkubala FYI

PR #162, opened 2017-09-23
Fix the runfiles path of the subcommands of push-all.bzl

mattmoor:
Without this push-all silently fails because "wait" is not checking exit statuses.

I will open an issue to track that, but wanted to fix the breakage from my recent change first.


mattmoor:
@dlorenc Once you've TAL at this I'm going to cut a `.1` release and update docs, since I know a number of folks are using this.

Issue #110, opened 2017-08-09
Support multiple container formats


mattmoor:
I'm opening this issue to discuss options for supporting multiple container formats (OCI, RKT) via some of the core logic in `docker_build`.

@guymers @dlorenc @jonjohnsonjr FYI

mattmoor:
This is something I've been pondering for some time, and so I have some thoughts, which I'll write down to kick things off. :)

At least for OCI vs. Docker v2.2, the differences are currently cosmetic (trivial schema changes and different mime types).  Moreover, the [new intermediate format](https://github.com/bazelbuild/rules_docker/blob/master/intermediate-form.md) for `rules_docker` doesn't know anything about mimes.

So I'm wondering if we want the evolution of things to be...

### container_image

Add support for a generic `container_image` rule, which builds our intermediate format and is agnostic of the final format.
```python
container_image(
    name = "foo",
    base = "@bar//image",
    ...
)
```
We can achieve this by either computing all forms simultaneously (and letting Bazel prune extraneous actions), or using a container agnostic form and projecting it into a particular format on consumption.  We could support form-specific functionality, but consuming that through `container_image` should warn.

These rules couldn't preserve the current `incremental_load.sh.tpl` or `:foo.tar` functionalities without a hint as to which what form the user wants things to default.  I think we could expose a kwarg for the default format, which makes the next part trivial.

I think moving this core definition into `bazelbuild/rules_container` makes complete sense, and we can hollow out `bazelbuild/rules_docker` to proxy to our core implementation.

### {docker,oci,rkt}_image

For clarity / brevity, I think it makes to wrap `container_image` with a default format with a name to reflect that specific form.  This should suppress form-specific warnings (for our form) or turn them into errors (other forms).

### {docker,oci,rkt}_pull

As today, these should pull a particular form into our agnostic intermediate format.  If needed, we can leverage the bridging utilities in `google/containerregistry`.

### {docker,oci,rkt}_push

As today, these should exit our agnostic intermediate format, and publish the image in the particular format.

### Open Questions

* Can we express everything we want via an agnostic intermediate format?
* How would `rules_k8s` (still a prototype) build on top of an agnostic format?
* Do we want/need a `container_bundle` (a la `docker_bundle`)?

guymers:
I like the idea of a generic `container_image` rule that could build different formats. Would it be worth it to maintain specific image rules for each form? Are there any form specific values outside of the OCI image spec that are useful enough to support? 

[...]

PR #171, opened 2017-09-25
Restructure the repo to be less Docker focused.

mattmoor:
This restructures the repository to be less docker-centric.  We may eventually want to move this repo to `@io_bazel_rules_containers`.

The general structure is going to be:
```
  # The bulk of the repo, defining format-agnostic concepts,
  # e.g. container_image, container_push, ...
  container/...

  # Docker-specific aliases (docker_build, docker_push, ...)
  docker/docker.bzl

  # OCI-specific aliases (oci_image, oci_push, ...)
  oci/oci.bzl

  # Shared language .bzl files
  lang/
  # Language-specific rules
  {lang}/

  # Test data
  testdata/...

  # Unspecified-format contributions
  contrib/...
```

Fixes: https://github.com/bazelbuild/rules_docker/issues/110


README.md
```diff
@@ -22,10 +29,10 @@ images.**  This means:
 `boot2docker` or `docker-machine` installed.
 * They do not require root access on your workstation.
 
-Also, unlike traditional Docker builds, the Docker images produced by
-`docker_build` are deterministic / reproducible.
+Also, unlike traditional container build (e.g. Dockerfile), the Docker images
```
jonjohnsonjr: s/build/builds/

README.md
```diff
@@ -76,9 +87,9 @@ your local Docker client by running:
 `docker load -i bazel-genfiles/my/image/helloworld.tar`.  Building
 this target can be expensive for large images.
 
-These work with both `docker_build` and `docker_bundle`.  For
-`docker_build` the image name will be `bazel/my/image:helloworld`.
-For `docker_bundle`, it will apply the tags you have specified.
+These work with both `container_image` and `container_bundle`.  For
+`container_image` the image name will be `bazel/my/image:helloworld`.
```
jonjohnsonjr: Consistency: add comma after `container_image`

README.md
```diff
@@ -1,16 +1,23 @@
-# Bazel Docker Rules
+# Bazel Container Image Rules
```
jonjohnsonjr: Just "Bazel Container Rules" ?

README.md
```diff
@@ -359,11 +376,11 @@ use with `docker_build`'s `base` attribute.
   </tbody>
 </table>
 
-<a name="docker_push"></a>
-## docker_push
+<a name="container_push"></a>
```
jonjohnsonjr: Does this do anything?

mattmoor:
@damienmg @dlorenc FYI.

I may also move the `{cc,java,py,go}_image` rules out of `contrib/`, make them top-level directories, and put together documentation for them.

README.md
```diff
@@ -387,6 +404,14 @@ An executable rule that pushes a Docker image to a Docker registry on `bazel run
       </td>
     </tr>
     <tr>
+      <td><code>format</code></td>
+      <td>
+        <p><code>Kind, required</code></p>
+        <p>The format as which to publish the image.  Currently, this supports
```
jonjohnsonjr: "The format of the published image." or even
"The desired format of the published image."?

README.md
mattmoor: This doesn't do stuff like Kubernetes, so I scoped it to images.

damienmg:
This is a big breaking change, Maybe still offer the ability to load the image with the old name. E.g load("....bzl", docker_build ="container_image") in docker.bzl?

mattmoor:
@damienmg That's exactly what I have in place :)

I also allow folks to load `oci/oci.bzl` and `oci_*`, but these are all aliases (except `oci_push`, which is an alias w/ one default kwarg).

mattmoor:
@damienmg Just to be completely explicit, there are a handful of things that are going to break with this change, which I've convinced myself is an acceptable trade-off:
```
# Contrib loads will fail because all of the files moved, this should be as simple
# as dropping docker/.  EXCEPT for the foo_image rules, which I want to graduate
# into first-class rules here
load("//docker/contrib:*", "*")

# If I circumvented the docker.bzl file, this will break.  Instead use docker.bzl.
load("//docker:build.bzl", "...")
```

I'm not aware of anything else that will break.

mattmoor:
@damienmg Yes, I thought about that, we'll want to think about how we extract that, what we use for OCI, etc, etc.

I've pushed the language `foo_image` changes, so this should be ready for a final look.

mattmoor:
@damienmg If this LGTY then I'll merge things and get started on an import into the mono-repo.  I've retitled to be clear that this should be in a state I expect to merge :)

PR #174, opened 2017-09-27
Implement bazel rule for container diff

priyawadhwa:


bazel-io:

docker/docker_diff.bzl
```diff
@@ -0,0 +1,71 @@
...
+  )
+
+  return struct(runfiles=ctx.runfiles(
+    files = [
+      ctx.executable._container_diff_tool,
+      ctx.executable.image
+    ]),
+  )
+
+_docker_diff = rule(
```
mattmoor: You can just name this `docker_diff` and drop the stuff below.

docker/docker_diff.bzl
```diff
@@ -0,0 +1,71 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
mattmoor: nit: I'd call this file `diff.bzl` to avoid the stutter.

docker/docker_diff.bzl
```diff
@@ -0,0 +1,71 @@
...
+
+def docker_diff(image=None, **kwargs):
+  """Diffs an image in bazel against in image in production .
+
+  This rule runs container-diff on the two images and prints the output.
+
+  Args:
+    name: name of the rule
+    image: bazel target to an image you have bazel built (must be a tar)
+    diff_base: Tag or digest in a remote registry you want to diff against
```
mattmoor: If this supported another local image, we could either feed it through from `container_pull` or use another `BUILD` label, which is nice flexibility.  Does `container-diff` support the same authentication schemes and registries as we currently list in the `README.md`?

docker/docker_diff.bzl
mattmoor: I'd also move this into `contrib/` for now.

docker/docker_diff.bzl
mattmoor: We should also make sure that all the attribute naming is consistent with other rules.

WORKSPACE
```diff
@@ -81,3 +81,9 @@ load(
 )
 
 _go_image_repos()
+
+git_repository(
```
mattmoor: This should not go here.  `diff.bzl` should define `repositories` function that imports Go (which this needs) and this repo.  You should use the same logic as in the other `repositories` functions to avoid reimporting an existing copy of either.

docker/docker_diff.bzl
```diff
@@ -0,0 +1,71 @@
...
+        ),
+    },
+    implementation = _impl,
+    executable = True,
+)
+
+def docker_diff(image=None, **kwargs):
+  """Diffs an image in bazel against in image in production .
+
+  This rule runs container-diff on the two images and prints the output.
```
mattmoor: I believe this is on `bazel run`, which should be stated outright.

priyawadhwa:
Hi @mattmoor ,
Thanks so much for your comments, I'll definitely make the changes you suggested. I think this will actually be moved into the container_diff repo (sorry for the trouble!), and thanks for your help :)
Priya

PR #190, opened 2017-10-12
Change passwd.bzl to support specifying an output_name

tejal29:
Add output_name to passwd.bzl which will write the password file paths to passwd.tar and password.


bazel-io:

tejal29:
adding comment for cla

tejal29:
@googlebot I signed it

tejal29:
@googlebot check again

tejal29:
@dlorenc Can you review this?

contrib/passwd.bzl
```diff
@@ -60,11 +60,12 @@ passwd_file = rule(
             executable = True,
             allow_files = True,
         ),
+        "output_name": attr.string(default = "passwd"),
```
mattmoor: Can we default to `%{name}`?

tejal29: this is changed now.
you can't default to name since attrs is not yet intiliazed and name is not available.

tejal29:
@mattmoor: @dlorenc and me talked about it and its fortunate that this will break other users.  If you know a way to not break users please let me know.
I tried naming the macro same as passwd_file rule, however it fails in bazel.

contrib/passwd.bzl
```diff
@@ -54,17 +39,28 @@ passwd_file = rule(
...
     executable = False,
     outputs = {
-        "out": "%{name}.passwd",
-        "tar": "%{name}.passwd.tar",
+        "out": "%{name}",
     },
     implementation = _impl,
 )
+
+def passwd_tar(name, username, uid, gid, info, home, shell, passwd_file_nm=""):
```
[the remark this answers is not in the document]
tejal29: done.

tejal29:
@mattmoor Can you take another look?

tejal29:
I can create an alias passwd_file which points to passwd_tar and rename the rule passwd_file to something like passwd_file_only ?

contrib/passwd.bzl
```diff
@@ -54,17 +39,28 @@ passwd_file = rule(
...
+    passwd_file(
+        name = file_name,
+        username = username,
+        uid = uid,
+        gid = gid,
+        info = info,
+        home = home,
+        shell = shell,
+    )
+    pkg_tar(
```
mattmoor: I don't think there is a way to make this rule emit a tarball with `/etc/passwd` since `passwd_file`s name can't have a `/` in it...

tejal29:
I made changes to support path for the passwd_file. Now the passwd tar contains the file at specific dir if specified.
`passwd_tar(
    name = "testpasswd_withpath",
    gid = 2345,
    home = "/myhomedir",
    info = "myusernameinfo",
    shell = "/myshell",
    uid = 1234,
    username = "foobar",
    passwd_file_name = "/tmp/passwd",
)`

Will now yield a tar like this:
`tar -tf bazel-bin/testdata/testpasswd_withpath.tar 
./
./tmp/
./tmp/passwd
`

skylib/path.bzl
```diff
@@ -62,3 +62,11 @@ def runfile(ctx, f):
     return ctx.workspace_name + "/" + f.short_path
   else:
     return f.short_path
+
+def filename(path):
```
mattmoor: This is generally called `basename`

contrib/passwd.bzl
```diff
@@ -54,17 +41,32 @@ passwd_file = rule(
...
+        gid = gid,
+        info = info,
+        home = home,
+        shell = shell,
+        output_file = output_file
+    )
+    pkg_tar(
+        name = "%s" %name,
+        package_dir = package_dir,
+        srcs =  [ output_file],
```
mattmoor: strange spacing

contrib/passwd.bzl
```diff
@@ -54,17 +41,32 @@ passwd_file = rule(
...
     executable = False,
     outputs = {
-        "out": "%{name}.passwd",
-        "tar": "%{name}.passwd.tar",
+        "out": "%{output_file}" or "%{name}",
     },
     implementation = _impl,
 )
+
+def passwd_tar(name, username, uid, gid, info, home, shell, passwd_file_name=""):
```
mattmoor: Do we have a use case where the tarball contains anything other than `/etc/passwd`?

tejal29: not that i know of. I can always default it to "/etc/passwd"

skylib/path.bzl
tejal29: fixed

contrib/passwd.bzl
tejal29: fixed

contrib/passwd.bzl
```diff
@@ -54,17 +41,32 @@ passwd_file = rule(
...
-            executable = True,
-            allow_files = True,
-        ),
+        "output_file": attr.string(default = ""),
     },
     executable = False,
     outputs = {
-        "out": "%{name}.passwd",
-        "tar": "%{name}.passwd.tar",
+        "out": "%{output_file}" or "%{name}",
```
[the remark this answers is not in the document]
tejal29: I just verified by writing tests it does not work :) removed it.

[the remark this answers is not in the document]
damienmg: /cc @dslomov @laurentlb for Skylark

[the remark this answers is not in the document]
laurentlb: "${output_file}" is expanded later, so this cannot work.
If you need more flexibility, use a function that returns a dictionary (https://docs.bazel.build/versions/master/skylark/lib/globals.html#rule).

container/image_test.py
```diff
@@ -281,16 +281,28 @@ def test_build_with_tag(self):
...
       buf = cStringIO.StringIO(img.blob(img.fs_layers()[0]))
       with tarfile.open(fileobj=buf, mode='r') as layer:
-        content = layer.extractfile('./etc/passwd').read()
+        content = layer.extractfile('./tmp/passwd').read()
         self.assertEqual(
           'foobar:x:1234:2345:myusernameinfo:/myhomedir:/myshell\n', content)
 
+  def test_with_passwddefault(self):
+    with TestImage('with_passwddefault') as img:
+      self.assertDigest(img, '7de06a352108df247a4b5a1cea2eb3e80ee39ddc1192decbca8038427b4a2da6')
```
mattmoor: This matches the digest above, but has different content, so I'm pretty sure (without looking) that this is a reason your tests are failing.  I would expect at least one of these tests to match the digest on the LHS: `9fd1e04...`.

contrib/passwd.bzl
```diff
@@ -54,17 +41,28 @@ passwd_file = rule(
...
     executable = False,
     outputs = {
-        "out": "%{name}.passwd",
-        "tar": "%{name}.passwd.tar",
+        "out": "%{name}",
     },
     implementation = _impl,
 )
+
+def passwd_tar(name, username, uid, gid, info, home, shell, passwd_file_name="/etc/passwd"):
```
mattmoor: At this point, I think it's cleaner to just have the user generate the tarball vs. understanding how `passwd_file_name` pollutes the label space with `basename(passwd_file_name)`.

contrib/passwd.bzl
tejal29: the pkg_tar package_dir attribute makes it possible.
You can check my tests to confirm.

tejal29:
@googlebot I signed it

contrib/passwd.bzl
tejal29: umm ok

tejal29:
Closing in favor of https://github.com/bazelbuild/rules_docker/pull/205

PR #191, opened 2017-10-12
Add a container_archive repository rule

siddharthab:
This rule generates a container_import target for `docker save` tarballs.

### DIFF
--- a/container/archive.bzl
+++ b/container/archive.bzl
@@ -16,70 +16,14 @@
 This extracts the tarball, examines the layers and creates a
 container_import target for use with container_image.
 """
-
-def _manifest_layers(manifest):
-  # Extracts layers, preserving order, from the manifest.
-
-  layers_begin_marker = "\"Layers\":["
-  start = manifest.index(layers_begin_marker)
-  end = manifest.index("]", start)
-
-  start += len(layers_begin_marker)
-  layers = manifest[start:end].replace("\"", "").split(",")
-
-  if not layers:
-    fail("Failed to extract layers from manifest.json")
-
-  return layers
-
-def _gzip_layers(ctx):
-  # Returns an array of gzipped layers, preserving order.
-
-  manifest = ctx.execute(["cat", "manifest.json"])
-  if manifest.return_code:
-    fail("Could not read manifest: %s" % manifest.stderr)
-
-  zipped_layers = []
-  layers = _manifest_layers(manifest.stdout)
-  for layer in layers:
-    zipped_layer = layer.split("/")[0] + ".tar.gz"
-    result = ctx.execute(["gzip", "-n", layer])
-    if result.return_code:
-      fail("Failed to gzip image layer %s: %s" % (layer, result.stderr))
-    result = ctx.execute(["mv", layer + ".gz", zipped_layer])
-    if result.return_code:
-      fail("Failed to gzip image layer %s: %s" % (layer, result.stderr))
-    zipped_layers += ["\"" + zipped_layer + "\","]
-
-  return zipped_layers
-
-def _config_json(ctx):
-  # Identifies the image config json file in the root directory.
-
-  config_candidates = ctx.execute(["ls", "-1"])
-  if config_candidates.return_code:
-    fail("Failed to list contents: %s" % config_candidates.stderr)
-
-  config = None
-  for candidate in config_candidates.stdout.splitlines():
-    if candidate.endswith(".json") and candidate != "manifest.json":
-      config = candidate
-      break
-
-  if not config:
-    fail("Failed to find config file")
-
-  return config
-
 def _container_archive_impl(ctx):
-  ctx.download_and_extract(url=ctx.attr.urls,
-                           sha256=ctx.attr.sha256,
-                           type=ctx.attr.type,
-                           stripPrefix=ctx.attr.strip_prefix)
+  result = ctx.execute([
+      ctx.path(ctx.attr._importer),
+      "--directory", ".",
+      "--tarball", ctx.path(ctx.attr.file)])
 
-  zipped_layers = _gzip_layers(ctx)
-
-  config = _config_json(ctx)
+  if result.return_code:
+    fail("Importing from tarball failed (status %s): %s" % (result.return_code, result.stderr))
 
   ctx.file("BUILD", """
 package(default_visibility = ["//visibility:public"])
@@ -88,23 +32,25 @@ load("@io_bazel_rules_docker//container:import.bzl", "container_import")
 
 container_import(
   name = \"""" + ctx.attr.image_tag + """\",
-  config = \"""" + config + """\",
-  layers = [
-    """ + "\n    ".join(zipped_layers) + """
-  ],
+  config = "config.json",
+  layers = glob(["*.tar.gz"]),
   repository = \"""" + ctx.attr.image_repository + """\",
 )
 """, executable=False)
-    
 
 container_archive = repository_rule(
     attrs = {
-        "urls": attr.string_list(allow_empty = False),
-        "sha256": attr.string(),
-        "type": attr.string(),
-        "strip_prefix": attr.string(),
+        "file": attr.label(
+            allow_single_file = True,
+            mandatory = True,
+        ),
         "image_repository": attr.string(default = "bazel"),
         "image_tag": attr.string(default = "image"),
+        "_importer": attr.label(
+            executable = True,
+            default = Label("@importer//file:importer.par"),
+            cfg = "host",
+        ),
     },
     implementation = _container_archive_impl,
 )
--- a/container/archive.bzl
+++ b/container/archive.bzl
@@ -1,0 +1,110 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Rule for importing an image from 'docker save' tarballs.
+
+This extracts the tarball, examines the layers and creates a
+container_import target for use with container_image.
+"""
+
+def _manifest_layers(manifest):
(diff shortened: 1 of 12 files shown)


bazel-io:

container/archive.bzl
```diff
@@ -0,0 +1,110 @@
...
+  ],
+  repository = \"""" + ctx.attr.image_repository + """\",
+)
+""", executable=False)
+    
+
+container_archive = repository_rule(
+    attrs = {
+        "urls": attr.string_list(allow_empty = False),
+        "sha256": attr.string(),
```
mattmoor: I think that if folks are going to download the tarball, I'd rather they write:
```python
http_file(
    name = "foo",
    ...
)

# This is about the extent of what I'd like the signature to be.
container_archive(
    name = "bar",
    file = "@foo//file",
)
```

docker/docker.bzl
```diff
@@ -20,6 +20,7 @@ load(
     docker_bundle = "container_bundle",
     docker_flatten = "container_flatten",
     docker_image = "container_image",
+    docker_archive = "container_archive",
```
mattmoor: We'll want it under `oci/oci.bzl` too.

container/archive.bzl
```diff
@@ -0,0 +1,110 @@
...
+  config = \"""" + config + """\",
+  layers = [
+    """ + "\n    ".join(zipped_layers) + """
+  ],
+  repository = \"""" + ctx.attr.image_repository + """\",
+)
+""", executable=False)
+    
+
+container_archive = repository_rule(
```
mattmoor: Given the analog to `docker load` I think I'd rename: `s/archive/load/g`

container/archive.bzl
```diff
@@ -0,0 +1,110 @@
...
+
+def _container_archive_impl(ctx):
+  ctx.download_and_extract(url=ctx.attr.urls,
+                           sha256=ctx.attr.sha256,
+                           type=ctx.attr.type,
+                           stripPrefix=ctx.attr.strip_prefix)
+
+  zipped_layers = _gzip_layers(ctx)
+
+  config = _config_json(ctx)
```
mattmoor: Turns out [I wrote a tool for this](https://github.com/google/containerregistry/blob/master/tools/fast_importer_.py#L46), but didn't add a `cloudbuild.yaml` to release the `PAR` file.

I'll do that, setup the build, and cut a release so we can pick that up.

mattmoor:
Here's the PR bringing in the importer: https://github.com/bazelbuild/rules_docker/pull/192

docker/docker.bzl
siddharthab: Done.

container/archive.bzl
siddharthab: I removed the http archive options and replaced them with a single file.

I think people would also want a way to specify the repository and the name/tag of the image target that is generated. Without this, all images imported like this will be `bazel:image` if loaded in docker.
So I have kept those attributes. If you feel strongly about removing them, I can take them out.

container/archive.bzl
siddharthab: IMO, it is more of an opposite of docker load actually.

Taking an image out of docker and bringing it into the intermediate form. 'load' will be confusing. I am not happy with 'archive' either but can not think of a better term.

container/archive.bzl
siddharthab: Looking through it now. Will update this PR when I am done.

docker/docker.bzl
siddharthab: Sorry pressed by accident. It will be done in my next commit.

container/archive.bzl
siddharthab: I tried it out. This PR is now using your tool. Thank you for pointing me to it and sending in the PR to include it.

However, I did notice that the tool is much slower than simply using gzip. For example, without using the importer tool, my image is imported in ~30s, but with the importer it takes ~5m30s. I am running on a Mac with 4 cores, and I saw that the importer was using 90%-120% CPU for when it was running.

I thought that the importer is using better compression, so I compared the sizes and the zipped layers are ~1% smaller with the importer than with gzip using default options.

Since the images, once imported, will be cached, perhaps it is not important but I thought I should let you know.

The code for trying out with command line gzip is at https://github.com/siddharthab/rules_docker/tree/container_archive

mattmoor: Ack slowness, Python gzip is known to be slow, but this should be more portable than shelling out.  If the speed proves to be a problem, then we could augment `container_import` to take the uncompressed blobs through an alternate attribute and change the importer to emit those instead (should be faster since it is stored uncompressed).

container/archive.bzl
```diff
@@ -0,0 +1,57 @@
...
+  if result.return_code:
+    fail("Importing from tarball failed (status %s): %s" % (result.return_code, result.stderr))
+
+  ctx.file("BUILD", """
+package(default_visibility = ["//visibility:public"])
+
+load("@io_bazel_rules_docker//container:import.bzl", "container_import")
+
+container_import(
+  name = \"""" + ctx.attr.image_tag + """\",
```
mattmoor: We should shoot for a similar signature to `@pulled//image`, which means we emit things into an `image` directory and make the `BUILD` target `:image`.

container/archive.bzl
```diff
@@ -0,0 +1,57 @@
...
+  config = "config.json",
+  layers = glob(["*.tar.gz"]),
+  repository = \"""" + ctx.attr.image_repository + """\",
```
mattmoor: This isn't a [supported attribute](https://github.com/bazelbuild/rules_docker#container_import), i'm surprised it works.

If users want to load these images with another name, they can go through `container_bundle`, as we offer for `container_image` in `contrib/with-tag.bzl`

mattmoor:
Test this please  (for Bazel CI)

mattmoor:
Oh, before I forget, two more things:
1. we need to add this to the README
1. it would be great to have a test, we should have a tarball checked in already that you can use (look in `testdata/`)

container/load.bzl
```diff
@@ -0,0 +1,58 @@
...
+"""Rule for importing an image from 'docker save' tarballs.
+
+This extracts the tarball, examines the layers and creates a
+container_import target for use with container_image.
+"""
+
+def _container_load_impl(ctx):
+  result = ctx.execute(["mkdir", "image"])
+  if result.return_code:
+    fail("Creating directory failed (status %s): %s" % (result.return_code, result.stderr))
```
mattmoor: IIRC `container_pull` doesn't do this, why did you need to?

container/load.bzl
```diff
@@ -0,0 +1,58 @@
...
+
+  result = ctx.execute([
+      ctx.path(ctx.attr._importer),
+      "--directory", "image",
```
mattmoor: perhaps you could avoid the above with `ctx.path("image")` here?

container/load.bzl
```diff
@@ -0,0 +1,58 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Rule for importing an image from 'docker save' tarballs.
+
+This extracts the tarball, examines the layers and creates a
+container_import target for use with container_image.
+"""
+
+def _container_load_impl(ctx):
```
mattmoor: Sorry for not noticing this before, but typically in the context of `repository_rule` `ctx` would be `repository_ctx` (they have different signatures, so this helps clarify).

container/load.bzl
```diff
@@ -0,0 +1,58 @@
...
+
+  result = ctx.execute([
+      ctx.path(ctx.attr._importer),
+      "--directory", "image",
+      "--tarball", ctx.path(ctx.attr.file)])
+
+  if result.return_code:
+    fail("Importing from tarball failed (status %s): %s" % (result.return_code, result.stderr))
+
+  ctx.file("image/BUILD", """
```
mattmoor: Peeking at [pull.bzl](https://github.com/bazelbuild/rules_docker/blob/master/container/pull.bzl#L38) it looks like you may need an empty `BUILD` file in the root directory as well.

container/load.bzl
```diff
@@ -0,0 +1,58 @@
...
+container_import target for use with container_image.
+"""
+
+def _container_load_impl(ctx):
+  result = ctx.execute(["mkdir", "image"])
+  if result.return_code:
+    fail("Creating directory failed (status %s): %s" % (result.return_code, result.stderr))
+
+  result = ctx.execute([
+      ctx.path(ctx.attr._importer),
```
mattmoor: For portability to Windows we need to prefix `PAR` files with the `python` path.  That's what [this](https://github.com/bazelbuild/rules_docker/blob/master/container/pull.bzl#L21) is doing in `container_pull`.

Sadly we haven't reached the point where this is on our Bazel CI config, but I'd like to avoid regressing here.  You can rename this (drop leading underscore) and then import it from `pull.bzl`.  I can worry about finding it a better home later.

testing/e2e.sh
```diff
@@ -64,6 +64,11 @@ docker_build(
   base = "@pause//image",
   workdir = "/tmp",
 )
+
+docker_build(
```
mattmoor: So this file is generally for testing stuff we can't easily test within Bazel, e.g. the behavior of `bazel run :foo` etc.  You should be able to `container_load(name="...", file="//testdata:pause.tar")` from our `WORKSPACE`, do something with it in `testdata/BUILD`, incorporate the new image into `container/BUILD`s `TEST_DATA` and then inspect the result in `container/image_test.py`.

Sorry for the confusion around the different kinds of testing :(

mattmoor:
Please test this

mattmoor:
Test this please

container/load.bzl
siddharthab: Done.

container/load.bzl
siddharthab: I had to because the importer tool expects the directory to exist already.

container/load.bzl
siddharthab: Done.

container/load.bzl
siddharthab: Done.

container/load.bzl
siddharthab: You are right. Done.

testing/e2e.sh
siddharthab: Done. Thanks for the instructions.

mattmoor:
It looks like buildifier is upset:
```shell
The command "buildifier -mode=check $(find . -name '*.bzl' -type f)" exited with 4.
```

mattmoor:
Test this please

@damienmg Do we have to approve every commit?

mattmoor:
Test this please

PR #194, opened 2017-10-13
Add dockerfile_build repository rule

damienmg:
This repository rule allow to build a dockerfile image and import
it as a docker_import rule.

### DIFF
--- a/contrib/convert_image_to_build.py
+++ b/contrib/convert_image_to_build.py
@@ -1,0 +1,60 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Convert a docker save JSON files to a BUILD file so docker_import can read it."""
+
+import gzip
+import json
+import os
+import os.path
+import shutil
+import sys
+import tarfile
+
+def convert_manifest(infile, outfile):
+  with open(infile, "r") as fp:
+    json_obj =  json.load(fp)
+    dirname = os.path.dirname(infile)
+    dir = os.path.relpath(dirname, os.path.dirname(outfile))
+    config = json_obj[0]["Config"]
+    layers = json_obj[0]["Layers"]
+    with open(outfile, "w") as of:
+      of.write("""
+load("@io_bazel_rules_docker//docker:docker.bzl", "docker_import")
+
+docker_import(
+    name = "image",
+    layers = [%s],
+    config = "%s/%s",
+    visibility = ["//visibility:public"],
+)
+""" % (",".join(['"%s/%s.tgz"' % (dir, os.path.dirname(l)) for l in layers]), dir, config))
+
+def extract_image(image, directory):
+  with tarfile.open(image) as tar:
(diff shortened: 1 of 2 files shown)


damienmg:
I only copied paste the rules from bazelbuild/continuous-integration, so I have 2 questions:

 - Do you want a quick test where we create an image using it?
 - Should this be documented elsewhere or just the skydoc is enough?

damienmg:
Also /cc @siddharthab because I don't know how much we can share with container_load rule.

mattmoor:
Adding @dlorenc who may have thoughts here.

mattmoor:
@damienmg I do want a test, but I'm happy to punt on docs until we document the `contrib/` directory :-/

contrib/dockerfile_build.bzl
```diff
@@ -0,0 +1,65 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Repository rule to build image with `docker build`."""
+
+def _impl(rctx):
+  tag = rctx.attr.tag
+  dockerfile = rctx.path(rctx.attr.dockerfile)
+  result = rctx.execute([
+      "docker",
```
dlorenc: Docker builds can run into trouble with build contexts that contain symlinks. In the past when doing things like this I've had to "resolve" symlinks by copying to a tempdir before doing doing the docker build. Does this cause any trouble here?

contrib/convert_image_to_build.py
```diff
@@ -0,0 +1,60 @@
...
+
+import gzip
+import json
+import os
+import os.path
+import shutil
+import sys
+import tarfile
+
+def convert_manifest(infile, outfile):
```
dlorenc: I think docker builds are not inherently reproducible, each time you do a docker build (with a clean cache), it will insert some timestamps into the manifest/config. It might be worth stripping those out here.

mattmoor: This should go away completely in favor of `container_load`.  This is an interesting twist for the `WORKSPACE` discussion: how do you "lock" this kind of rule? :)

dlorenc:
Also I just realized this is a repository rule... Can you explain why? I imagined this as a normal build rule.

dlorenc:
Just thought of a use-case we have where a build version of this would be a lot simpler. We have a base image we're building with bazel, from scratch. Then we need to install some things into that via a Dockerfile.

A mode where you could specify the base image as a target, docker load that, and have the FROM line of the Dockerfile templated out to the docker loaded name before a build would be really useful.

dlorenc:
FYI: This is a prototype on how I was imagining this rule:
https://github.com/GoogleCloudPlatform/debian-docker/pull/105

damienmg:
(testing CI: retest this please)

mattmoor:
So the Databricks guys had a neat take on this, which I actually like because it addresses the reproducibility problem.

@mgyucht FYI (I don't have Ahir's GitHub handle)

mgyucht:
It's @ahirreddy :) I'll take a look at this when I have a moment!

contrib/dockerfile_build.bzl
```diff
@@ -0,0 +1,65 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Repository rule to build image with `docker build`."""
+
+def _impl(rctx):
+  tag = rctx.attr.tag
+  dockerfile = rctx.path(rctx.attr.dockerfile)
+  result = rctx.execute([
```
tejal29: This function is not documented in here https://docs.bazel.build/versions/master/skylark/lib/ctx.html.
I am assuming this executes a command as a subprocess?

pcj: Drive-by comment: object is a repository_ctx object, not "regular" context object.

https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#execute

damienmg:
I am leaving the bazel team so I won't continue that change.

PR #195, opened 2017-10-13
Allow uncompressed layers in container_import

siddharthab:
This allows to let the rule decide if it needs the layers compressed or not. If one form is preferred over the other, it can be documented as such.

An added bonus is that the container_load rule can be made faster by passing on uncompressed layers instead of using python's gzip library. For now, the layers will still be compressed but using the faster gzip from shell.

### DIFF
--- a/container/import.bzl
+++ b/container/import.bzl
@@ -42,15 +42,25 @@ load(
     _join_path = "join",
 )
 
-def _layer_pair(ctx, layer):
-  zipped = False
-  for filetype in tgz_filetype:
-    if layer.basename.endswith(filetype):
-      zipped = True
-      break
+def _is_filetype(filename, extensions):
+  for filetype in extensions:
+    if filename.endswith(filetype):
+      return True
 
-  zipped_layer = _gzip(ctx, layer) if not zipped else layer
-  unzipped_layer = _gunzip(ctx, layer) if zipped else layer
+def _is_tgz(layer):
+  return _is_filetype(layer.basename, tgz_filetype)
+
+def _is_tar(layer):
+  return _is_filetype(layer.basename, tar_filetype)
+
+def _layer_pair(ctx, layer):
+  zipped = _is_tgz(layer)
+  unzipped = not zipped and _is_tar(layer)
+  if not (zipped or unzipped):
+    fail("Unknown filetype provided (need .tar or .tar.gz): %s" % layer)
+
+  zipped_layer = layer if zipped else _gzip(ctx, layer)
+  unzipped_layer = layer if unzipped else _gunzip(ctx, layer)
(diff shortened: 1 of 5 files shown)


bazel-io:

container/import.bzl
```diff
@@ -102,7 +113,7 @@ def _container_import_impl(ctx):
 container_import = rule(
     attrs = {
         "config": attr.label(allow_files = [".json"]),
-        "layers": attr.label_list(allow_files = tgz_filetype),
+        "layers": attr.label_list(allow_files = tar_filetype),
```
mattmoor: I think this should be `[tar_filetype, tgz_filetype]`

container/import.bzl
```diff
@@ -40,9 +42,16 @@ load(
     _join_path = "join",
 )
 
-def _unzip_layer(ctx, zipped_layer):
-  unzipped_layer = _gunzip(ctx, zipped_layer)
-  return unzipped_layer, _sha256(ctx, unzipped_layer)
+def _layer_pair(ctx, layer):
+  zipped = False
+  for filetype in tgz_filetype:
```
mattmoor: Can you hoist a function for each of `is_tgz` and `is_tar` (I think it will read a bit easier)

I'd kind of like to check that it matches one or the other and fail if neither instead of silently doing the wrong thing below.

container/import.bzl
```diff
@@ -40,9 +42,16 @@ load(
...
+    if layer.basename.endswith(filetype):
+      zipped = True
+      break
+
+  zipped_layer = _gzip(ctx, layer) if not zipped else layer
```
mattmoor: I think it reads slightly better as:
```python
zipped_layer = layer if zipped else _gzip_layer(ctx, layer)
unzipped_layer = layer if unzipped else _gunzip_layer(ctx, layer)
```

Even the second I think reads better as `not zipped`, since we've made the predicates match the sense of the target variable.

Total nit.

container/import.bzl
siddharthab: Done. Although the check should also be enforced by allow_files in the rule.

Note that tar filetypes are a superset of tgz filetypes.

container/import.bzl
siddharthab: Done.

container/import.bzl
siddharthab: Done.

mattmoor:
Test this please

siddharthab:
Thanks for the review.

For the record, I noticed that depending on the incremental load (default target) does not compress unzipped input layers, only building the .tar output will invoke gzip. So I guess it worked better than I expected.

mattmoor:
Yeah Bazel prunes stuff that isn't needed :)

PR #198, opened 2017-10-19
expose more attributes in container_image rule

erain:
* Add `env`, `debs` and `output` as extra args to container_image rule's
implementation.

* These args are neccessary so that other extensions of docker rules can
use them.


erain:
@nlopezgi  Please help review this PR.

erain:
Looks like the changes include more than the extension rule need. Close it for now.

PR #212, opened 2017-11-02
use dict.update instead of deprecated + operator

kevingessner:
The + operator is being removed for dicts:
https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#dictionary-concatenation

### DIFF
--- a/container/push.bzl
+++ b/container/push.bzl
@@ -83,39 +83,41 @@ def _impl(ctx):
   stamp_inputs + ([image["legacy"]] if image.get("legacy") else []) +
   list(ctx.attr._pusher.default_runfiles.files)))
 
+attrs = {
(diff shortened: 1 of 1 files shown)


bazel-io:

mattmoor:
Thanks for the PR, I guess we missed one in the recent update.

Can you leave the `attr` definition inlined for consistency with other rules?  Also you need to run buildifier (Why Travis is upset).

kevingessner:
I don't see another uses of `dict.update` around.  Unfortunately `update` is a void method, so it can't be inlined (unless there's some trick i don't see).

kevingessner:
ah i see, this is why it was using .items(). I didn't realize that got around the problem. So this isn't needed then.

PR #214, opened 2017-11-03
Use Skylib's `skylark_library` for rules.

Capstan:
This only addresses docker directory and its dependencies: the container and skylib directories.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -13,6 +13,14 @@
 # limitations under the License.
 workspace(name = "io_bazel_rules_docker")
 
+# While this exists as bazel_tools, that version is tied to the installed
+# bazel tool rather than a specific source version.
+git_repository(
+    name = "io_bazel",
+    remote = "https://github.com/bazelbuild/bazel.git",
+    commit = "97f0290cc1197311d60f45d49f7c52b70f879a18",
+)
+
 load(
     "//container:container.bzl",
     "container_pull",
@@ -135,3 +143,10 @@ git_repository(
 load("@io_bazel_rules_d//d:d.bzl", "d_repositories")
 
 d_repositories()
+
+# For skylib_library.
+git_repository(
+    name = "bazel_skylib",
+    remote = "https://github.com/bazelbuild/bazel-skylib.git",
+    tag = "0.2.0",
+)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -13,14 +13,6 @@
 # limitations under the License.
 workspace(name = "io_bazel_rules_docker")
 
-# While this exists as bazel_tools, that version is tied to the installed
-# bazel tool rather than a specific source version.
-git_repository(
-    name = "io_bazel",
-    commit = "97f0290cc1197311d60f45d49f7c52b70f879a18",
-    remote = "https://github.com/bazelbuild/bazel.git",
-)
-
 load(
     "//container:container.bzl",
     "container_pull",
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -143,10 +143,3 @@ git_repository(
 load("@io_bazel_rules_d//d:d.bzl", "d_repositories")
 
 d_repositories()
-
-# For skylib_library.
-git_repository(
-    name = "bazel_skylib",
-    remote = "https://github.com/bazelbuild/bazel-skylib.git",
-    tag = "0.2.0",
-)
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -17,8 +17,8 @@ workspace(name = "io_bazel_rules_docker")
 # bazel tool rather than a specific source version.
 git_repository(
     name = "io_bazel",
-    remote = "https://github.com/bazelbuild/bazel.git",
     commit = "97f0290cc1197311d60f45d49f7c52b70f879a18",
+    remote = "https://github.com/bazelbuild/bazel.git",
 )
 
 load(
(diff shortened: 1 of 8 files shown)


mattmoor:
Travis is unhappy:
```shell
The command "buildifier -mode=check $(find . -name WORKSPACE -type f)" exited with 4.
```

WORKSPACE
```diff
@@ -135,3 +143,10 @@ git_repository(
 load("@io_bazel_rules_d//d:d.bzl", "d_repositories")
 
 d_repositories()
+
+# For skylib_library.
+git_repository(
```
mattmoor: I think dependents will need these dependencies, so can you add them to the `repositories()` function in `container/container.bzl`?

Capstan:
`buildifier` wanted a slightly different order in the `WORKSPACE` file. Fixed.

WORKSPACE
Capstan: I can do `bazel_skylib`, but apparently trying to do that with `io_bazel` ends up with a chicken-and-egg problem: 

```
ERROR: Failed to load Skylark extension '@io_bazel//tools/build_defs/hash:hash.bzl'.
It usually happens when the repository is not defined prior to being used.
Maybe repository 'io_bazel' was defined later in your WORKSPACE file?
```

Since `WORKSPACE` loads `container.bzl` and that loads `container/image.bzl` and that refers to `io_bazel`, I'm not sure how to extract that nicely into repositories only and structural defines separately. Would you want a `container_repos.bzl` that was specific to `WORKSPACE` includes? This model kinda sucks. I would think it'd want to at least warn you that you were asking for a different (esp. if later) tag or commit instead of the "I just hope somebody includes it" model.

Capstan: Alright, I have some kludges that do this in https://github.com/Capstan/rules_docker/commit/ec938fa1b9bbac998581fa8e3a0b06b7287b4090 It makes there be a separation between Skylark to establish repos and Skylark for rules. This allows the latter to refer to other repositories but ensures that all of those repository names have been defined first. A top-level WORKSPACE must load and invoke the repos macro before trying to load other Skylark in that project.

I'm not sure I like it better, but it works (well, it gets further along in e3e). Also, `/bin/true` doesn't exist by default on an Xcode-laden Mac (it's `/usr/bin/true`), and I'm not sure yet what else doesn't work when locally running e2e test on a Mac.

mattmoor:
@Capstan IIUC, if we wait until a sufficiently modern Bazel is released, and start to require that version, we could avoid the breaking change?

If so, can we determine when said change would land in a Bazel release?  /cc @damienmg 

I think my preference would be to force users upgrading `rules_docker` with older Bazel versions to upgrade Bazel too vs. breaking everyone that upgrades.  :(

If we're talking about O(weeks), then I think I'd rather we wait on merging the PR (and perhaps merge the change directly into the mono-repo to unblock you).

Hopefully that all makes sense?  WDYT?

damienmg:
It will be in 0.8 which is currently in RC so should be released in roughly
10 days.

Capstan:
@mattmoor Yes, a modern Bazel would also work here. It doesn't ultimately solve the problem that upstream Bazel source changes will not be applied to rules_docker identically: two developers with different Bazel versions with different `hash.bzl` files may build rules_docker differently. It's a pretty narrow boundary case, so perhaps find to leave as a known issue. Or, you could institute a "you must build this with this version of Bazel" as something analogous to importing a specific tag/commit at the`WORKSPACE` level.

@damienmg Ooh, that's good to hear. I thought my commit was going to be after the cutoff for 0.8. Too bad we don't have labels to show "commit first released in *release*" I can wait 10 days. though this is starting to block forward progress on my end.

Capstan:
I manually installed 0.8-rc2 and moved build references back to build_tools and it worked. Additionally, I made travis point at bazel testing branch so that it should at least validate that e2e tests will run with the release is live.

Capstan:
@mattmoor SGTM.

Capstan:
Doh, inadvertently closed it.

bazel-io:

Capstan:
Now that 0.8 has been released, I've backed out the `.travis.yml` change to use the testing release instead of the stable release.

mattmoor:
Test this please

mattmoor:
Please test this

mattmoor:
Please test this

mattmoor:
Test this please

Capstan:
@mattmoor I think you launched this before Chocolatey was updated (https://github.com/bazelbuild/bazel/issues/4196#issuecomment-349329168), and so got 1.8.0 anyway.

mattmoor:
Bazel CI is green.  I'm going to close / reopen to retrigger everything now.

bazel-io:

mattmoor:
Test this please

mattmoor:
@Capstan Looks like all that's left is buildifier :)

mattmoor:
Test this please

mattmoor:
I tried to "Resolve conflicts" through the web editor, but that seems to have done nothing but retrigger CI...  I fixed the `buildifier` issues separately (sorry for the churn), so this should be good to go after a rebase.

I am going to work through the various other PRs that have accumulated (I'm assuming you aren't still online), but ping me on hangouts if you want to take a lock to avoid going back and forth with merge conflicts :)

mattmoor:
Test this please

mattmoor:
@Capstan thanks for your patience and persistence :)

PR #216, opened 2017-11-03
Apply lang image visibility to underlying image targets.

BlueDragonX:
It is currently not possible to reference a language specific image target from another package even with the correct `visibility` set. The cause of this is that the language specific image macros do not pass the visibility attribute to the generated image targets. This PR passes visibility down to the image targets.

### DIFF
--- a/python/image.bzl
+++ b/python/image.bzl
@@ -87,5 +87,6 @@ def py_image(name, base=None, deps=[], layers=[], **kwargs):
     base = this_name
     index += 1
 
+  visibility = kwargs.get('visibility', None)
   app_layer(name=name, base=base, entrypoint=['/usr/bin/python'],
-            binary=binary_name, layers=layers)
+            binary=binary_name, layers=layers, visibility=visibility)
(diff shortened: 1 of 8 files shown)


bazel-io:

mattmoor:
I think you need to plumb the kwarg further, but I'm cool with this change if you can get it working.  thanks for the PR!

BlueDragonX:
I wasn't sure if anything else would be useful to pass down to the `container_image`, but this fixes our immediate problem of pushing images from elsewhere in the package tree. I've tested it against `go_image` and `scala_image` so far.

python/image.bzl
```diff
@@ -88,4 +88,4 @@ def py_image(name, base=None, deps=[], layers=[], **kwargs):
     index += 1
 
   app_layer(name=name, base=base, entrypoint=['/usr/bin/python'],
-            binary=binary_name, layers=layers)
+            binary=binary_name, layers=layers, visibility=visibility)
```
mattmoor: This is missing a definition of `visibility`.

BlueDragonX: @mattmoor fixed.

mattmoor:
Test this please

PR #222, opened 2017-11-08
Add example for the Docker Hub

drigz:
It's not at all obvious what `repository` should be set to for the
Docker Hub. Hopefully this example helps people if they need it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -630,6 +630,15 @@ container_pull(
 ```
 
 This can then be referenced in `BUILD` files as `@base//image`.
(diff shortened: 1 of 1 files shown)


mattmoor:
I've been very resistant to special case DockerHub as they do in the Docker CLI, but perhaps it would make sense to expose under `contrib/` a `dockerhub_pull` that points at DockerHub and special cases single-word `repositories` as the client does?

drigz:
Ah, nice :) Not sure how @ensonic and I both missed that!

PR #230, opened 2017-11-13
allow override of `tars` for docker_build extensions

erain:
We need to override `tars` attr in addition to `debs` attr.

### DIFF
--- a/container/image.bzl
+++ b/container/image.bzl
@@ -114,8 +114,8 @@ def _build_layer(ctx, files=None, file_map=None, empty_files=None,
   args += ["--file=%s=%s" % (f.path, magic_path(ctx, f)) for f in files]
   args += ["--file=%s=%s" % (f.path, path) for (path, f) in file_map.items()]
   args += ["--empty_file=%s" % f for f in empty_files or []]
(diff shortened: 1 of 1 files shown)


container/image.bzl
```diff
@@ -114,7 +114,7 @@ def _build_layer(ctx, files=None, file_map=None, empty_files=None,
   args += ["--file=%s=%s" % (f.path, magic_path(ctx, f)) for f in files]
   args += ["--file=%s=%s" % (f.path, path) for (path, f) in file_map.items()]
   args += ["--empty_file=%s" % f for f in empty_files or []]
-  args += ["--tar=" + f.path for f in ctx.files.tars]
+  args += ["--tar=" + f.path for f in tars if f.path.endswith(".tar")]
```
mattmoor: This will silently filter things, which worries me...

erain: OK, how about I change both `tars` and `debs` so it will accept the values in the parameters without the file extension check?

mattmoor:
Can you rebase this?

erain:
Thanks for the review! Rebased.

PR #233, opened 2017-11-15
Groovy sandwich

mattmoor:
Add fine-layering support to `rules_groovy`.

This is based on my PR that does that same for `rules_scala`.

### DIFF
--- a/groovy/image.bzl
+++ b/groovy/image.bzl
@@ -16,16 +16,19 @@
 The signature of groovy_image is compatible with groovy_binary.
 """
 
-load("@io_bazel_rules_groovy//groovy:groovy.bzl", "groovy_binary")
+load("@io_bazel_rules_groovy//groovy:groovy.bzl", "groovy_library")
 load("//container:container.bzl", "container_image")
 load(
     "//java:image.bzl",
     "DEFAULT_JAVA_BASE",
+    "jar_dep_layer",
+    "jar_app_layer",
     _repositories = "repositories",
 )
 
-# TODO(mattmoor): Take advantage of layering.
-def groovy_image(name, base=None, deps=[], layers=[], **kwargs):
+def groovy_image(name, base=None, main_class=None,
+                 srcs=[], deps=[], layers=[], jvm_flags=[],
+                 **kwargs):
   """Builds a container image overlaying the groovy_binary.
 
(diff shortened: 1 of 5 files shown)


damienmg:
Sandwich on java kind so irina :)

mattmoor:
Ok, the diffbase is merged and this rebased, so RFAL.

mattmoor:
@iirina Any chance you can TAL?  I'd like to land this before I go on holiday.

mattmoor:
@iirina @damienmg ping

mattmoor:
Right now all of the `lang_image` rules live here (for better or for worse), but tl;dr I'd love for the language rule owners to take over some measure of ownership (and of the distroless reference runtimes).

As some background, you can think of `container_image` as a universal replacement for our internal package format and the mono-archives that Blaze traditionally produces (e.g. PAR, deploy-JAR, SAR, ...).  However, in addition to bundling all of the `runfiles` together we also ~link it against a base Docker image containing a reference implementation of the target environment (e.g. JRE, Python interpreter, glibc, ...).

With `lang_image` the aim is to produce a (largely) drop-in replacement for `lang_binary` rules that instead of producing a mono-archive, produce a Docker image.  In cases like Python, JRE languages and (hopefully today) Node.js we can take better advantage of the Docker image format and separate out unchanging elements into their own layers to speed up developer iteration (if it doesn't change then you avoid repushing/repulling and get significantly faster redeployments, on the order of 99% less push/pull).

I hadn't fully take advantage of the layering aspects of this in Scala (until recently) and Groovy (until this PR).  It turned out that the same thing blocking my initial super-naive attempt at layering was the same problem as JRE-languages face trying to use each others libraries

The bulk of this change involves inlining `groovy_binary` for two reasons (IIRC):
1. The `//external:groovy` dependency isn't properly added to `runfiles` (again IIRC, this is fixable upstream)
1. We want `//external:groovy` in its own layer (not fixable upstream)

mattmoor:
@iirina Did I answer your question?  Anything else I can speak to?

mattmoor:
@iirina Ping :)

iirina:
@mattmoor I'm so sorry, all my emails from rules_docker were filtered away :( Thanks for the explanation.

Issue #229, opened 2017-11-10
Python 3 `py_image` support


hwright:
The base image used for `py_image` uses Python 2.7:
https://github.com/bazelbuild/rules_docker/blob/master/python/image.bzl#L43

It would be nice to have a python 3 `py_image` rule available as well, rather than having to construct the rule manually using lower-level constructs.  I realize the base images are grabbed from https://github.com/GoogleCloudPlatform/distroless, and it doesn't yet look like they have a python 3 image available, so there's probably a bit of work involved here.

The relevant Python 3 bug for distroless is here: GoogleCloudPlatform/distroless/issues/111

mattmoor:
/cc @duggelz @dlorenc @r2d4 FYI

Yeah, having a distroless Python 3 base is not strictly a pre-requisite, but would be nice for consistency with the other images.

Having not messed with Python versions too much in Bazel, is there an appropriate kwarg to `py_image` we could key base image selection off of?

hwright:
@mattmoor I'd almost be tempted to call it `py3_image` rather than `py_image` with some special attributes.  There's enough of a language difference that it might make sense to treat it as its own beast.

As for the Bazel support: python 3 support is all over the map, and I'm not sure what the overall plan is there.  I don't know that we want to wait for it to get sorted out before moving forward here.

PR #235, opened 2017-11-15
Add a py3_image rule

hwright:
This rule is an analog to `py_image`, only the resulting image runs
Python 3 instead of Python 2.  I'm certainly open to suggestions on
a more elegant way of doing this using attributes on the stock
`py_image` rule, but I'm not sure if there is one.  This was mostly
made by copying the implementation of `py_image`.

This change also updates a bunch of unrelated stuff as a consequence
of running `update_deps.sh`.

This fixes bazelbuild/rules_docker#229


bazel-io:

python3/image.bzl
```diff
@@ -0,0 +1,92 @@
...
+  binary_name = name + ".binary"
+
+  # TODO(mattmoor): Consider using par_binary instead, so that
+  # a single target can be used for all three.
+  native.py_binary(name=binary_name, deps=deps + layers, **kwargs)
+
+  # TODO(mattmoor): Consider what the right way to switch between
+  # Python 2/3 support might be.  Perhaps just overriding `base`,
+  # but perhaps we can be smarter about selecting a py2 vs. py3
+  # distroless base?
```
mattmoor: Seems like we're making a call here :)

Can you drop it here and the source?

testdata/BUILD
```diff
@@ -502,6 +502,22 @@ py_image(
     main = "py_image.py",
 )
 
+load("//python3:image.bzl", "py3_image")
+
+py_library(
+    name = "py3_image_library",
+    srcs = ["py3_image_library.py"],
+)
```
mattmoor: Not sure this is meaningfully different, can we just use the existing `py_libary`?  LMK if I'm missing a change.

mattmoor:
Test this please

testdata/BUILD
hwright: Nope, this can be removed and the existing one reused.

python3/image.bzl
hwright: Done.

Though I'm still on the fence about the approach here: we could logically extend it toward treating `py_binary` and `py_library` as python 2-only, and then adding `py3_binary` and `py3_library` rules to `rules_python` and so forth.  They are two different languages in some senses, though I don't know if I want that precedent here.

mattmoor: I will leave that to the Python gods to decide.  I just play one on GitHub :)

@duggelz FYI

mattmoor:
Test this please

mattmoor:
Thanks again @hwright :)

mattmoor:
It occurred to me immediately after pressing merge that you didn't update [this](https://github.com/bazelbuild/rules_docker#language-rules), which will improve discoverability.

hwright:
And I haven't even managed to start playing one on github: #238 :)

PR #239, opened 2017-11-16
Generate a per package metadata file in /var/lib/dpkg/status.d directory for drydock

tejal29:
Hello Reviewers, 
The changes are as per the [DD](https://docs.google.com/document/d/12X3R6gIfseKXvvZ3fQKs-YyOA_SN1xqThnU_xQERIMc/edit#heading=h.i0q0yuihg4f1)

@r2d4 PTAL

PR #249, opened 2017-11-29
Add back container_test

tejal29:
This is to add back container_tests.
I was going to combine it with https://github.com/bazelbuild/rules_docker/pull/239 however the diff was getting bigger.


container/container.bzl
```diff
@@ -20,6 +20,7 @@ load("//container:import.bzl", "container_import")
 load("//container:load.bzl", "container_load")
 load("//container:pull.bzl", "container_pull")
 load("//container:push.bzl", "container_push")
+load("//contrib:test.bzl", "container_test")
```
mattmoor: We don't include `contrib` things here.  If it's ready to graduate it should move to `//container`, but I don't think that's the intent of this change?

tests/docker/BUILD
```diff
@@ -0,0 +1,22 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+package(default_visibility = ["//visibility:public"])
+
+load("//container:container.bzl", "container_test")
+
+container_test(
```
mattmoor: Why do we want the `..../docker/...` path part?

container/container.bzl
tejal29: no.. let me just get it to work here. We have a task internally to convert all image_test.py to use container_test
Maybe then we will ready.. Just created one here https://github.com/bazelbuild/rules_docker/issues/256

tests/docker/BUILD
tejal29: You mean absolute target reference?
I am used to writing full target names. I think its cleaner and easier when you move code around.

mattmoor:
Twiddling for CI

container/container.bzl
dlorenc: Should this line be dropped?

tejal29:
@dlorenc in order to make commit history look pretty i messed up the files. You comment is gone, but the container/BUILD does not load container_test from contrib/test.bzl

PR #252, opened 2017-11-30
Support external binaries in lang_image rules.

mattmoor:
This applies to cc, rust, d, and go (none of which benefit appreciable from layering).

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,10 +39,6 @@ registry interactions.
 
 ## Language Rules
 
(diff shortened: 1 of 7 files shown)


mattmoor:
Part of the motivation for this is that there is no real way to tell Gazelle to manage `go_image` for you as-if if were `go_binary`.

mattmoor:
Twiddling for CI

PR #258, opened 2017-12-05
Fix new buildifier errors.

mattmoor:
These were introduced by a change to buildifier or Bazel.


mattmoor:
@dlorenc @Capstan @hwright FYI

PR #263, opened 2017-12-07
Update gcr.io/distroless digests.

greggdonovan:
Updating the `gcr.io/distroless` DIGESTS via [`./update-deps.sh`](https://github.com/bazelbuild/rules_docker/blob/master/update_deps.sh).

My motivation is to get a version of the python 2.7. with `libssl.so.1.1`. This is present in the latest image but not in the version current specified in DIGESTS. I can update the PR to just include the Python 2.7 DIGESTS changes, if that is preferable.

### DIFF
--- a/cc/cc.bzl
+++ b/cc/cc.bzl
@@ -18,8 +18,8 @@
 # git repository.
 
 DIGESTS = {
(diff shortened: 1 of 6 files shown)


greggdonovan:
Thanks, @mattmoor!

PR #279, opened 2018-01-04
implement container_layer rule.

erain:
* each container_layer rule will create its own tar files.
* container_image rule was refactored so that it can composed by
multiple container_layers via such new attr.

Tested:
- new unit tests in image_test.py


container/image.bzl
```diff
@@ -244,35 +194,41 @@ def _impl(ctx, files=None, file_map=None, empty_files=None, directory=None,
...
+  zipped_layer, blob_sum = zip_layer(ctx, unzipped_layer)
+
+  # Get the layers and shas from our base.
+  # These are ordered as they'd appear in the v2.2 config,
+  # so they grow at the end.
+  parent_parts = _get_layers(ctx, ctx.attr.base, ctx.files.base)
+  zipped_layers = parent_parts.get("zipped_layer", []) + _getLayerInfo(layers, "zipped_layer") + [zipped_layer]
+  shas = parent_parts.get("blobsum", [])  + _getLayerInfo(layers, "blob_sum") + [blob_sum]
+  unzipped_layers = parent_parts.get("unzipped_layer", []) + _getLayerInfo(layers, "unzipped_layer") + [unzipped_layer]
+  layer_diff_ids = _getLayerInfo(layers, "diff_id") + [diff_id]
```
mattmoor: nit: You are inconsistent about whether this portion is hoisted (like this one) or a nested expression.  IMO pick one and keep these lines more consistent.

container/image.bzl
```diff
@@ -244,35 +194,41 @@ def _impl(ctx, files=None, file_map=None, empty_files=None, directory=None,
...
+  diff_ids = parent_parts.get("diff_id", []) + layer_diff_ids
+
+  # Get and merge environment variables
+  env = {}
+  [env.update(layer[LayerInfo].env) for layer in layers]
```
mattmoor: This doesn't work, it doesn't properly evaluate read-modify-write environment variables like `PATH`.

I believe that for this to work properly, we actually want N calls to `_image_config(...)`.

container/image.bzl
erain: Sorry that I didn't quite get the the meaning of "hoisted". 

Do you mean I should use all nested expression e.g. list comprehension in this case?

container/image.bzl
erain: It looks like _image_config will deal with more things than envvars. 

I wrote a skylark function `_evalute_merge_env ` to handle envvar here properly. And I have unit test to cover the read-modify-write case like `PATH`.

container/image.bzl
```diff
@@ -342,6 +298,7 @@ _attrs = dict({
     ),
     "label_file_strings": attr.string_list(),
     "empty_files": attr.string_list(),
+    "container_layers": attr.label_list(providers = [LayerInfo]),
```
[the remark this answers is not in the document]
erain: Emm... this is tricky since `layers` is also used in `lang_image` rule, which extends the attrs from `container_image` rule. 

So if we want to keep `layers` in `container_image` rule, we need to change the `layers` in `lang_image` to something else. 

Do you have some suggestions for me?

container/image_test.py
```diff
@@ -66,6 +66,14 @@ def test_files_with_file_base(self):
       self.assertEqual(2, len(img.fs_layers()))
       self.assertTopLayerContains(img, ['.', './bar'])
 
+  def test_files_with_file_base(self):
```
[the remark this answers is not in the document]
erain: Thanks for the catch. Done.

container/image_test.py
```diff
@@ -83,6 +91,17 @@ def test_tar_with_tar_base(self):
         './asdf', './usr', './usr/bin',
         './usr/bin/miraclegrow'])
 
+  def test_tar_with_tar_base(self):
```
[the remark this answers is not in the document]
erain: Done.

container/layer.bzl
```diff
@@ -0,0 +1,169 @@
...
+)
+load(
+    "//skylib:path.bzl",
+    "dirname",
+    "strip_prefix",
+    _canonicalize_path = "canonicalize",
+    _join_path = "join",
+)
+
+def magic_path(ctx, f):
```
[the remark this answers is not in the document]
erain: Done. It is not used in `lang_image` as you said.

container/layer.bzl
```diff
@@ -0,0 +1,169 @@
...
+  debs = debs or ctx.files.debs
+  tars = tars or ctx.files.tars
+  # Generate the unzipped filesystem layer, and its sha256 (aka diff_id)
+  unzipped_layer, diff_id = build_layer(ctx, files=files, file_map=file_map,
+                                        empty_files=empty_files,
+                                        directory=directory, symlinks=symlinks,
+                                        debs=debs, tars=tars)
+  # Generate the zipped filesystem layer, and its sha256 (aka blob sum)
+  zipped_layer, blob_sum = zip_layer(ctx, unzipped_layer)
+  # Returns constituent parts of the Container layer as provider
```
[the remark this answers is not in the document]
erain: Done. Added doc explaining why we use provider.

container/layer.bzl
```diff
@@ -0,0 +1,169 @@
...
+  return [LayerInfo(zipped_layer=zipped_layer,
+                    blob_sum=blob_sum,
+                    unzipped_layer=unzipped_layer,
+                    diff_id=diff_id,
+                    env=env)]
+
+_layer_attrs = dict({
```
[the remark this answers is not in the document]
erain: I didn't handle this comments yet, since I am still not sure about how to change `container_image` to a macro. 

There are still some attrs which are `container_image` only, e.g. `legacy_run_behavior`, `labels` etc., which are not in `container_layer` rule.  Also how to deal with the backward compatibility of the existing usage of `container_image` rule? 

So how about 1) I re-use all the `container_layer` attrs and 2) re-use the `container_layer` impl for the related attrs, but still keep `container_image` as a rule?

testdata/BUILD
```diff
@@ -83,6 +84,20 @@ container_image(
     mode = "0644",
 )
 
+container_image(
+    name = "files_in_layer_with_files_base",
+    base = ":files_base",
+    container_layers = [":files_in_layer"],
```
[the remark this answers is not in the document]
erain: I don't quite understand this. `:files_base` is the base (which is a bunch of file), and `layers` is the files on top of that. Am I wrong?

erain:
Hi Matt, 

I have done more refactor of this PR, including: 
- reuse `container_layer` attrs in `container_image` rule
- reuse the implementation of `container_layer` in `container_image` rule, and treat the data from `container_image` as one more layer of `layers` attr. 
- rename: `container_layers` -> `layers`. In order to do this, I rename quite a few attrs in `lang_image` rules. 
- add documents for `container_layer` rule. 

Could you take another review of this PR? 

Many thanks, 
Yu

container/image.bzl
mattmoor: > It looks like _image_config will deal with more things than envvars.

That's fine, so just pass env vars :)

> I wrote a skylark function _evalute_merge_env to handle envvar here properly

I'd rather not fork this logic across two places.  I'd definitely prefer we use `_image_config` iteratively.

container/image.bzl
mattmoor: By hoisted I mean:
```
foo = _getLayerInfo(...)
bar = parent_parts.get(...) + foo
```
vs:
```
bar = parent_parts.get(...) + _getLayerInfo(...)
```

container/image.bzl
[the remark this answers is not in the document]
mattmoor: I don't think this is a problem.  `lang_image` wraps `dep_layer` and `app_layer` rules which extend the `container_image` rule.  I don't believe `layers` is passed through.

container/layer.bzl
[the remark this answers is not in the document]
mattmoor: I'm ok with punting on this for now, but I'd still like the rules to head this general direction to avoid repeating logic.  This is the motivation for some of my comments like `_image_config`.

The idea is that `container_image` would still be a rule, but likely `_container_image`, and `container_image` becomes a macro that wraps `container_layer` and `_container_image`.

container/flatten.bzl
```diff
@@ -14,7 +14,11 @@
 """A rule to flatten container images."""
 
 load(
-    "//container:layers.bzl",
+    "//skylib:path.bzl",
+    _get_runfile_path = "runfile",
+)
```
mattmoor: Is this used?  trying to understand the diff.

container/image.bzl
```diff
@@ -216,9 +161,25 @@ def _repository_name(ctx):
   # the v2 registry specification.
   return _join_path(ctx.attr.repository, ctx.label.package)
 
+def _evalute_merge_env(env, new_env):
```
mattmoor: typo: evaluAte

container/image.bzl
```diff
@@ -216,9 +161,25 @@ def _repository_name(ctx):
...
+  """Expand new environment variables properly and merge them.
+  Example:
+  - input: env={'foo':'bar', 'PATH':'$PATH:a:b'}
+           new_env={'X':'Y', 'PATH':'$PATH:c' }
+  - output: env={'foo':'bar', 'X':'Y', 'PATH':'$PATH:a:b:c'}
+  """
+  for k, v in new_env.items():
+    elems = []
+    for e in v.split(":"):
+      if e.startswith("$") and e[1:] in env.keys():
```
mattmoor: This doesn't handle `a:$PATH:b` or `${PATH}:a:b`.

I'd really rather see this go away in favor of `_image_config`.

mattmoor: `PATH` is also not the environment variable that supports read-modify-write, so splitting on `:` is also insufficient.

java/image.bzl
```diff
@@ -286,7 +286,7 @@ def _war_app_layer_impl(ctx):
   """Appends the app layer with all remaining runfiles."""
 
   available = depset()
-  for jar in ctx.attr.layers:
+  for jar in ctx.attr.jar_layers:
```
mattmoor: I see, we do have layers.  This rename is better because these are an implementation detail.

mattmoor:
@philwo No CI is running for this repo, what's up?

philwo:
@mattmoor I've setup post-/presubmit checks for this repo now. The presubmit for this PR will run when a new commit is pushed to it.

I also sent you an invite to Buildkite - please accept it. You'll need to verify PRs from external contributors to make sure that they're safe to run on CI. :)

Let me know if you have any questions or if it doesn't work.

container/flatten.bzl
erain: Sorry I was not familiar with the PR previously and I rebased my change instead of merging master. So this was unrelated change somehow got in.

Fixed right now.

container/image.bzl
erain: Fixed as this function was deleted.

container/image.bzl
erain: Done. 

Now use //container:create_image_config to generate config layer by layer so the envvars can be handled correctly.

java/image.bzl
erain: Ack.

erain:
All comments addressed, please help review again.  

As for the re-use `container_layer` rule in `container_image`:
> I'm ok with punting on this for now, but I'd still like the rules to head this general direction to avoid repeating logic.

Yes, currently I just use `container_layer` in `container_image` to generate the last layer. I think maybe later I can send other refactor PR to make `container_layer` more reusable. 

Many Thanks!

container/image.bzl
```diff
@@ -221,15 +206,14 @@ def _impl(ctx, files=None, file_map=None, empty_files=None, directory=None,
...
-  # Generate the new config using the attributes specified and the diff_id
-  config_file, config_digest = _image_config(
-      ctx, layer_diff_ids,
-      entrypoint=entrypoint or ctx.attr.entrypoint,
-      cmd=cmd or ctx.attr.cmd,
-      env=env)
+  # Get the config for the base layer
+  config_file = _get_base_config(ctx)
+  # Generate the new config layer by layer, using the attributes specified and the diff_id
+  for i in range(len(layers)):
```
mattmoor: for i, layer in enumerate(layers):

erain: Done. 

Many thanks for the review and approval!!

philwo:
> @philwo No CI is running for this repo, what's up?

Fixed. Let me know if you encounter any issues.

Issue #218, opened 2017-11-05
testdata/BUILD warning: target 'passwd' is both a rule and a file


Capstan:
`WARNING: /…/rules_docker/testdata/BUILD:408:1: target 'passwd' is both a rule and a file; please choose another name for the rule.`

Capstan:
Repros with `bazel build //testdata:cc_image`

mattmoor:
@tejal29 I think we can fix this by simply referencing the output differently, can you TAL at this?

@dlorenc FYI

ash2k:
I think this has been fixed in https://github.com/bazelbuild/rules_docker/pull/306.

nlopezgi:
is this still an issue? please reopen if so.

PR #295, opened 2018-01-26
using different output name to avoid errors

schoren:
This commit fixes bazelbuild/rules_docker#218 by appending ".passwd" to the output file, making it different from the name. This was the suggested solution by @mattmoor.

### DIFF
--- a/README.md
+++ b/README.md
@@ -463,12 +463,27 @@ modified sample from the `distroless` repository:
 ```python
 load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
 
-# Create a passwd file with a nonroot user and uid.
-passwd_file(
-    name = "passwd",
+# Create a passwd file with a root and nonroot user and uid.
+passwd_entry(
+    username = "root",
+    uid = 0,
+    gid = 0,
+    name = "root_user",
+)
+
+passwd_entry(
+    username = "nonroot",
     info = "nonroot",
     uid = 1002,
-    username = "nonroot",
+    name = "nonroot_user",
+)
+
+passwd_file(
+    name = "passwd",
+    entries = [
+        ":root_user",
+        ":nonroot_user",
+    ],
 )
 
 # Create a tar file containing the created passwd file
--- a/container/BUILD
+++ b/container/BUILD
@@ -101,6 +101,7 @@ TEST_TARGETS = [
     "link_with_files_base",
     "build_with_tag",
     "with_passwd",
+    "with_group",
     "gen_image",
     "data_path_image",
     "no_data_path_image",
--- a/container/image_test.py
+++ b/container/image_test.py
@@ -281,7 +281,7 @@ class ImageTest(unittest.TestCase):
 
   def test_with_passwd(self):
     with TestImage('with_passwd') as img:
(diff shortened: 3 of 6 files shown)


bazel-io:

mattmoor:
FWIW, what I meant in that issue is that I think Bazel has a specific way of referring to its main output file, which has the same name as the rule itself.

schoren:
@mattmoor sorry, this is my first contribution to a bazel related project. Is the fix ok, or is the approach wrong? (besides the failing tests, which I'm wokring on)

mattmoor:
I think the fix I'd like is different.  I think the proper fix deletes `outputs = {}` entirely.

@tejal29 @dlorenc Can one of you help @schoren ?

schoren:
I've been reading the Bazel docs (https://docs.bazel.build/versions/master/skylark/rules.html#output-files) which point to this example: https://github.com/bazelbuild/examples/blob/master/rules/implicit_output/hash.bzl

In the example, the outputs are defined pretty much the way I did in this PR. I didn't find any docs or example that creates a default output. Maybe you can point me in the right direction here?

Thanks

tejal29:
@schoren Thanks for picking this up!

Here is an example of what you can do, 
```
passwd_file  = ctx.actions.declare_file(ctx.label.name + ".passwd")
...
...
return DefaultInfo(files=depset([passwd_file]))
```

And then you totally emit the outputs.

mattmoor:
@tejal29 Doesn't this bring back the problem that you originally solved?

tejal29:
@bazel-io please test this

tejal29:
@mattmoor The way @schoren has it implemented now works.
```
load("contrib//:passwd_file.bzl", "passwd_file")
load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
passwd_file(
    name = "passwd",
    gid = 2345,
    home = "/myhomedir",
    info = "myusernameinfo",
    shell = "/myshell",
    uid = 1234,
    username = "foobar",
)

pkg_tar(
  name="tar",
  srcs = [":passwd"],
  mode = "0644",
  package_dir = "etc",
)
```
You do not see any warning.
```
$ bazel clean
INFO: Starting clean (this may take a while). Consider using --async if the clean takes more than several minutes.
 $ bazel build :tar
INFO: Analysed target //:tar (19 packages loaded).
INFO: Found 1 target...
Target //:tar up-to-date:
  bazel-bin/tar.tar
INFO: Elapsed time: 2.176s, Critical Path: 0.21s
INFO: Build completed successfully, 7 total actions
$ tar -tf bazel-bin/tar.tar 
./
./etc/
./etc/passwd
 ```

tejal29:
@schoren Can you please rebase your branch off latest master and then i can approve this request!

Thanks
Tejal

schoren:
@tejal29 Done!

dlorenc:
@bazel-io test this please

tejal29:
Looks like your diff now contains previous PR contrib.
@dlorenc i am not sure if this would corrupt the git history. 
Rebase would have been much better.

schoren:
Sorry, I guess I messed up the commands :S. If you want I can create a new fork and redo the work, with a new pull request. Would that be better?

tejal29:
closed in favor of #306

PR #324, opened 2018-02-16
Improve py_image layering

scele:
Consider the following more complex python image that uses the pip requirement feature of rules_python:

    load("@pip_deps//:requirements.bzl", "requirement")

    py_library(
        name = "py_image_library_with_external_deps",
        srcs = ["py_image_library.py"],
        deps = [requirement("tensorflow")],
    )

    py_library(
        name = "py_image_library",
        srcs = ["py_image_library.py"],
        deps = [":py_image_library_with_external_deps"],
    )

    py_image(
        name = "py_image_complex",
        srcs = ["py_image.py"],
        layers = [
            ":py_image_library_with_external_deps",
            ":py_image_library",
        ],
        main = "py_image.py",
    )

Currently, it will produce the following layers:

    301M bazel-bin/testdata/py_image_complex.0-layer.tar
    301M bazel-bin/testdata/py_image_complex.1-layer.tar
     15M bazel-bin/testdata/py_image_complex-layer.tar

This change improves the layering by fixing the following issues:
    
1. The middle layer does not detect that all tensorflow files were already added by the first layer.
    
2. The last layer is 15MB in size due to the huge symlink forest. Create a separate symlink layer for each dependency layer to avoid bloating the top layer, while still keeping the dependency layers image agnostic.
    
3. Getting "tensorflow" to its own layer requires the author of the "py_image" rule to know that libraries transitively depend on "tensorflow".  This is not great, since library authors may addand remove dependencies without remembering to update the "layers" list.  To counter this, add a new py_layer rule that can be used to select from transitive dependencies and group them to layers, without explicitly naming the transitive dependencies like "tensorflow".

With these improvements, the above example can be rewritten as:

    load("@pip_deps//:requirements.bzl", "requirement")

    py_library(
        name = "py_image_library_with_external_deps",
        srcs = ["py_image_library.py"],
        deps = [requirement("tensorflow")],
    )

    py_library(
        name = "py_image_library",
        srcs = ["py_image_library.py"],
        deps = [":py_image_library_with_external_deps"],
    )

    py_layer(
        name = "external_deps",
        deps = [":py_image_library"],
        filter = "@",
    )

    py_image(
        name = "py_image_complex",
        srcs = ["py_image.py"],
        layers = [
            ":external_deps",
            ":py_image_library",
        ],
        main = "py_image.py",
    )

This will yield the following layer structure:

    310M bazel-bin/testdata/py_image_complex.0-layer.tar           (agnostic)
     15M bazel-bin/testdata/py_image_complex.0-symlinks-layer.tar  (non-agnostic)
     10K bazel-bin/testdata/py_image_complex.1-layer.tar           (agnostic)
     10K bazel-bin/testdata/py_image_complex.1-symlinks-layer.tar  (non-agnostic)
     20K bazel-bin/testdata/py_image_complex-layer.tar             (non-agnostic)


bazel-io:

container/image.bzl
```diff
@@ -238,7 +238,7 @@ def _impl(ctx, files=None, file_map=None, empty_files=None, directory=None,
 
   file_map = file_map or {}
   files = files or ctx.files.files
-  empty_files = empty_files or ctx.attr.empty_files
+  empty_files = empty_files or {f: f for f in ctx.attr.empty_files}
```
mirandaconrado: Why do we need to change this from a list to dictionary? The only uses seem to be in calling `_build_layer`, which isn't changed in this PR so I imagine it doesn't care, and in creating `transient_emptyfiles`, which could be done with the list directly. Am I missing something?
Also, if this change is indeed necessary, we should probably update the documentation of this function to reflect the new type.

scele: I used to pass relative paths in the dict value. But now I reshuffled things and found out it's not needed, so I removed this confusing thing.

scele:
@mattmoor would you have time to take a look?

container/image_test.py
```diff
@@ -336,6 +333,96 @@ def test_py_image(self):
...
         './app/io_bazel_rules_docker/testdata/py_image_library.py',
         './app/io_bazel_rules_docker/testdata/__init__.py',
+        '/app',
+        '/app/testdata',
+        '/app/testdata/py_image.binary.runfiles',
+        '/app/testdata/py_image.binary.runfiles/io_bazel_rules_docker',
+        '/app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/testdata',
+        '/app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/testdata/py_image_library.py',
+        '/app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/testdata/__init__.py',
+      ])
```
mattmoor: Putting the symlinks for all layers into the binary's layer was (subtle but) intentional because it enables the same dependency's layer to be shared by multiple binary targets.

Looking at the "before" diff you see no mention of `py_image.binary` in the files, this means that I could have two `py_image` targets `foo` and `bar` consume the same dependency and both would utilize the same layer for that dependency in the registry (eliding uploads and downloads).

The concern isn't space in the registry (e.g. avoiding N copies of a `grpcio` dependency) it's more about cache density.  The Registry is effectively a distributed cache and if we make the layer for the same dependency always produce the same layer, then all consumers of that layer hit once one of them publishes it.  The same is true of pull latency, since the daemon shouldn't repull a layer it already has.

mattmoor: It is notable that if we're going to head this route, we could also simply forgo the symlinks altogether, right?

container/image.bzl
```diff
@@ -272,6 +272,8 @@ def _impl(ctx, files=None, file_map=None, empty_files=None, directory=None,
   shas = parent_parts.get("blobsum", []) + [blob_sum]
   unzipped_layers = parent_parts.get("unzipped_layer", []) + [unzipped_layer]
   diff_ids = parent_parts.get("diff_id", []) + [diff_id]
+  transient_files = parent_parts.get("transient_files", depset()) + [f for f in file_map]
+  transient_emptyfiles = parent_parts.get("transient_emptyfiles", depset()) + empty_files
```
mattmoor: Do you mean transitive?

python/image.bzl
```diff
@@ -58,7 +58,7 @@ DEFAULT_BASE = select({
     "//conditions:default": "@py_image_base//image",
 })
 
-def py_image(name, base=None, deps=[], layers=[], **kwargs):
+def py_image(name, base=None, deps=[], layers=[], external_layer=False, **kwargs):
```
mattmoor: This new kwarg isn't documented below.

testdata/py_image_complex.py
```diff
@@ -0,0 +1,22 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
mattmoor: nit: 2018 :)

python/image.bzl
```diff
@@ -72,14 +72,21 @@ def py_image(name, base=None, deps=[], layers=[], **kwargs):
   # a single target can be used for all three.
   native.py_binary(name=binary_name, deps=deps + layers, **kwargs)
 
+  base = base or DEFAULT_BASE
+
+  # Add a separate layer for external deps, usually they don't change very often.
+  if external_layer:
```
mattmoor: Making this a single layer will significantly hurt performance when a small dependency is added, changed, or removed.  It also creates an interesting combinatorial problem for cross-image dependency sharing.

I agree that this is a 10-20% case, not an 80-90% case.

For rules_python, I wonder if simply using:
```
py_image(
    name = "foo",
    ...
    layers = all_requirements,
)
```

Would get you what you want?  Maybe, maybe not because IIRC it's intentionally dumb about deduplicating across deps (to keep them agnostic of the binary context).

README.md
```diff
@@ -362,6 +362,9 @@ py_image(
 )
 ```
 
+You can also put all external dependencies to a separate layer at the bottom,
+by setting `external_layer=True`.
```
mattmoor: This isn't specific to `py_image`, and isn't really "fine layering".  Perhaps add a new example illustrating how to do this grouping all external deps?

fahhem:
@scele, this PR seems very useful! Do you want help getting it updated for the review and merge conflicts? I'd love to see this get in

container/image_test.py
scele: > It is notable that if we're going to head this route, we could also simply forgo the symlinks altogether, right?

Yes. I wonder if we could just expose `agnostic_dep_layout` attribute in `py_image`...? Then the user could pick from the following combinations:

`layering_strategy='group_external' agnostic_dep_layout=False`
- Create external layer, no symlinks, no cache sharing between images

`layering_strategy='group_external' agnostic_dep_layout=True`
- Create "external layer", and create "external symlinks layer". The latter would create symlinks to the former for the runfiles dir of the current image. The external layer could be shared across images, the symlinks layer not.

`layering_strategy='default' agnostic_dep_layout=False`
- Current behavior of `py_image`, except don't create symlinks, no cache sharing between images. This can still reduce symlink overhead in the app layer, even if you don't want to lump all external deps to a single layer.

`layering_strategy='default' agnostic_dep_layout=True`
- Current behavior of `py_image`.

container/image.bzl
scele: Yes!

python/image.bzl
scele: The issue with `layers = all_requirements` is that the `requirements.txt` file in the repo might contain a LOT of pip packages, while only some of them are used in the `py_image` that is being built. I wouldn't want to include the useless packages to the image, and I wouldn't want the built image to change just because somebody adds an unrelated pip package to the `requirements.txt`.

scele:
Sorry for letting this sit for so long.. Thanks for your thoughtful review @mattmoor ! 

re: my ideal world

I like the idea of putting every pip dep to its own layer. I worry about the 127 layer limit that docker apparently has (at least for some FS drivers), and it's also documented as a limitation of ECR. The code I'm building the image for has > 60 pip packages already today, and it won't be going down. Also, docker push and pull seem to be slower for high number of layers, even if none of the layers changed (I haven't measured this though).

> I think my preference in this case would be to replace external_layer = True with a layering_strategy = "a-descriptive-strategy-name". WDYT?

Sounds good! Perhaps `layering_strategy='group_external|default'`?

I also wonder if we could expose equivalent functionality in a more flexible manner by building on top of the new `container_layer`. Something like:

```bzl
container_layer(
    name = "layer_1_external",
    files = ":lib_A_that_depends_on_pip_set_A",
    external_only = True,
)
container_layer(
    name = "layer_1",
    files = ":lib_A_that_depends_on_pip_set_A",
)
container_layer(
    name = "layer_2_external",
    files = ":lib_B_that_depends_on_pip_set_B",
    external_only = True,
)
container_layer(
    name = "layer_2",
    files = ":lib_B_that_depends_on_pip_set_B",
)
py_image(
    name = "app",
    src = "app.py",
    layers = [
        ":layer_1_external", # pip set A
        ":layer_1",          # source of lib A
        ":layer_2_external", # pip set B-A
        ":layer_2",          # source of lib B
    ],
)
```

scele:
Thanks also @fahhem for rebasing! :)

scele:
@mattmoor I uploaded a new version that tries to make layering even more flexible (see the updated cover letter of this PR and the updated README.md). I also took the liberty to reduce code duplication by merging "dep_layer" and "app_layer" into a single rule since they already were very similar.

What do you think?

scele:
@mattmoor would you have time to comment on the revised proposal?

nlopezgi:
Hi @scele sorry for the long time to respond to this PR, we have been transitioning the team owning these rules and some PRs fell through the cracks in the meantime. Is this work still needed? If so could you try to resolve conflicts and let me know when this is ready again for me to take a look? Again, apologies for the delay.

scele:
@nlopezgi rebased, ready for your review!

nlopezgi:
Thanks again for sending this PR. Kicked off tests and started reviewing this today. Will take some time to do a full review but will try to get back with, at least, some comments by EoW.
To followup on some comments from above,

> Sounds good! Perhaps layering_strategy='group_external|default'?

I do think it would be useful to have the alternative of still using the default layering strategy, but to be 100% honest I'm not sure if the effort to support both is worth it (do you have opinions on if this change in layering strategy could negatively impact other use cases?)

> I also wonder if we could expose equivalent functionality in a more flexible manner by building on top of the new container_layer.

I think that is worth exploring more. The container_layer was added after all these xx_images rules and leveraging it would be interesting. Do you think this is something that replaces/improves what you currently have in this PR or is this follow up improvements that you think could be made after this PR?

scele:
Yeah, I realize this MR can be quite a bit to digest.. :)

> I do think it would be useful to have the alternative of still using the default layering strategy, but to be 100% honest I'm not sure if the effort to support both is worth it (do you have opinions on if this change in layering strategy could negatively impact other use cases?)

The only change in behavior (for users who don't explicitly start using the new `py_layer` primitive) is as follows: If user has `py_image` or `py3_image` rules that use the `layers` attribute, each such layer will generate two container layers: one agnostic with the real files, and one symlink layer that links the files in the right places under runfiles.

It's done by these lines:

```py
    for index, dep in enumerate(layers):
        base = app_layer(name = "%s_%d" % (name, index), base = base, dep = dep)
        base = app_layer(name = "%s_%d-symlinks" % (name, index), base = base, dep = dep, binary = binary)
```

(I added these lines also to rust, go, d and cc images for consistency, but all of those also emit a warning print that says "xxx_image does not benefit from layers=[]" so we could also drop the symlink layer from those if you'd like. The java rules appear to benefit from explicit layering, but I'm not changing any behavior there.)

So, I think your question about supporting the old "default layering strategy" could be rephrased as "should we still support generating python images where symlinks are not separated to their own container layers?" The easiest way to support that would be to expose some new attribute like `py_image(..., squash_symlinks=True)` but exposing a hacky thing like that might be premature.. Perhaps we could see if people report any drawbacks of the new scheme, and then see how to best address those specific issues?

> I think that is worth exploring more. The container_layer was added after all these xx_images rules and leveraging it would be interesting. Do you think this is something that replaces/improves what you currently have in this PR or is this follow up improvements that you think could be made after this PR?

The last version of this PR already exposes `py_layer` which allows quite flexible layer partitioning as shown in README (the original version of this PR had `external_layer` boolean attribute to `py_image`, and Matt suggested renaming it to `layering_strategy`, but I abandoned both of these ideas in favor of the more flexible `py_layer`). However `py_layer` does not build on `container_layer`, and I'm not sure how that could even be leveraged.. For example, `container_layer` is dumb in the sense that it always adds all input files to the layer tarball, while the `py_image(layers=[...])` (which `py_layer` integrates to) only adds files if they are not already present on the previous layers.

So, I think there is some discrepancy between `container_layer` and the layers passed to `lang_image` rules. This MR does not solve the discrepancy, but I hope that it doesn't make it worse either.

container/image_test.py
scele: I can't speak for Matt, but I believe that I have addressed his concern in the latest version of the PR. In my initial version (it's a shame that force-push has lost that..), I had combined external files and external symlinks in a single "external layer", and Matt correctly pointed out that such layers could not be reused between different binary targets. However, in the latest version, we no longer have such "external layer" that would mix both files and symlinks. Instead, we have separate layers for *files* that can be reused across binary targets (layers with `./app/*`), and separate layers for *syminks* that cannot be shared (layers with `/app/testdata/py_image.binary.runfiles/*`).

The difference from the current top-of-tree is that my PR does not lump *all* symlinks to the top layer. This is to split the top layer (15MB in my tensorflow example) to stable symlinks (e.g. symlinks to tensorflow-gpu files, change rarely) and app files (change frequently) and hence reduce build+push delta.

WORKSPACE
```diff
@@ -66,6 +66,25 @@ load(
 
 _py_image_repos()
 
+git_repository(
```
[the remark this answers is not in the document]
scele: Done

lang/image.bzl
```diff
@@ -265,5 +239,69 @@ app_layer = rule(
     }.items()),
     executable = True,
     outputs = _container.image.outputs,
-    implementation = _app_layer_impl,
+    implementation = app_layer_impl,
+)
+
+# Convenience function that instantiates tha _app_layer rule and returns
```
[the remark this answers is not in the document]
scele: Done.

scele:
Fixed WORKSPACE, typo, and [replied](https://github.com/bazelbuild/rules_docker/pull/324#discussion_r221439623) about layer sharing.

lang/image.bzl
```diff
@@ -18,13 +18,30 @@ load(
     "//container:container.bzl",
     _container = "container",
 )
+load(
+    "//container:layer_tools.bzl",
+    _get_layers = "get_from_target",
+)
+
+FilterAspectInfo = provider(
```
nlopezgi: Do you think these should be in container/providers.bzl?

scele: Yes! Done.

scele:
Moved providers, and rebased on top of master. I also squashed the review fixes to the main commit (if that was a mistake, the unsquashed and unrebased changes are still in my `improve_layers_before_rebase` branch).

container/image_test.py
nlopezgi: sgtm

PR #325, opened 2018-02-17
Fix py3 debug image install error

samuraisam:
When using the py3 image in debug mode (via `-c dbg`) the image fails to load due to what looks like a typo. This pull request fixes that bug.

I've also added a note to the readme for py3 users, to call out that they will need to use the images from the "python3" package instead of the "python" package - what is probably a common gotcha.

### DIFF
--- a/README.md
+++ b/README.md
@@ -319,6 +319,8 @@ cc_image(
 To use `py_image`, add the following to `WORKSPACE`:
 
 ```python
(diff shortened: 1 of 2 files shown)


bazel-io:

mattmoor:
@bazel-io Test this please

mattmoor:
Can you rebase this?

mattmoor:
@samuraisam Can you rebase this so that I can make Bazel's new CI setup run over the PR?

Issue #269, opened 2017-12-20
build_tag_filters not working as expected


tejal29:
Hi, 
I want to disable building and testing a target on OSX. 
I have added the following in [rules_docker.json](https://github.com/tejal29/rules_docker/blob/build_tag_filter_repro/.ci/rules_docker.json) 
```
    {
        "node": "darwin-x86_64",
        "parameters": {
            "tests": ["//tests/docker/... except //tests/docker/..."],
            "build_tag_filters": ["-needs_xz"]
        }
```
and tests/docker:deb_image_with_dpkgs has the tag.
```
container_image(
    name = "deb_image_with_dpkgs",
    debs = [
        "//testdata:pkg1.deb",
        "//testdata:pkg2.deb",
        ":busybox-static_ubuntu1_amd64.deb",
    ],
    tags = ["needs_xz"],
)
```
However, i still see the target being built on OSX node.
Any advise would be appreciated.

Thanks
Tejal

tejal29:
Build error https://ci.bazel.io/blue/organizations/jenkins/PR%2Frules_docker/detail/rules_docker/350/pipeline/25

philwo:
Hi @tejal29,

the target will always be built if it is a dependency of anything else in your build, so you have to make sure that nothing that you build and test depends on it. I think this might be what's happening here.

The build_tag_filters and test_tag_filters can be helpful to exclude the targets and tests that rely on `deb_image_with_dpkgs`.

Hope that helps, let me know if you need more information.

Cheers,
Philipp

ash2k:
I think there is a bug in `rules_docker`. Seems like `tags` from the `go_image` rule are not "propagated" to the underlying rules it generates and hence there is still a dependency on `go_binary` and hence everything is built anyway even with `--build_tag_filters=-exclude_from_tests` option. Here is the snippet that is reproducing the behaviour:
```bzl
go_binary(
    name = "ops_race",
    embed = [":go_default_library"],
    race = "on",
    tags = [
        "exclude_from_tests",
    ],
    visibility = ["//visibility:public"],
)

go_image(
    name = "container_race",
    binary = ":ops_race",
    tags = [
        "exclude_from_tests",
    ],
    visibility = ["//visibility:public"],
)
```
If I comment out `go_image` and specify `--build_tag_filters=-exclude_from_tests` then that `go_binary` is not built (as expected). But if `go_image` is there then both are built even with `build_tag_filters`.
/cc @mattmoor

[...]

PR #333, opened 2018-02-28
The lang image rules like go_image did not propagate tags

davidmankin:
https://github.com/bazelbuild/rules_docker/issues/269

The lang image rules like `go_image` did not propagate tags
through to the `lang_image` rule so one could not use
build_tag_filters correctly (since there was always an
untagged image rule that depended on the binary rule
to be excluded).


davidmankin:
I fixed the CLA signature problem.

mattmoor:
Can you rebase this?

davidmankin:
Rebased.

PR #335, opened 2018-02-28
container_image produce container_layer provider

erain:
* The LayerInfo provider produced by the container_image attrs will be
returned as provider as well

* This enables composible container_images:

```python
container_image(
    name = "B",
    ...
    layers = [":A"],
)

container_image(
    name = "A",
    ...
)
```

Tested:
- unit test: one container_image depends on another

### DIFF
--- a/README.md
+++ b/README.md
@@ -1418,7 +1418,7 @@ container_image(name, base, data_path, directory, files, legacy_repository_namin
       <td><code>layers</code></td>
       <td>
         <code>Label list, optional</code>
-        <p>List of `container_layer` or `container_image` targets. </p>
+        <p>List of `container_layer` targets. </p>
         <p>The data from each `container_layer` or `container_image` without base will be part of produced container image, and the environment variable will be available in the image as well.</p>
       </td>
     </tr>
--- a/README.md
+++ b/README.md
@@ -1418,8 +1418,8 @@ container_image(name, base, data_path, directory, files, legacy_repository_namin
       <td><code>layers</code></td>
       <td>
         <code>Label list, optional</code>
-        <p>List of `container_layer` targets. </p>
-        <p>The data from each `container_layer` will be part of container image, and the environment variable will be available in the image as well.</p>
+        <p>List of `container_layer` or `container_image` targets. </p>
+        <p>The data from each `container_layer` or `container_image` without base will be part of produced container image, and the environment variable will be available in the image as well.</p>
       </td>
     </tr>
     <tr>
--- a/container/image.bzl
+++ b/container/image.bzl
@@ -68,6 +68,7 @@ load(
 load(
     "//container:layer.bzl",
     "LayerInfo",
+    "container_layer_",
     _layer = "layer",
 )
 load(
@@ -259,8 +260,7 @@ def _impl(ctx, files=None, file_map=None, empty_files=None, empty_dirs=None,
 
   return struct(runfiles = runfiles,
                 files = depset([ctx.outputs.layer]),
-                container_parts = container_parts,
-                providers = [image_layer[0]])
+                container_parts = container_parts)
 
 _attrs = dict(_layer.attrs.items() + {
     "base": attr.label(allow_files = container_filetype),
@@ -443,6 +443,10 @@ def container_image(**kwargs):
   loaded (an internal detail).  The expectation is that the images produced
   by these rules will be uploaded using the 'docker_push' rule below.
 
+  Also, this rule generates a `container_layer` target which can be reference
+  in the `layers` attr as `name.top_layer`. It is the layer generated from
+  `container_image` attrs only.
+
   Args:
     **kwargs: See above.
   """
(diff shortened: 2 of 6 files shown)


mattmoor:
This feels unnatural to me :)

I feel like if this is what you want, then you should:
```python
container_layer(
    name = "blah",
    ...
)

container_image(
    name = "B",
    ...
    layers = [":blah"],
)

container_image(
    name = "A",
    layers = [":blah"],
)
```

Perhaps I'm misunderstanding the purpose of this?

erain:
Hi Matt, 

The one in your example is one natural way to use `container_layer`. 

I am thinking about two things: 

1. If a team has a `container_image(name=java_image)`, and another team has a `container_image(name=python_image)`, from this PR I can directly depend on the layers in their images without them changing their implementation. 

2. [base-images-docker](https://github.com/GoogleCloudPlatform/base-images-docker/) already implemented a new [package_manager](https://github.com/GoogleCloudPlatform/base-images-docker/tree/master/package_managers), which will install deb packages via apt-get properly. However, since `container_layer` is without base, so we can only use the new package_manager in `container_image` rule.  This PR enables us to install deb packages properly in a `container_image` rule, and get everything except base as layer from another `container_image` rule. 

@nlopezgi WDYT?

Thanks, 
Yu

mattmoor:
I think that if we were going to expose this, it should be after `container_image` is just a macro that:
1. Delegates to: `container_layer(name = name + ".top_layer", ...)`, and
1. Links the image: `_container_image(name = name, base = base, layers = [name + ".top_layer"])`

Then the way folks would write your original example would be:
```python
container_image(
    name = "B",
    ...
    layers = [":A.top_layer"],   # This seems more natural.
)

container_image(
    name = "A",
    ...
)
```

WDYT?  Does that make sense?

nlopezgi:
as Yu mentions, what we want is to be able to have container_image targets that can be used to directly produce an image as well as to compose their layer as part of a different image. Does that sound like a valid use case Matt or is there some other reason why you feel this is wrong?

mattmoor:
@nlopezgi I think the part that feels unnatural (and non-obvious) is referencing an *image* as a *layer*.  Are all the layers added?  Is the top layer added?

I don't mind the variation I suggested above, where you reference the layer produced by the `container_image(name = "A")` via `:A.top_layer` (I'm not married to the name).

erain:
Thanks Matt for the clarification! 

I think to expose the LayerInfo provider as `my_image.top_layer` sounds good, and I will go with this path in this PR and will send update commit soon.

erain:
Hi Matt, 

I already adopted this PR with your `:my_impage.top_layer` method, by having a `container_layer_` target as part the `container_image` macro. 

PTAL when you have time and many thanks!

README.md
```diff
@@ -1419,7 +1419,7 @@ container_image(name, base, data_path, directory, files, legacy_repository_namin
       <td>
         <code>Label list, optional</code>
         <p>List of `container_layer` targets. </p>
-        <p>The data from each `container_layer` will be part of container image, and the environment variable will be available in the image as well.</p>
+        <p>The data from each `container_layer` or `container_image` without base will be part of produced container image, and the environment variable will be available in the image as well.</p>
```
mattmoor: I think you can revert this.  You should document the `.top_layer` side-output instead.

container/image.bzl
```diff
@@ -456,3 +462,10 @@ def container_image(**kwargs):
   if "entrypoint" in kwargs:
     kwargs["entrypoint"] = _validate_command("entrypoint", kwargs["entrypoint"])
   container_image_(**kwargs)
+  # Also create a "name.top_layer" container_layer target
+  top_layer_attrs = {}
+  for key in _layer.attrs.keys():
+    if key in kwargs:
+      top_layer_attrs[key] = kwargs[key]
+  top_layer_attrs["name"] = kwargs["name"] + ".top_layer"
+  container_layer_(**top_layer_attrs)
```
mattmoor: I was thinking more that `container_image` would *consume* the `.top_layer` itself here, otherwise we're effectively creating the top layer two different ways, which isn't what I'd expect as a user.

erain: Actually `container_image` is already doing so in the implementation (use container_layer's implementation for generating the top_layer). 

The only thing I am worry about is about the other user of `container_image`'s **implementation** (not the rule itself). Since we move the logic for generating the `top_layer` into the macro:
- people use the implementation will NOT not get the `top_layer` built
- at the same time people cannot reuse `container_image` macro in their implementations. 

WDYT?

mattmoor: Sorry for the delay.  I think it is fine (and actually should be by-design) that `lang_image` doesn't expose `.top_layer`.  For `lang_image`, it is my expectation that the layering choice is an implementation detail that could change at any time.

erain: Hi Matt, 

So sorry for not working lately. My main concern about splitting the logic `container_image` into pieces in macro is, internally we extend the implementation of `container_image`, and if we split part of its logic in macro, we can no longer re-use it easily by extending the implementation. 

Given that we probably need to migrate to the [new package manager](https://github.com/GoogleCloudPlatform/base-images-docker/tree/master/package_managers), I don't think this PR is urgent for us. So I am also OK to close this PR and park the changes for now.

erain:
Not using this feature, close for now. Will re-open in the future when needed.

PR #336, opened 2018-02-28
Fix groovy_image crash on recent bazel versions

bol:
The layers object is frozen so attempts to modify it will result in a crash.

This will happen if it is called without settint the layer parameter.
```
groovy_image(
    name = "hello",
    srcs = glob(["Application.groovy"]),
    main_class = "example.Application",
)
```

All test cases are currently called with the layer parameter set and thus avoid triggering this behaviour.
Unfortunately, the example in the readme does not have it and is currently failing.


bol:
I signed it!

PR #340, opened 2018-03-05
Expose base and outputs to be overriden.

xingao267:
This adds capability to override base and outputs of
_container.image.implementation(ctx, ...) from another rule's
implementation.

### DIFF
--- a/container/image.bzl
+++ b/container/image.bzl
@@ -82,10 +82,10 @@ load(
     _serialize_dict = "dict_to_associative_list",
 )
 
-def _get_base_config(ctx):
+def _get_base_config(ctx, base):
   if ctx.files.base:
     # The base is the first layer in container_parts if provided.
-    l = _get_layers(ctx, ctx.attr.base, ctx.files.base)
+    l = _get_layers(ctx, ctx.attr.base, ctx.files.base, base)
     return l.get("config")
 
 def _image_config(ctx, layer_names, entrypoint=None, cmd=None, env=None, base_config=None, layer_name=None):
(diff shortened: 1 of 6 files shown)


container/image.bzl
```diff
@@ -175,15 +177,19 @@ def _impl(ctx, files=None, file_map=None, empty_files=None, empty_dirs=None,
     entrypoint: str List, overrides ctx.attr.entrypoint
     cmd: str List, overrides ctx.attr.cmd
     symlinks: str Dict, overrides ctx.attr.symlinks
-    output: File to use as output for script to load docker image
     env: str Dict, overrides ctx.attr.env
     layers: label List, overrides ctx.attr.layers
     debs: File list, overrides ctx.files.debs
     tars: File list, overrides ctx.files.tars
+    output_executable: File to use as output for script to load docker image
+    output_image: File, overrides ctx.outputs.out
```
mattmoor: Can we rename this `output_tarball` for clarity

xingao267: Done.

xingao267:
Hi Matt, this is to enable 2 things:

1. passing a different base image to _container.image.implementation(ctx)
For example, new_base is a tarball file resulted from another rule, and I would like to use this file as the base for the container_image rule, I can do  _container.image.implementation(ctx, base=my_base).

2. concatenate multiple _container.image.implementation(ctx) in another rule. 
For example, I can achieve something like this:

intermediate_tarball = ctx.actions.declare_file("intermediate_tarball.tar")
intermediate_layer = ctx.actions.declare_file("intermediate_layer.tar")
intermediate_executable = ctx.actions.declare_file("intermediate_executable.sh")
 _container.image.implementation(ctx, output_tarball=intermediate_tarball, output_layer=intermediate_layer, output_executable=intermediate_executable, ...)

# Do something else here...

 _container.image.implementation(ctx, base=intermediate_tarball, ...)

...

xingao267:
Friendly ping~

mattmoor:
I think my primary concern is how complex things are getting.  I think I understand what you are doing, but am uncertain why a macro isn't adequate.  I assume you have some sort of internal use-case.  If you can share more privately, I'd appreciate it.

container/layer_tools.bzl
```diff
@@ -36,8 +36,10 @@ def _extract_layers(ctx, artifact):
       "legacy": artifact,
   }
 
-def get_from_target(ctx, attr_target, file_target):
-  if hasattr(attr_target, "container_parts"):
+def get_from_target(ctx, attr_target, file_target, overriding_file_target=None):
+  if overriding_file_target:
+    return _extract_layers(ctx, overriding_file_target)
+  elif hasattr(attr_target, "container_parts"):
```
mattmoor: I find this part of the change awkward / hard to reason about.

It is already awkward because of Bazel's `ctx.attr` vs. `ctx.file` split.

Why isn't the value passed to `overriding_file_target` simply passed to `file_target` instead?  What can we do to improve this interface?

xingao267: ctx.attr.base.files.to_list() seems to give the same info as ctx.files.base. I refactored the function a little bit, and now the single file_target passed to the function will override the rest. Let me know what do you think.

xingao267:
Thanks for the review! I've rebased.

PR #342, opened 2018-03-06
node.js: actually put node_modules in separate layer + export app_layer_impl

hchauvin:
Because the run files in app_layer_impl was not overridden before this
commit, the node_modules folder for a node.js image was not actually
put in a different layer.

I also updated the dependency on rules_nodejs to HEAD because of a bug with the npm version.


nlopezgi:
Hi @hchauvin, is this PR still needed? Can you please address the feedback when you have the chance? Thanks for your patience and contribution!

nlopezgi:
Please re-open when you are ready to pick this up. Thanks again for your contributions!

PR #343, opened 2018-03-06
extend logic in _layer_emptyfile_path to layer_file_path

hchauvin:
The behavior of layer_file_path fails to give the proper symlink if in the dep_layer there is no mention of the workspace path (it only has external dependencies).

Indeed, if "wspace" is the workspace name, then the short name of an external dependency is "../" and the present algorithm has "wspace/../" in the symlink.  Since the "wspace" folder does not exist, the symlink is not resolved.


xingao267:
Hi @hchauvin, sorry for the delay.

Could you give an example target of the scenario you described? I tried to reproduce the error you got, but I couldn't. I'm not sure I understand the scenario you described where there is no mention of the workspace path.

nlopezgi:
Hi @hchauvin, is this PR still needed? Can you please address the feedback when you have the chance? Thanks for your patience and contribution!

nlopezgi:
Please re-open when you are ready to pick this up. Thanks again for your contributions!

PR #345, opened 2018-03-07
Replaced git_repository usages by http_archive and updated rules_nodejs

tassis-metavision:
- Replaced `git_repository` by `http_archive` in a loop.
- Updated the `rules_go` to the most recent version.
- Updated the `rules_nodejs` since it was broken.

I downloaded each <commit>.tar.gz from the repos and calculated the sha256 hashes manually.

Cheers.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -80,39 +80,52 @@ maven_jar(
     sha1 = "cce0823396aa693798f8882e64213b1772032b09",
 )
 
+LANGUAGES = [
+    {
+        "name": "scala",
+        "commit": "0bac7fe86fdde1cfba3bb2c8a04de5e12de47bcd",
+        "sha256": "f5f35de94d2d64e48fb4aef87cf89248b8980cb25f9ff1449575af8d904f41be",
+    },
+    {
+        "name": "groovy",
+        "commit": "6b8e32ce0f7e33ae1b859706c2dc0c169b966e7e",
+        "sha256": "ced0b80f4c32805abc61592eb2e5bd3e1ab809abdc24e39f888ce33055bfee5a",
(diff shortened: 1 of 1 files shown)


WORKSPACE
```diff
@@ -80,77 +80,77 @@ maven_jar(
...
+        "sha256": "527908e02d7bccf5a4eb89b690b003247eb6c57d69cc3234977c034d27c59d6e",
+    },
+]
+
+[http_archive(
+    name = "io_bazel_rules_{}".format(l["name"]),
+    sha256 = l["sha256"],
+    url = "https://github.com/bazelbuild/rules_{}/archive/{}.tar.gz".format(l["name"], l["commit"]),
+    strip_prefix = "rules_{}-{}".format(l["name"], l["commit"]),
+) for l in LANGUAGES]
```
mattmoor: I have mixed feelings about the loop vs. just unrolling it.  I think it ultimately ends up being roughly the same number of lines, but takes a bit more parsing to know exactly what's going on.

I understand that `http_archive` can result in better build performance than `git_repository`, but I have seen at least one instance where things broke because GitHub changed something about the way they produce these archives.  I'm also not a particularly big fan of having to juggle two SHAs instead of one.

I think these are distractions from the core change you are after.  I'm happy to discuss the benefits of `http_archive`, but perhaps you could limit this first change to the `rules_go` and `rules_nodejs` updates?

erain:
Hi @tassis-metavision : Thanks for the contribution. I understand `http_achieve` is more recommended so I implemented in #401 , so I guess this PR is for the same purpose and is not needed anymore now. 

I am going to close this PR for now.

PR #347, opened 2018-03-08
Improve image layering (Based on #324)

fahhem:
Based on #324, but updated to be based on HEAD. The only review comment I handled was the transient -> transitive one. The rest I'll handle iff @scele isn't interested. If he is, then he can use my PR as a free "git merge master" replacement.


fahhem:
Looks like @scele is responding, so I'll close this to avoid spamming the other PR

PR #354, opened 2018-03-16
Support the args attribute on app image rules

### LINKED ISSUE #304: Allow specifying args to containers in the BUILD file
Currently, if you define something like:

```
  go_image(
    name = "something_image",
    importpath = "unused",
    embed = embed,
    args = ["--port=8080"],
  )
```

then you are able to run: `bazel run :something_image.binary` and bazel will add the `--port=8080` to the binary and run it for you.

This behavior is defined for `*_binary`. The documentation says "Most binary rules permit an args attribute, but where this attribute is not allowed, this fact is documented under the specific rule.".

Unfortunately, this doesn't seem to work when the container itself is built and run: the CMD and/or Entrypoint in the container specification are not updated with the specified args.

Would it be acceptable for rules_docker to respect the `args` parameter add it to the container arguments' in the container image?

(originally issued at https://github.com/bazelbuild/rules_go/issues/1296 )

pcj:
This change propagates args from application
image rule macros into the app_layer rule.

Arguments passed in this way become appended
to the container entrypoint.

Fixes #304


pcj:
Open to suggestions on whether it is best to append args to the entrypoint or to cmd. 

My reasoning was that args on the entrypoint represent more "baked in" metadata.

cc @drigz @mattmoor

pcj:
Done.  `image_test` now includes more diverse lang_image inputs, so this will slow down the test a bit.

drigz:
Thank you @pcj! Deleting a workaround is a nice job for a Monday :+1:

PR #356, opened 2018-03-20
3rd party repos: Use http_archive rule for `containerregistry` remote.

mickeyzoox:
This was formerly a git_repository rule, which is not recommended (see
https://docs.bazel.build/versions/master/be/workspace.html#git_repository).
It also seemed to break the build for me in Bazel 0.7.


container/container.bzl
```diff
@@ -27,7 +27,7 @@ container = struct(
 )
 
 # The release of the github.com/google/containerregistry to consume.
-CONTAINERREGISTRY_RELEASE = "v0.0.26"
```
erain: Other part of this PR LGTM. Only this one thing: 
Any reason we want to change the format of version number, i.e. "v<version>" to "<version>"? 

As you can see we have the single letter 'v' repeated multiple times in the URLs below, which doesn't look more readable for me.

mickeyzoox: I moved this into the urls in order to make the `strip_prefix = "containerregistry-" + CONTAINERREGISTRY_RELEASE` line cleaner.

I've changed it back to the old way. Please let me know if you'd like me to make any other changes.

mickeyzoox:
@erain Please let me know if there are any more steps on my part to get this merged.

There seems to be a CI failure but it look unrelated to this change.

erain:
Hi @mickeyzoox : we've already fixed the CI. Could you please rebased your and it should be ready for merge?

mickeyzoox:
I have rebased and squashed. It should be ready for merge now

erain:
Sorry for coming back to this a little late. The TravisCi failed because of bzl files's formatting: 
> The command "buildifier -mode=check $(find . -name '*.bzl' -type f)" exited with 4.

Could you try to run buildifier with your changes? The setup is simple, see:
  https://github.com/bazelbuild/rules_docker/blob/e5ebe3d241775a220e37aceaa24c3e78700a4e0b/.travis.yml#L29
And remove  `-mode=check` will automatically fix the formatting for you.

mickeyzoox:
The violations (of buildifier formatting) are not in the file that I've changed, they are in `master`.

I've rebased again (it seems that `master` was recently fixed), so I expect that this time the branch will pass CI.

erain:
Thanks!

Also could please you merge from the master? Otherwise I cannot approve/merge it directly.

Issue #155, opened 2017-09-20
Set creation time to be {SOURCE_DATE_EPOCH} when stamping is enabled


mattmoor:
The release of `google/containerregistry` that we are on supports overriding `creation_time`, and I am adding stamping support to `docker_build` more broadly in https://github.com/bazelbuild/rules_docker/pull/154.

We should consider either:
1. Making `{SOURCE_DATE_EPOCH}` the default when `stamp=True`, or
1. Enabling users to specify a `creation_time`, which can be `{SOURCE_DATE_EPOCH}`.

mattmoor:
@damienmg Is `{SOURCE_DATE_EPOCH}` supported at all now?

mattmoor:
If I just tweak my PR to use this, I get:
```shell
<explosion>
KeyError: 'SOURCE_DATE_EPOCH'
```

So it definitely isn't in 0.5, maybe 0.6...?  Is there an issue we can link this to to cover support?

damienmg:
https://github.com/bazelbuild/bazel/issues/2240

damienmg:
Also note that inside the volatile info text file it will always be BUILD_TIMESTAMP (not SOURCE_DATE_EPOCH).

ixdy:
Bazel 0.9 uses `SOURCE_DATE_EPOCH` to set the `BUILD_TIMESTAMP`, so we might be able to look into using this soon?

leifmadsen:
Just wanting to ping this issue, as I'm still seeing containers from 47 years ago, and it isn't clear how to override that (or if it can yet).

mattmoor:
@leifmadsen I assume @ixdy is right, and once Bazel 0.9 cuts we can switch "stamped" docker builds to use this stamp variable in its timestamp.

ixdy:
what's still blocking us here? it seems like it should be safe to use `BUILD_TIMESTAMP` to set `creation_time` when stamping is enabled, right? Environments that want repeatable builds should be able to set `SOURCE_DATE_EPOCH` before calling bazel and have it do the right thing.

mattmoor:
Nothing AFAIK.

ixdy:
I think there might still be missing functionality - the language specific image rules (such as `go_image`) don't seem to timestamp properly.

nlopezgi:
[...]

PR #364, opened 2018-03-24
Allow for users to declare explicit build timestamps

dekkagaijin:
Fixes: #155 

Signed-off-by: Jake Sanders <jsand@google.com>

### DIFF
--- a/container/create_image_config.py
+++ b/container/create_image_config.py
@@ -13,6 +13,8 @@
 # limitations under the License.
 """This package manipulates v2.2 image configuration metadata."""
 
+from __future__ import division
+
 import argparse
 import datetime
 import json
@@ -43,7 +45,8 @@ parser.add_argument('--command', action='append', default=[],
 
 parser.add_argument('--creation_time', action='store', required=False,
                     help='The creation timestamp. Acceptable formats: '
-                    'Integer milliseconds since Unix Epoch, RFC 3339 date/time')
+                    'Integer or floating point seconds since Unix Epoch, RFC '
+                    '3339 date/time')
 
 parser.add_argument('--user', action='store',
                     help='The username to run commands under.')
@@ -130,17 +133,20 @@ def main():
   if args.creation_time:
     creation_time = Stamp(args.creation_time)
     try:
-      # If creation_time is parsable as an int type, assume unix epoch timestamp.
-      if six.PY3:
-        parsed_unix_timestamp = int(creation_time) // 1000
-      else:
-        parsed_unix_timestamp = int(long(creation_time) / 1000)
+      # If creation_time is parsable as a floating point type, assume unix epoch
+      # timestamp.
+      parsed_unix_timestamp = float(creation_time)
+      if parsed_unix_timestamp > 1.0e+11:
+          # Bazel < 0.12 was bugged and used milliseconds since unix epoch as
+          # the default. Values > 1e11 are assumed to be unix epoch
+          # milliseconds.
+          parsed_unix_timestamp = parsed_unix_timestamp / 1000.0
 
       # Construct a RFC 3339 date/time from the Unix epoch.
       creation_time = (
           datetime.datetime.utcfromtimestamp(
               parsed_unix_timestamp
-          ).strftime("%Y-%m-%dT%H:%M:%SZ")
+          ).strftime("%Y-%m-%dT%H:%M:%S.%fZ")
       )
     except ValueError:
       # Otherwise, assume RFC 3339 date/time format.
--- a/container/create_image_config.py
+++ b/container/create_image_config.py
@@ -43,7 +43,7 @@ parser.add_argument('--command', action='append', default=[],
 
 parser.add_argument('--creation_time', action='store', required=False,
                     help='The creation timestamp. Acceptable formats: '
-                    'Milliseconds since Unix Epoch, RFC 3339 date/time')
+                    'Integer milliseconds since Unix Epoch, RFC 3339 date/time')
 
 parser.add_argument('--user', action='store',
                     help='The username to run commands under.')
--- a/container/create_image_config.py
+++ b/container/create_image_config.py
@@ -14,6 +14,7 @@
 """This package manipulates v2.2 image configuration metadata."""
 
 import argparse
+import datetime
 import json
 import sys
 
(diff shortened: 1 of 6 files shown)


dekkagaijin:
Can do the `stamp=True` implementation, as well, but didn't know if there was one preference over the other.

container/create_image_config.py
```diff
@@ -121,10 +125,30 @@ def Stamp(inp):
...
   output = v2_2_metadata.Override(data, v2_2_metadata.Overrides(
-      author='Bazel', created_by='bazel build ...',
-      layers=layers, entrypoint=list(map(Stamp, fix_dashdash(args.entrypoint))),
-      cmd=list(map(Stamp, fix_dashdash(args.command))), user=Stamp(args.user),
+      author='Bazel',
+      created_by='bazel build ...',
+      layers=layers,
+      entrypoint=list(map(Stamp, fix_dashdash(args.entrypoint))),
+      cmd=list(map(Stamp, fix_dashdash(args.command))),
+      creation_time=creation_time,
```
mattmoor: This doesn't fix the issue unless users can use: `{SOURCE_DATE_EPOCH}` (or w/e the magic stamp variable @ixdy wants is) for this value.

testdata/BUILD
```diff
@@ -244,6 +244,18 @@ container_image(
...
+container_image(
+    name = "with_unix_epoch_creation_time",
+    base = ":base_with_volume",
+    creation_time = "610203480",
+)
+
+container_image(
+    name = "with_rfc_3339_creation_time",
+    base = ":base_with_volume",
+    creation_time = "1989-05-03T12:58:00Z",
```
mattmoor: This kwarg doesn't feel particularly valuable unless its value can be `{SOURCE_DATE_EPOCH}`.  In fact, I wasn't even imagining exposing this, just making `{SOURCE_DATE_EPOCH}` the default when stamping is enabled.

mattmoor:
Thanks for picking this up.

testdata/BUILD
dekkagaijin: My understanding was that {SOURCE_DATE_EPOCH} would be set as an environment variable, which could be substituted in the build attributes, but I'd believe that I misunderstood the conversation in #155 .

I'll do the `stamp`-based solution.

dekkagaijin: I was imagining that people could define their own values, or use `--define` to inject their own values here, without going through `stamp = True` and `SOURCE_DATE_EPOCH`.

container/create_image_config.py
dekkagaijin: They can, kind of, though `$(SOURCE_DATE_EPOCH)`  is seconds since the unix epoch and BUILD_TIMESTAMP is in milliseconds.

dekkagaijin:
De nada...

Anyway, done. FYI `bazel test` and `bazel build` have started failing locally within the last week.

container/create_image_config.py
ixdy: right, as I understand it, `BUILD_TIMESTAMP` is normally set to the current time, but if `SOURCE_DATE_EPOCH` is set, Bazel sets `BUILD_TIMESTAMP` to that value instead.

tl;dr I think this is doing what we want

testdata/BUILD
ixdy: maybe also update the `README.md` describing the `creation_time` arg?

container/create_image_config.py
ixdy: (it's confusing because they're both yelling case, but `SOURCE_DATE_EPOCH` is an environment variable Bazel recognizes, but it isn't set in the workspace status; only `BUILD_TIMESTAMP` is (unless you set it yourself))

testdata/BUILD
```diff
@@ -244,6 +244,25 @@ container_image(
...
+)
+
+container_image(
+    name = "with_stamped_creation_time",
+    base = ":base_with_volume",
+    creation_time = "{BUILD_TIMESTAMP}",
```
ixdy: I sort of wonder whether we want to allow the flexibility of choosing a particular workspace status variable, or if we should just always use `BUILD_TIMESTAMP` when stamping is enabled.

mattmoor: That's kind of what I had in mind as well.

ixdy: `SOURCE_DATE_EPOCH` is the standard proposed by Reproducible Builds: https://reproducible-builds.org/specs/source-date-epoch/

It makes sense to me to standardize on `BUILD_TIMESTAMP` since that's likely what other rules might use as well.

ixdy: I feel that giving the option to specify a different workspace status variable is unnecessary.

I don't have a strong opinion about allowing hard-coded values here, though at least for Kubernetes we plan to just set `SOURCE_DATE_EPOCH` for reproducibility (and cacheability).

mattmoor: I'm fine with allowing users to set this, but let's default to `{BUILD_TIMESTAMP}` when `stamp=True` so that they don't have to?

dekkagaijin: Works for me. Done.

testdata/BUILD
dekkagaijin: @ixdy Done.

dekkagaijin:
@mattmoor ping <3

container/create_image_config.py
```diff
@@ -121,10 +126,34 @@ def Stamp(inp):
     elif '{' in value:
       labels[label] = Stamp(value)
 
+  creation_time = None
+  if args.creation_time:
+    creation_time = Stamp(args.creation_time)
+    try:
+      # If creation_time is parsable as an int type, assume unix epoch timestamp.
```
ixdy: oh good, you handled the breaking behavior soon to arrive in bazel (https://bazel-review.googlesource.com/c/bazel/+/48211).

maybe add a TODO to fix this before 2038. :)

dekkagaijin: T_T
I'll wait until the change lands

dekkagaijin: Should I make this change require bazel 0.12? Or should I just create a reasonable test to support both, e.g. parseable as float -> unix timestamp, >1e10 -> milliseconds, <= 1e10 -> seconds

ixdy: I'd vote for "gracefully" handling both forms of timestamps, rather than requiring bazel 0.12+.

dekkagaijin: Done.

dekkagaijin:
@mattmoor ping now that you're back from vacation

PR #365, opened 2018-03-26
Expand location for args and symlink data dependencies

Globegitter:
I was happy to see the recent `args` commit and started using it by wanting to pass in a path to a file the container is depended on via the `data` field. To get the location of this file it is common to use the `$(location ...)` variable which is not supported here. I added support for that but then also noticed that the binary does not have access to that file due to the symlinking. So I also added symlinking for data dependencies and now everything works in our project.

I have not worked much with bzl rules before so especially the symlink adding I added just in a quick way and can probably be done much cleaner. Also there is no test to check the existence of the symlink. If that is something that would be accepted I would be happy to clean this up a bit. Also interestingly enough it seems `d_binary` and `nodejs_binary` are having some problem either with `data` or `location` so these two do currently not support this functionality.

### DIFF
--- a/lang/image.bzl
+++ b/lang/image.bzl
@@ -207,18 +207,17 @@ def _app_layer_impl(ctx, runfiles=None, emptyfiles=None):
     # root, since they may be accessed via either path.
     _external_dir(ctx): _runfiles_dir(ctx),
   })
+
   for data in ctx.attr.data:
-    print(data.label.name)
     loc = ctx.expand_location('$(location :' + data.label.name + ')', ctx.attr.data)
     data_path = "/".join([
       ctx.attr.directory,
       ctx.label.package,
       data.label.name
-  ])
+    ])
     symlinks[data_path] = '/'.join([_runfiles_dir(ctx), '__main__', loc])
-  args = [ctx.expand_location(arg, ctx.attr.data) for arg in ctx.attr.args]
 
-  print(symlinks)
+  args = [ctx.expand_location(arg, ctx.attr.data) for arg in ctx.attr.args]
 
   return _container.image.implementation(
     ctx,
--- a/lang/image.bzl
+++ b/lang/image.bzl
@@ -247,6 +247,7 @@ app_layer = rule(
         "workdir": attr.string(default = "/app"),
         "directory": attr.string(default = "/app"),
         "legacy_run_behavior": attr.bool(default = False),
+        "data": attr.string_list(default = []),
     }.items()),
     executable = True,
     outputs = _container.image.outputs,
--- a/lang/image.bzl
+++ b/lang/image.bzl
@@ -207,6 +207,10 @@ def _app_layer_impl(ctx, runfiles=None, emptyfiles=None):
     # root, since they may be accessed via either path.
     _external_dir(ctx): _runfiles_dir(ctx),
   })
+  for data in ctx.attr.data:
+    symlinks['/'.join(_runfiles_dir(ctx), '__main__', ctx.expand_location(data, ctx.attr.data))] = _runfiles_dir(ctx)
+  # for data in ctx.attr.data:
+  #   symlinks["/app/" + data] = _final_file_path(ctx, data)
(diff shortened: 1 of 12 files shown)


Globegitter:
I signed it!

Globegitter:
Travis is failing on go get, so as far as I can see the error is not related to this PR.

lang/image.bzl
```diff
@@ -208,6 +208,17 @@ def _app_layer_impl(ctx, runfiles=None, emptyfiles=None):
     _external_dir(ctx): _runfiles_dir(ctx),
   })
 
+  for data in ctx.attr.data:
```
nlopezgi: You should not need to create these symlinks. The files that you are adding via data are already getting copied into the container under the directory:
/app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/
(e.g., the container includes the file /app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/testdata/BUILD)
The issue you have is that expand_location was not meant to be used in this way (it resolves the location of the file wrt the root of the runfiles tree, e.g., /app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/), I'm suspecting your expectation is to just use the arg "testdata/BUILD" as the location of the BUILD file relative to the "/app" directory. Could you include an example of how you expect to use the arguments as part of the execution of the main command in the image so that we can try to figure out a way to properly get this working?

Globegitter: Yeah it kind of seemed like this is not quite the right thing. The use-case is, we have a `par_binary` that reads in a config and a secret key file, specified as a cli arg and it is possible to just define this as follows:

```
filegroup(
    name = "config",
    srcs = [
        "config/thumbor.conf",
    ],
)

filegroup(
    name = "secrets",
    srcs = [
        "secrets/thumbor.key",
    ],
)

par_binary(
    name = "thumbor_server",
    srcs = [
        "thumbor_server.py",
    ],
    args = [
        "-c",
        "$(location config)",
        "-a",
        "app.App",
        "--keyfile",
        "$(location :secrets)",
    ],
    data = [
        ":config",
        ":secrets",
    ],
    default_python_version = "PY2",
    deps = [
        ":app",
        requirement("thumbor"),
    ],
)
```

I expected exactly the same thing to work for `py_image` so:
```
py_image(
    name = "image",
    srcs = ["thumbor_server.py"],
    args = [
        "--app",
        "app.App",
        "--conf",
        "$(location :config)",
        "--keyfile",
        "$(location :secrets)",
    ],
    base = "//common/docker:py2_libcurl_image",
    data = [
        ":config",
        ":secrets",
    ],
    main = "thumbor_server.py",
    deps = [
        ":app",
        requirement("thumbor"),
    ],
)
```

the location returns the relative path`config/thumbor.conf` which works in the case for the binary but not for the docker image.

Is that example clear enough? What would be the right way to go about this?

nlopezgi: Thanks for the clarifying example. I think that the issue we have is that the workdir is not being properly set on the _container.image.implementation call below. The problem is, that you cant override atm the workdir, it seems (https://github.com/bazelbuild/rules_docker/blob/master/container/image.bzl#L230). Adding that new attr to the image impl should not be hard though (you would just need to plumb it down so that it is passed to _image_config and used here: https://github.com/bazelbuild/rules_docker/blob/master/container/image.bzl#L149 from the attr and not from the ctx). Once you have that you should be able to set the workdir in the call in line 22 to 
workdir = '/'.join([_runfiles_dir(ctx), ctx.workspace_name]) 
(i.e., it should resolve to "./testdata/py_image_with_main.binary.runfiles/io_bazel_rules_docker"). Once you have that workdir set, this should work automatically with your example without the need for any additional symlinks. Please try it out and let me know if it works.

Globegitter: @nlopezgi this makes mostly sense to me. But where should I set `workdir = '/'.join([_runfiles_dir(ctx), ctx.workspace_name])` exactly? Can not seem to find which line 22 you are referring to. Would be happy to try and implement a fix for it otherwise.

Globegitter: I think I found it and submitted a new PR: #452

PR #373, opened 2018-04-04
Make non-hermetic gzip/gunzip more portable

benley:
This ought to make the gzip and gunzip actions work on systems where the gzip binary is in a path that would not normally show up in the action sandbox.

For example, on NixOS the gzip binary is typically at `/run/current-system/sw/bin/gzip` (or possibly elsewhere if a user has installed it into their own profile).  While these locations are definitely present in `$PATH`, Bazel does not normally make them available in execution sandboxes.

There should be no noticeable difference on any system where these rules were already working; this change broadens support and makes the non-hermetic steps explicit.

### DIFF
--- a/container/container.bzl
+++ b/container/container.bzl
@@ -33,6 +33,32 @@ CONTAINERREGISTRY_RELEASE = "v0.0.26"
 # Updated around 1/22/2018.
 STRUCTURE_TEST_COMMIT = "b97925142b1a09309537e648ade11b4af47ff7ad"
 
+_local_tool_build_template = """
+sh_binary(
+    name = "{name}",
+    srcs = ["bin/{name}"],
+    visibility = ["//visibility:public"],
+)
+"""
+
+def _local_tool(repository_ctx):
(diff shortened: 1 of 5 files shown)


erain:
Hi @benley : sorry for coming to this PR late (as we just started maintaining this PR and had some backlogs). 

This change LGTM, so it is possible for you to rebased on the latest branch (where there some changes in `buildifier` about skylark formatting)  and I could merge this PR? 

The particular `buildifier` setup and commands (remove `--mode=check` will reformat files automaticially):
``` bash
go get -u github.com/bazelbuild/buildifier/buildifier
buildifier -mode=check $(find . -name BUILD -type f)
buildifier -mode=check $(find . -name WORKSPACE -type f)
buildifier -mode=check $(find . -name '*.bzl' -type f)
```

Many thanks!

benley:
Sure thing, I will get the PR updated today.  Thanks!

benley:
Updated, looks like buildifier and the test suites are happy again.

PR #390, opened 2018-04-18
Updated Go documentation to include goarch, goos and pure parameters

f0rmiga:
When using the `go_image` in an OS other than Linux, those parameters are mandatory.


erain:
Could you also explain in the document that `goarch`, `goos` and `pure` should be specify for other values when used in other OS? 

The current explanation looks to me that go_image can be ONLY used in Linux

Issue #394, opened 2018-04-27
Switch setup example to http_archive


ash2k:
[Bazel 0.12.0 release notes](https://blog.bazel.build/2018/04/11/bazel-0.12.html) say:
> The native http_archive rule has been deprecated. Use the Skylark version available via load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive") instead.
>
> The native git_repository rule has been deprecated. Use the Skylark version available via load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository") instead.

[Documentation for native `git_repository`](https://docs.bazel.build/versions/master/be/workspace.html#git_repository) says (follow the link for the details):
> Warning: this rule has several limitations. We recommend using http_archive instead for more robust and secure performance.

Example should be something like this:
```bzl
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")

http_archive(
    name = "io_bazel_rules_docker",
    sha256 = "<download and calculate it manually>",
    strip_prefix = "rules_docker-553d5506bb7325185950f91533b967da8f5bc536",
    url = "https://github.com/bazelbuild/rules_docker/archive/553d5506bb7325185950f91533b967da8f5bc536.zip",
)
```

erain:
I already sent a #396 for fixing this issue, and also #401 for fixing `WORKSPACE` file for real. 

 @mpassell : I don't see anything in project settings about this. I guess this is related to how a "Release" is drafted, and I will remember to include the tar file downloads for the next release.

PR #396, opened 2018-05-03
Use http_archive instead of git_repository in examples

erain:
Fixes: #394


erain:
@xingao267 Could you help review this?

xingao267:
Do we still need https://github.com/bazelbuild/rules_docker/pull/401?

PR #400, opened 2018-05-08
Update the Travis base to fix Travis CI

### LINKED ISSUE #397: Travis CI is broken
Current Travis CI is broken with the error message: 

```
$ go get -u github.com/bazelbuild/buildifier/buildifier
# github.com/bazelbuild/buildtools/build
../../../gopath/src/github.com/bazelbuild/buildtools/build/lex.go:818: undefined: sort.Slice
The command "go get -u github.com/bazelbuild/buildifier/buildifier" failed and exited with 2 during .
```

We need to either fix it, or disable it to use Bazel CI only.

erain:
Fix #397 

- The current error message, `Undefined Sort.SliceStable`, was
  introduced in Go version 1.8
  https://beta.golang.org/doc/go1.8#sort_slice

- It looks like the tusty images doesn't have go in common
  environment:
  https://docs.travis-ci.com/user/reference/trusty/#Environment-common-to-all-Trusty-images

- On the other hand, precise has go in common environment, and with
  default to 1.8.1:
  https://docs.travis-ci.com/user/reference/precise/#Environment-common-to-all-Precise-images

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,6 +14,9 @@ language: go
 go: 1.10.x
 install: true
 
+services:
+  - docker
+
 addons:
   apt:
     sources:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: precise
 language:
   - java
 jdk:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,19 @@
+dist: trusty
 sudo: required
-dist: precise
-language:
-  - java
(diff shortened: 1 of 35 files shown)


erain:
It looks like downgrading to precise was not the right thing to do, nor fixed the problem. 

The only way we can upgrade go > 1.8 is to install one manually. Will send a new commit for that.

erain:
After fixing format issue in .bzl files, now travis CI becomes green. 

@xingao267 could you help review this PR?

Issue #217, opened 2017-11-05
testing/e2e.sh fails on Mac OS X


Capstan:
This quickly fails due to a missing `/bin/true` on line 42 of testing/e2e.sh. Swapping that out for `/usr/bin/true` goes much further but still fails:

```
____Loading complete.  Analyzing...
____Loading package: @rust_default_toolchains//
____Loading package: @cc_image_base//image
____Found 1 target...
____Building...
Target //testdata:cc_image up-to-date:
  bazel-bin/testdata/cc_image-layer.tar
____Elapsed time: 8.804s, Critical Path: 1.76s

____Running command line: bazel-bin/testdata/cc_image
standard_init_linux.go:185: exec user process caused "exec format error"
ERROR: Non-zero return code '1' from command: Process exited with status 1.
Checking Loaded image ID: sha256:09069941bff078c779c8c5cf1dc7576f35bf032f6b6b945e4e041fe29ea6a7b2
Tagging 09069941bff078c779c8c5cf1dc7576f35bf032f6b6b945e4e041fe29ea6a7b2 as bazel/testdata:cc_image contains Hello World
FAILURE: Expected 'Hello World' not found in 'Loaded image ID: sha256:09069941bff078c779c8c5cf1dc7576f35bf032f6b6b945e4e041fe29ea6a7b2
Tagging 09069941bff078c779c8c5cf1dc7576f35bf032f6b6b945e4e041fe29ea6a7b2 as bazel/testdata:cc_image'
```

System configuration:
* Mac OS X Sierra 10.12.6 (16G29)
* xCode 8.3.3 (8E3004b)
* Docker Version 17.09.0-ce-mac35 (19611) stable a98b7c1b7c
* bazel release 0.7.0-homebrew
* homebrew config:

```
HOMEBREW_VERSION: 1.3.6
ORIGIN: https://github.com/Homebrew/brew.git
HEAD: 270b752f5d9d218bfbed6fe85b6974fa653fb25f
Last commit: 3 weeks ago
Core tap ORIGIN: https://github.com/Homebrew/homebrew-core
Core tap HEAD: 9f0714bc010f55a59b07174dd5fcd6704ae88c13
Core tap last commit: 23 hours ago
HOMEBREW_PREFIX: /Users/nherring/homebrew
HOMEBREW_REPOSITORY: /Users/nherring/homebrew
HOMEBREW_CELLAR: /Users/nherring/homebrew/Cellar
[...]

Issue #73, opened 2017-06-29
Make testing/e2e.sh a bazel test...


damienmg:
So we can run it on Bazel.

We have docker test in Bazel, and we can depends on the bazel installer to get bazel inside the test so it should be possible.

mattmoor:
AIUI this is blocked on the Bazel issue about making writing e2e testing easier?  I did this specifically because doing this in Bazel was a PITA, but maybe things have evolved some.

damienmg:
Actually @dslomov suggested to just use a http_archive on the bazel installer to ship Bazel inside the repo. That has the advantage that the test can use the bazel for execution platform so it make the test ready for remote execution.

We also have another part about making the testing framework easier but that's does not prevent e2e.sh to run in Bazel by itself.

If I get some time in the summer I'll tackle this one.

ittaiz:
Just FYI- I tried this and it takes ~20 seconds just to do the installation of Bazel

damienmg:
https://bazel-review.googlesource.com/c/15432 is doing that. You only pay once the install of Bazel (but not the unzipping, we might want to do that ahead of time).

If you use --experimental_repository_cache, you will not repay the download accross build.

ittaiz:
@damienmg What in that CL does that? AFAICT the repository rule just downloads and extracts. If I remember correctly from my experiment the first time you run bazel after that you pay 20 seconds.
Am I incorrect?

damienmg:
20seconds seems a lot but that is true, you unpack bazel. What I said is we
can make that extraction part of the remote repository to not pay it using
--install_path.

ittaiz:
`--install_path`? Couldn't find it in the codebase or command line reference.
How can we pay the extraction in the repository rule? by just executing "bazel version" for example in the rule?

damienmg:
--install_base sorry. Yes we can simply unzip bazel itself.

ittaiz:
nice. Should I open an issue (in bazel obviously) about extracting it to another repo and about doing the unzip?

damienmg:
I think we can piggy back to the existing issue about testing skylark code?

ittaiz:
as usual you're right :)

ittaiz:
@mattmoor I think this is no longer blocked.
@damienmg released https://github.com/bazelbuild/bazel-integration-testing a while ago and we're iterating over it to add more features.
It has basic support for java, support for python (not sure at which state) and there's an active PR for go support.

[...]

PR #403, opened 2018-05-15
Update rules_nodejs to 0.9.1 and fix node target

nimerritt:
This https://github.com/bazelbuild/rules_nodejs/pull/193/files#diff-09498dbadf45966909850dc8a47ebb13R30 will break this rule when it is released.

This isn't yet released with `rules_nodejs` but will be in the next release.
@gregmagolan just aheads up.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -153,9 +153,9 @@ d_repositories()
 
 http_archive(
     name = "build_bazel_rules_nodejs",
-    sha256 = "a672bbb4eb8c49363942fe9a491f35214b5d7a0000c86e0152ea8cd3261b1c12",
-    strip_prefix = "rules_nodejs-0.8.0",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/archive/0.8.0.tar.gz"],
+    sha256 = "6139762b62b37c1fd171d7f22aa39566cb7dc2916f0f801d505a9aaf118c117f",
+    strip_prefix = "rules_nodejs-0.9.1",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/archive/0.9.1.zip"],
 )
 
 load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories", "npm_install")
--- a/nodejs/image.bzl
+++ b/nodejs/image.bzl
@@ -117,13 +117,14 @@ def nodejs_image(
 
     layers = [
         # Put the Node binary into its own layer.
-        "@nodejs//:bin/node",
+        "@nodejs//:node/bin/node",
         # node_modules can get large, it should be in its own layer.
         node_modules,
     ] + layers
 
     nodejs_binary(
         name = binary_name,
+        node = "@nodejs//:node/bin/node",
         node_modules = node_modules,
         data = data + layers,
         **kwargs


gregmagolan:
`@nodejs//:node` will now work for all platforms and it sets the environment so that `node` will execute the bazel managed node.

nimerritt:
@gregmagolan breaking change isn't a problem for us as we're generally on the latest commit with the rules at the moment, but it might be worth noting which rules_docker versions work with which rules_nodejs. It would be nice if this PR gets merged though so we can drop our fork :)

gregmagolan:
@nimerritt `external/nodejs/node/bin/node` is the location of the actual node binary which is downloaded by the node_repositories() repository rule. Is this failure local or on CI? Locally you could try `bazel clean --expunge` first, which clears out the external repositories. Bazel should see that rules_nodejs has changed and handle that but there may be a bug in Bazel.

gregmagolan:
`external/nodejs/repository_bin/node` is a shell script that sets the PATH in the environment and then calls `external/nodejs/node/bin/node`

gregmagolan:
@nimerritt I tried `bazel build ... && bazel test ...` on a fresh checkout of rules_docker with the changes in this PR and everything passed. Ran on a mac.

nimerritt:
@gregmagolan The test isn't actually run with `bazel test ...` but rather Travis CI runs `./testing/e2e.sh` which fails both locally (mac clean install) and on the CI with the same error.
Do I need to include more targets in the node layer maybe?

nimerritt:
Inspecting the image built by the e2e test, `bazel build //testdata:nodejs_image` shows that the node binary isn't copied into the image any more. With the 0.9.0 changes the image looks like:
```
.
└── app
    └── testdata
        ├── nodejs_image.binary -> /app/testdata/nodejs_image.binary.runfiles/io_bazel_rules_docker/testdata/nodejs_image.binary
        └── nodejs_image.binary.runfiles
            ├── io_bazel_rules_docker
            │   ├── external -> /app/testdata/nodejs_image.binary.runfiles
            │   └── testdata
            │       ├── nodejs_image.binary
            │       ├── nodejs_image.binary_bin
            │       ├── nodejs_image.binary_bin.sh
            │       ├── nodejs_image.binary_bin_loader.js
            │       └── nodejs_image.js
            ├── nodejs
            │   ├── args.sh
            │   └── bin
            │       └── node
            └── npm_deps
                └── node_modules
                    └── jsesc
                        ├── LICENSE-MIT.txt
                        ├── README.md
                        ├── bin
                        │   └── jsesc
                        ├── jsesc.js
                        ├── man
                        │   └── jsesc.1
                        └── package.json
```

compared to the previously working image,

```
.
└── app
    └── testdata
        ├── nodejs_image.binary -> /app/testdata/nodejs_image.binary.runfiles/io_bazel_rules_docker/testdata/nodejs_image.binary
        └── nodejs_image.binary.runfiles
            ├── io_bazel_rules_docker
            │   ├── external -> /app/testdata/nodejs_image.binary.runfiles
            │   └── testdata
            │       ├── nodejs_image.binary
            │       ├── nodejs_image.binary_bin
            │       ├── nodejs_image.binary_bin.sh
            │       ├── nodejs_image.binary_bin_loader.js
            │       └── nodejs_image.js
            ├── nodejs
            │   └── repository_bin
            │       ├── node
            │       └── node_args.sh
            └── npm_deps
                └── node_modules
                    └── jsesc
                        ├── LICENSE-MIT.txt
                        ├── README.md
                        ├── bin
                        │   └── jsesc
                        ├── jsesc.js
                        ├── man
                        │   └── jsesc.1
                        └── package.json
```

nimerritt:
@gregmagolan I think I figured out the problem, but not really sure the best way to fix it since I'm not familiar enough with the rules_docker and rules_nodejs. As far as I've understood, the target `@nodejs//:node` doesn't include all the runtime files needed to actually run the node binary but rather assumes that everything is present in the external folder.

gregmagolan:
@nimerritt Ok.  That's makes sense. Thanks for tracking that down. All of the required files need to be included in layers to build the docker images. There is a shell script and two javascript files that need to be included on top of the node binary itself. I missed exporting one of them out of @nodejs so I'll have to make a quick PR to fix that. Once I get that in then the following in rules_docker should work:

```layers = [
        # Put the Node binary into its own layer.
        "@nodejs//:bin/node",
        "@nodejs//:bin/node.js",
        "@nodejs//:bin/node_args.js",
        "@nodejs//:node/bin/node",
        # node_modules can get large, it should be in its own layer.
        node_modules,
    ] + layers
```
(note: renaming repository_bin to just bin--it used to conflict with bin which no longer exists and just bin is better)

gregmagolan:
@nimerritt Hmmm. Yes. That makes sense. I'll work on resolving that. How are you able to reproduce the failure? Running `bazel test ...` on rules_docker works for me so I'm assuming there is another test to run that fails?

nimerritt:
I think I fixed this (e2e test never runs on mac due to cross-compiling issues) in the short term by using the original `node/bin/node` target. I had to change the _node attribute to be public. I'm open for other suggestions :)

nimerritt:
@gregmagolan The `./testing/e2e.sh` test is the one that needs to pass. It just builds `//testdata:nodejs_image` and runs it and checks that `Hello world` is output.

nimerritt:
https://github.com/bazelbuild/rules_docker/issues/73 and https://github.com/bazelbuild/rules_docker/issues/217 explain the confusion. It would be nice if `bazel test //...` was all that was needed. It confused me for a bit.

gregmagolan:
> I think I fixed this (e2e test never runs on mac due to cross-compiling issues) in the short term by using the original node/bin/node target. I had to change the _node attribute to be public. I'm open for other suggestions :)

Yes. That's a better solution. Thanks! There is no need for the files from `@nodejs//:bin/**` in the layer that case since those files just end up calling `node/bin/node` after setting the PATH.

nimerritt:
Yeah, I'll remove those and update to the latest commit. Once a new release
is published we can use the `http_repository` rule and this can get merged
at some point.

Thanks for the help and quick responses! It's much easier to sell my
colleges on Bazel with the alpha node_js rules when we have the Angular
team helping us! :)

gregmagolan:
@nimerritt Thank you for your help! Tagged rules_nodejs at 0.9.1 with your change.

```
http_archive(
    name = "build_bazel_rules_nodejs",
    url = "https://github.com/bazelbuild/rules_nodejs/archive/0.9.1.zip",
    strip_prefix = "rules_nodejs-0.9.1",
    sha256 = "6139762b62b37c1fd171d7f22aa39566cb7dc2916f0f801d505a9aaf118c117f",
)
```

WORKSPACE
```diff
@@ -151,13 +151,20 @@ load("@io_bazel_rules_d//d:d.bzl", "d_repositories")
 
 d_repositories()
 
-http_archive(
+# Until rules_nodejs is updated
+git_repository(
```
erain: Could we keep http_archive rule instead of git_repository rule?

In [git_repository's documentation](Warning: this rule has several limitations. We recommend using http_archive instead for more robust and secure performance.)
> Warning: this rule has several limitations. We recommend using http_archive instead for more robust and secure performance.

nimerritt: Yeah, no problem. I had it there only until the rules_nodejs was actually tagged. I'll make a new PR with just the formatting changes. The CI fails without them, that's why I added them.

nimerritt:
This PR is rebased off https://github.com/bazelbuild/rules_docker/pull/406 FYI.

nimerritt:
@erain and @tejal29 this is now just the fixes for the new nodejs rules. Do you need any changes before merging this? Thank's for the quick turn-around time! :)

erain:
Thanks for the fix. I've approve the PR and am waiting for the tests to pass.

nimerritt:
Great thanks for merging! Time to get rid of our fork 👍

PR #410, opened 2018-05-24
Add file permissions attribute to passwd_tar

tejal29:
Forgot to add permissions mode to passwd_tar rule.
@erain one more favor from you!


tejal29:
Thanks! i dont have write access. I guess only @dlorenc has.

PR #416, opened 2018-06-04
Update structure_test to v1.2.2.

xingao267:

PR #419, opened 2018-06-05
Move container structure test pin to fix entrypoint.

xingao267:
Pin to https://github.com/GoogleContainerTools/container-structure-test/commit/7411da1d4cb505f1a70ae5d1ca3d961a3fa66cda. This is required to test any image with an entrypoint other than bash or sh.


xingao267:
Let's merge this in first instead of https://github.com/bazelbuild/rules_docker/pull/416

PR #421, opened 2018-06-06
Add rename_image

ArthurRab:
Added rename_image macro and a test for it


contrib/rename_image.bzl
```diff
@@ -0,0 +1,29 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("//container:bundle.bzl", "container_bundle")
+
+def rename_image(image, name):
```
xingao267: Add a comment here the macro is migrated from https://github.com/GoogleContainerTools/base-images-docker/blob/4f2fc8da248a61c3f8e13bbb43e9db6c0ed44ba3/util/run.bzl#L264. In the commit message as well.

ArthurRab:
Recheck cla

tests/contrib/BUILD
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+load("//contrib:rename_image.bzl", "rename_image")
+
+sh_test(
+    name = "rename_image_test",
+    srcs = ["rename_image_test.sh"],
+    data = [":intermediate_bundle_renamed_image.tar"],
```
erain: Is there a way to just use `:renamed_image` target as `data` here? 

If so, I'd suggest to use it; otherwise, let's at least to document it the name very clearly in the rename_image macro. Also, I believe we can come up with a name better than ":intermediate_bundle_%{name}.tar" 😃

ArthurRab: Changed it to be more clear.

tests/contrib/BUILD
```diff
@@ -0,0 +1,28 @@
...
+    name = "rename_image_test",
+    srcs = ["rename_image_test.sh"],
+    data = [":renamed_image.tar"],
+)
+
+rename_image(
+    name = "renamed_image",
+    image = "@distroless_base//image",
+    new_repo = "new_image_repo",
+    new_tag = "new_image_name",
```
xingao267: new_image_tag

tests/contrib/rename_image_test.sh
```diff
@@ -0,0 +1,27 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/bin/bash
+
+set -ex
+BASEDIR=$(dirname "$0")
+
+if [ $(docker load -i "$BASEDIR/renamed_image.tar" |
+    cut -d " " -f3)  = 'new_image_repo:new_image_name' ]
```
xingao267: new_image_name --> new_image_tag

contrib/rename_image.bzl
```diff
@@ -0,0 +1,42 @@
...
+        name: A unique name for the rule, the output tarball is also ${name}.tar
+        image: Label, representing the image to be renamed
+        new_repo: String, new repo name to give to the image
+        new_tag: String, new tag to give to the image
+
+    Produces a tarball '${name}.tar' which contains the same image but now
+    with the name '${new_repo}:${new_tag}'
+    """
+
+    image_name = new_repo + ":" + new_tag
```
xingao267: image_name --> new_image_name

ArthurRab: All fixed, thanks.

tests/contrib/BUILD
ArthurRab: Good call

tests/contrib/rename_image_test.sh
ArthurRab: Fixed

contrib/extract_image_name.sh
```diff
@@ -0,0 +1,35 @@
+# Copyright 2015 The Bazel Authors. All rights reserved.
```
nlopezgi: Is this only being used for test? If so move it to tests/contrib

contrib/extract_image_name.sh
```diff
@@ -0,0 +1,35 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/bin/bash
+
+# Extracts the image name (repo:tag) from the tarball without running it
+# Does not require docker to be installed
+
+mkdir temp_for_extracting_id
+cd temp_for_extracting_id
```
nlopezgi: Why do you need to do this in a subdirectory?

tests/contrib/rename_image_test.sh
```diff
@@ -0,0 +1,26 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/bin/bash
+
+set -ex
+BASEDIR=$(dirname "$0")
+
+if [ $(./contrib/extract_image_name.sh ./tests/contrib/renamed_image.tar) = 'new_image_repo:new_image_tag' ]
```
nlopezgi: all these values should be passed as args in the BUILD rule so that they can be moved / renamed w/o having to change this script

nlopezgi: Also, if this is the only place where you're using extract_image_name.sh, just do the extracting of the name directly here (no need to separate extracting from the evaluating if its the same as the one produced by the rules)

contrib/extract_image_name.sh
ArthurRab: Originally this was for easy cleanup, and to avoid accidentally overriding files. I just made a change that makes cleanup easy regardless. Should I not be concerned about overriding files? If so, I can remove the subdirectory usage.

tests/contrib/rename_image_test.sh
ArthurRab: I thought this might be useful in the future, so I made it a separate script. That's also why I put it in contrib. Should I just do it in the this file and remove the script?

contrib/extract_image_name.sh
ArthurRab: Sounds good, fixed.

tests/contrib/rename_image_test.sh
ArthurRab: Fixed.

tests/contrib/rename_image_test.sh
```diff
@@ -0,0 +1,42 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+#!/bin/bash
+
+function extract_image_name () {
+    # Extracts the image name (repo:tag) from the tarball without running it
+    # Does not require docker to be installed
+
+    tar -xf "$1" "manifest.json"
```
nlopezgi: create a var somewhere with a meaningful name and assign it as value $1, so that you dont have to use $1 enlsewhere (same for $2)

ArthurRab: Good call. Changed.

Issue #379, opened 2018-04-09
How to reset to the default `null` entrypoint?


kevingessner:
When building a `container_image` on top of a base that already defines an entrypoint, how can I reset the new image to have docker's default `null` (i.e. empty) entrypoint?  I tried setting `entrypoint = []` and `entrypoint = None`, which both result in the base's entrypoint being preserved.  Is there a value that will reset to the default, or one that behaves like it?

mattmoor:
Does `entrypoint = None` work?

kevingessner:
No, it doesn't: that seems to be equivalent to not adding the `entrypoint` attribute at all, which inherits the entrypoint from base.

I'm looking to do what would be `ENTRYPOINT []` in a Dockerfile (per https://stackoverflow.com/questions/40122152/how-to-remove-entrypoint-from-parent-image-on-dockerfile).  I believe that `None`, `[]`, etc don't work in `container_image` because they are all falsy in Python and so end up being ignored as "unset" but the image builder; they aren't special-cased like that value is in a Dockerfile.

mattmoor:
> that seems to be equivalent to not adding the entrypoint attribute at all

@kevingessner good point. :)

I think the appropriate fix here would be to make `entrypoint = []` do what you want (despite its "falsy" nature) by explicitly checking for `None`.

WDYT?

kevingessner:
That sounds right to me.

siddharthab:
Can we also do this for `cmd` please?

Most base images have `cmd` defined and the `app_layer` rule sets the `entrypoint`. Docker behavior is to unset the `cmd` of any parent image if `entrypoint` is (re)defined ([source](https://github.com/moby/moby/issues/19611)). But this does not happen in `container_image`, and we end up with unwanted arguments to the entrypoint command when running the image.

At least being able to manually reset `cmd` would be sweet.

xingao267:
Created https://github.com/bazelbuild/rules_docker/pull/471 to set cmd or entrypoint to null if the value is None, [] or ""

xingao267:
https://github.com/bazelbuild/rules_docker/pull/471 is merged. Now setting entrypoint or cmd in container_image to one of {None, [] or ""} will make entrypoint or cmd "null" in the final image config. I'm closing this issue. But feel free to reopen this if you encounter any issue.

PR #426, opened 2018-06-08
Add a `null_entrypoint` attribute to app_layer

siddharthab:
This allows people to not use the default entrypoint which is geared
towards runnig the container as executable with one or more arguments.

With zero arguments also, it is a problem when the base image already
specifies a Cmd in its config. Because that becomes the argument to
the executable and there is no way to override it currently.

See https://github.com/google/containerregistry/pull/86

Resolves https://github.com/bazelbuild/rules_docker/issues/379

### DIFF
--- a/lang/image.bzl
+++ b/lang/image.bzl
@@ -214,6 +214,15 @@ def _app_layer_impl(ctx, runfiles = None, emptyfiles = None):
         _external_dir(ctx): _runfiles_dir(ctx),
     })
 
+    if ctx.attr.null_entrypoint and ctx.attr.entrypoint:
+        fail("You can specify only one of entrypoint and null_entrypoint")
+
+    # Use entrypoint so we can easily add arguments when the resulting
+    # image is `docker run ...`.
+    # Per: https://docs.docker.com/engine/reference/builder/#entrypoint
+    # we should use the "exec" (list) form of entrypoint.
+    entrypoint = (None if ctx.attr.null_entrypoint else ctx.attr.entrypoint + [_binary_name(ctx)] + ctx.attr.args)
+
     return _container.image.implementation(
         ctx,
         # We use all absolute paths.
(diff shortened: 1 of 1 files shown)


erain:
Hi @siddharthab : IIUC `app_layer` is not meant to be directly, rather it is used in those `lang_image` rules which needs to be compatible with the corresponding bazel `lang_binary` rules. 

Also, if you are reusing `app_layer` rules, could you give me a concrete case where you need an extra attr `null_entrypoint` rather than set `entrypoint` to null directly?

siddharthab:
I think all of this is discussed in https://github.com/bazelbuild/rules_docker/issues/379

The lang_image rules allows you to specify a base image. Some of these base images already have a Cmd attribute defined in their layers. This Cmd attribute then becomes an argument to the executable which is undesirable behavior (suggested https://github.com/google/containerregistry/pull/86 for this). See for example, the workaround I had to do for R images -- https://github.com/grailbio/rules_r/blob/928e1300dd9b59794bdb8a8234bc9ed47ff8f90e/R/container/binary.bzl#L104

The other use case is when you want to use the lang_image in an external system where they will provide their own Cmd, where they assume that the entry point is null. Like se do so in https://github.com/grailbio/reflow. Ideally, the entry point should be reset by such systems when executing the image, but sometimes end users can not control these things.

Overall, I think it is a good idea to give some control to the end users.

siddharthab:
One more thing if it's not clear. The lang_image rules are macros that pass on their attributes to app_layer rules. So end users will be able to specify a null entrypoint per image target.

siddharthab:
Just realised https://github.com/bazelbuild/rules_docker/issues/379 talks about a related problem but not this. Sorry about that. My comment there is what this PR tries to resolve.

siddharthab:
On thinking more about this, I realize now that a proper solution will be to have a mechanism [here](https://github.com/siddharthab/containerregistry/blob/master/transform/v2_2/metadata_.py) to clear the default values without setting a new value. Right now, there is no such mechanism. If we can think of a way to clear attributes from base images without specifying new ones, then this PR is not needed.

erain:
Hi @siddharthab sorry to late for this issue. 

My understand in #379 is to solve the problem in `container_image` layer as @mattmoor indicated, to have `entrypoint=[]` behave in a way that it will override the entrypoint of the base container (or any layers in between. Therefore, we can have the same behavior as you implemented without adding a new attr. 

As for `cmd` attr, I guess we can implement the same as `entrypoint`, so users have the flexibility to set whatever combination they want. 

Would you like to help implement the feature listed?

siddharthab:
That is correct. Except I can not think of a way to do this in skylark/python without breaking existing behaviour.

xingao267:
Created https://github.com/bazelbuild/rules_docker/pull/471 to properly set null entrypoint and cmd in container_image rule.

siddharthab:
Closing this in favor of the other PR.

PR #430, opened 2018-06-15
Test to check correctness of failure cases of compare_ids_test

ArthurRab:
There is now a create_failing_test_for_compare_ids_test test rule which
takes in the same attributes as the compare_ids_test test rule and
creates a new compare_ids_test from that in a new workspaces, and passes
only if the new test fails.

Added some uses of this test rule into //contrib/tests to test correctness of compare_ids_test


tests/contrib/compare_ids_test_fail_test.sh.tpl
```diff
@@ -0,0 +1,46 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# The test script generated by the create_failing_test_for_compare_ids_test test rule
+# This creates a new workspace and copies all required files into it, then runs the test
+# which is expected to fail, and passes only if it fails.
+
+#!/bin/bash
```
xingao267: Add docs for what you are doing here.

tests/contrib/compare_ids_test_fail_test.sh.tpl
```diff
@@ -0,0 +1,46 @@
...
+ln ../{tpl_path}
+
+k=0
+for i in {tars}
+do
+  mv $(ln -v ../$i | cut -d "'" -f2) $k.tar
+  k=`expr $k + 1`
+done
+
+if bazel test --test_output=all //:test
```
xingao267: Add feature to check expected output regex from the test. 

Also you can do the if/else in one line. https://github.com/bazelbuild/bazel/blob/master/src/test/shell/bazel/bazel_toolchain_test.sh#L53

tests/contrib/compare_ids_test_fail_test.sh.tpl
```diff
@@ -0,0 +1,46 @@
...
+set -eux
+
+mkdir -p new_temp_workspace_{name}
+cd new_temp_workspace_{name}
+
+touch WORKSPACE
+
+echo > BUILD {test_code}
+
+ln ../{bzl_path}
```
xingao267: Give both operand should be more clear. Also why hard link?

tests/contrib/compare_ids_test_fail_test.sh.tpl
```diff
@@ -0,0 +1,46 @@
...
+
+ln ../{tpl_path}
+
+k=0
+for i in {tars}
+do
+  mv $(ln -v ../$i | cut -d "'" -f2) $k.tar
+  k=`expr $k + 1`
```
xingao267: In contrib/extract_image_id.sh, you use i=$(expr $i + 1) format. Please be consistent. Also, if you can give the indices more meaningful names, that will be better.

tests/contrib/create_failing_test_for_compare_ids_test.bzl
```diff
@@ -0,0 +1,103 @@
...
+load("//:compare_ids_test.bzl", "compare_ids_test")
+
+compare_ids_test(
+    name = "test",
+    id = {id},
+    tars = {tars},
+)
+'
+    """.format(
+        id = repr(ctx.attr.id),
```
xingao267: why repr here? ctx.attr.id seems already a string

tests/contrib/create_failing_test_for_compare_ids_test.bzl
```diff
@@ -0,0 +1,103 @@
...
+        # When linked, the tars will be placed into the base folder, and renamed 0.tar, 1.tar, ... etc
+        tars = repr([str(i) + ".tar" for i in range(len(ctx.attr.tars))]),
+    )
+
+    tar_files = []
+    for tar in ctx.attr.tars:
+        tar_files += list(tar.files)
```
xingao267: Is there a way that we can get the single file from tar?

tests/contrib/compare_ids_test_fail_test.sh.tpl
xingao267: It seems this operation is not permitted. Try to look into why or find an alternative.

tests/contrib/compare_ids_test_fail_test.sh.tpl
ArthurRab: Regex was a good idea. I changed the implementation to no longer use the exit code of the bazel call, but instead look for a regex expression. This way we can make sure the correct failure happens, and a test failing to build does not accidentally pass the fail test.

tests/contrib/compare_ids_test_fail_test.sh.tpl
ArthurRab: Added

tests/contrib/compare_ids_test_fail_test.sh.tpl
ArthurRab: Added comments for first thing as discussed in person. I was originally using hard links because if you create a soft link and the file doesn't exist, no error is thrown. I changed it to symbolic links and testing existence manually. Changing to symbolic links also fixed the operation not being permitted issue.

tests/contrib/create_failing_test_for_compare_ids_test.bzl
ArthurRab: Here repr is effectively adding an extra pair of quotation marks. This is needed since the code that is being formatted needs to have id = "<some_id>" not id = <some_id>.

tests/contrib/create_failing_test_for_compare_ids_test.bzl
ArthurRab: Because the image targets aren't actually files, bazel doesn't know how many files they 'output'. Normally there should be only one but, this method allows using one target that outputs multiple tarballs. If you prefer, I could change it to take just the first tarball in the file list, and make sure the list is of length 1. Thoughts?

tests/contrib/compare_ids_test_fail_test.sh.tpl
```diff
@@ -0,0 +1,70 @@
...
+# call is expected to fail, but should not terminate the script)
+if out="$(bazel test --test_output=all //:test 2>&1)"; then :; fi
+
+for reg_exp in {reg_exps}
+do
+  if [[ {if_modifier} $out =~ $reg_exp ]]
+  then
+    :
+  else
+    echo "'$reg_exp'" did not match >&2
```
xingao267: Does it also print out the output? If not, maybe we can print it to make debug easier for the user.

tests/contrib/create_failing_test_for_compare_ids_test.bzl
```diff
@@ -0,0 +1,120 @@
...
+    for tar in ctx.attr.images:
+        tar_files += list(tar.files)
+
+    tars_string = ""
+    for tar_file in tar_files:
+        tars_string += tar_file.short_path + " "
+
+    runfiles = ctx.runfiles(files = tar_files + [ctx.file._compare_ids_test_bzl, ctx.file._compare_ids_test_sh_tpl, ctx.file._extract_image_id])
+
+    reg_exps = "'" + "' '".join(ctx.attr.reg_exps) + "'"  # Produces string of form (Necessary because of spaces): " 'reg exp 1' 'reg exp 2'"
```
xingao267: Put the comment above it.

tests/contrib/compare_ids_test_fail_test.sh.tpl
```diff
@@ -0,0 +1,70 @@
...
+  tar_num=$(expr $tar_num + 1)
+done
+
+# Save the output from the bazel call (this is in an if because the bazel
+# call is expected to fail, but should not terminate the script)
+if out="$(bazel test --test_output=all //:test 2>&1)"; then :; fi
+
+for reg_exp in {reg_exps}
+do
+  if [[ {if_modifier} $out =~ $reg_exp ]]
```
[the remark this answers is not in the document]
xingao267: You can also simplify this by 

if not match
  echo ....
  exit 1


The code here wants a match with all the input regexes right? Document that in the skylark rule.

tests/contrib/BUILD
```diff
@@ -57,3 +58,53 @@ compare_ids_test(
...
+)
+
+# This test makes sure that passing tests don't pass the failure case test
+create_failing_test_for_compare_ids_test(
+    name = "passing_compare_ids_test_test",
+    images = [
+        ":duplicate1",
+        ":duplicate2",
+    ],
+    negate = True,
```
xingao267: I don't quite understand this one. iiuc, `negate` alters the check for regex. If negate is true, then we see a match, it will fail. We kinda of using negate to foo it here?

Also, make it clear in the rule documentation that:

If negate is false, then all the input regex should match.
If negate is true, none of the input regex should match.
(If I understand your logic correctly)

tests/contrib/compare_ids_test_fail_test.sh.tpl
```diff
@@ -0,0 +1,70 @@
...
+ln -s ../{extractor_path}
+test -f $(basename {extractor_path})
+
+
+
+tar_num=0
+for i in {tars}
+do
+  # Link the supplied tars and rename them to 0.tar, 1.tar, etc.
+  eval mv $(ln -vs ../$i | cut -d " " -f1) ${tar_num}.tar
```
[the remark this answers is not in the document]
ArthurRab: Because $(ln -vs ../$tar | cut -d " " -f1) extracts the name with an extra pair of single quotes around it, so mv crashes because of 'mv: cannot stat "'./image.tar'": No such file or directory'
The eval basically removes the extra single quotes.

tests/contrib/compare_ids_test_fail_test.sh.tpl
```diff
@@ -0,0 +1,71 @@
...
+do
+  # Link the supplied tars and rename them to 0.tar, 1.tar, etc.
+  eval mv $(ln -vs ../$tar | cut -d " " -f1) ${tar_num}.tar
+  test -f ${tar_num}.tar
+  tar_num=$(expr $tar_num + 1)
+done
+
+# Save the output from the bazel call (this is in an if because the bazel
+# call is expected to fail, but should not terminate the script)
+if out="$(bazel test --test_output=all //:test 2>&1)"; then :; fi
```
xingao267: Can you try Can you try out="$(bazel test --test_output=all //:test 2>&1)" || true ?

tests/contrib/create_failing_test_for_compare_ids_test.bzl
```diff
@@ -0,0 +1,119 @@
...
+
+    return [DefaultInfo(runfiles = runfiles)]
+
+"""
+Test to test correctness of failure cases of the compare_ids_test.
+
+Args:
+    images: List of Labels which refer to the docker image tarballs (from docker save)
+    id: (optional) the id we want the images in the tarballs to have
+    reg_exps: a list of regular expressions that must match the output text
```
xingao267: reg_exps: (optional) ...

tests/contrib/BUILD
```diff
@@ -57,3 +58,66 @@ compare_ids_test(
...
+
+create_failing_test_for_compare_ids_test(
+    name = "test_compare_ids_invalid_tar_fails",
+    id = "some_id",
+    images = ["invalid.tar"],
+    reg_exps = [".*Unable to extract manifest\.json, make sure .* is a valid docker image.*"],
+)
+
+create_failing_test_for_compare_ids_test(
+    name = "test_compare_ids_test_some_diff_ids_fails",
```
xingao267: test_compare_ids_test_diff_ids_fails_no_regex

tests/contrib/BUILD
```diff
@@ -57,3 +58,66 @@ compare_ids_test(
...
+)
+
+genrule(
+    name = "make_invalid_tar",
+    outs = ["invalid.tar"],
+    cmd = "touch random_file.txt && tar -c random_file.txt > $@",
+)
+
+create_failing_test_for_compare_ids_test(
+    name = "test_compare_ids_invalid_tar_fails",
```
xingao267: test_compare_ids_test_invalid_tar_fails

tests/contrib/BUILD
```diff
@@ -57,3 +58,66 @@ compare_ids_test(
...
+    name = "test_compare_ids_test_some_diff_ids_fails",
+    images = [
+        "duplicate1",
+        "duplicate2",
+        "@distroless_fixed_id_2//image",
+    ],
+)
+
+create_failing_test_for_compare_ids_test(
+    name = "test_compare_ids_some_invalid_tar_fails",
```
xingao267: test_compare_ids_test_invalid_tar_fails_multi_regex

tests/contrib/BUILD
ArthurRab: Changed

tests/contrib/BUILD
ArthurRab: Changed

tests/contrib/create_failing_test_for_compare_ids_test.bzl
```diff
@@ -41,7 +41,10 @@ compare_ids_test(
     runfiles = ctx.runfiles(files = tar_files + [ctx.file._compare_ids_test_bzl, ctx.file._compare_ids_test_sh_tpl, ctx.file._extract_image_id])
 
     # Produces string of form (Necessary because of spaces): " 'reg exp 1' 'reg exp 2'"
-    reg_exps = "'" + "' '".join(ctx.attr.reg_exps) + "'"
+    if len(ctx.attr.reg_exps) > 0:
```
xingao267: Initialize the variable.

reg_exps = ''
if len(ctx.attr.reg_exps) > 0:
  reg_exps = "'" + "' '".join(ctx.attr.reg_exps) + "'"

xingao267:
buildifier -mode=check $(find . -name '*.bzl' -type f)
./tests/contrib/create_failing_test_for_compare_ids_test.bzl # reformat 
The command "buildifier -mode=check $(find . -name '*.bzl' -type f)" exited with 4.

tests/contrib/create_failing_test_for_compare_ids_test.bzl
```diff
@@ -0,0 +1,120 @@
...
+    images = ["image.tar", "image_with_diff_id.tar"],
+)
+
+create_failing_test_for_compare_ids_test(
+    name = "test2",
+    images = ["image.tar"],
+    id = "<my_wrong_image_sha256>",
+)
+"""
+create_failing_test_for_compare_ids_test = rule(
```
nlopezgi: rename to "compare_ids_fail_test"

tests/contrib/create_failing_test_for_compare_ids_test.bzl
```diff
@@ -0,0 +1,120 @@
...
+    implementation = _create_failing_test_for_compare_ids_test_impl,
```
nlopezgi: rename to _impl

tests/contrib/create_failing_test_for_compare_ids_test.bzl
```diff
@@ -0,0 +1,120 @@
...
+    test = True,
+    attrs = {
+        "images": attr.label_list(mandatory = True, allow_files = True),
+        "id": attr.string(mandatory = False, default = ""),
+        "reg_exps": attr.string_list(mandatory = False, default = []),
+        "_executable_template": attr.label(
+            allow_files = True,
+            single_file = True,
+            default = "compare_ids_test_fail_test.sh.tpl",
```
nlopezgi: rename to compare_ids_fail_test.sh.tpl

tests/contrib/create_failing_test_for_compare_ids_test.bzl
```diff
@@ -0,0 +1,120 @@
...
+
+    tar_files = []
+    for tar in ctx.attr.images:
+        tar_files += list(tar.files)
+
+    tars_string = ""
+    for tar_file in tar_files:
+        tars_string += tar_file.short_path + " "
+
+    runfiles = ctx.runfiles(files = tar_files + [ctx.file._compare_ids_test_bzl, ctx.file._compare_ids_test_sh_tpl, ctx.file._extract_image_id])
```
nlopezgi: wrap line

ArthurRab:
Ok fixed those comments and submitted a pr at [https://github.com/bazelbuild/rules_docker/pull/446](https://github.com/bazelbuild/rules_docker/pull/446)

PR #431, opened 2018-06-18
Made function extract_image_id into its own script

ArthurRab:
Also changed the script that had that function to now use the new script instead

(This was done so that the script can be used to replace the more flaky one used in the base-image-docker repo)

### DIFF
--- a/contrib/compare_ids_test.bzl
+++ b/contrib/compare_ids_test.bzl
@@ -28,12 +28,16 @@ def _compare_ids_test_impl(ctx):
     for tar in tar_files:
         tars_string += tar.short_path + " "
 
-    runfiles = ctx.runfiles(files = tar_files)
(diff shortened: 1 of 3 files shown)


contrib/compare_ids_test.bzl
```diff
@@ -76,5 +80,10 @@ compare_ids_test = rule(
             single_file = True,
             default = "compare_ids_test.sh.tpl",
         ),
+        "_id_extract_script": attr.label(
+            allow_files = True,
+            single_file = True,
+            default = "extract_image_id.sh",
```
xingao267: Can you try to reference this file from base-image-docker repo. you might need to give the whole repo and package path here: like @ io_bazel_rules_docker//contrib:extract_image_id.sh

xingao267: Discussed offline. No need to change.

PR #432, opened 2018-06-20
 Fixed bug involving passing in invalid image tarball 

ArthurRab:
Passing an invalid tarball into compare_ids_test would give a false
passing test. Added fix for this as well as a message to stderr which clearly
states that there was an issue extracting a manifest.json, and to make
sure the tarball is a valid image tarball.

Also changed compare_ids_test attr 'tars' to 'images' for clarity
(This will break the usage of the rule in our private repo and base-images-docker. I will fix this.)

### DIFF
--- a/contrib/extract_image_id.sh
+++ b/contrib/extract_image_id.sh
@@ -20,7 +20,11 @@ test -e $tar_path
 
 # Extracts the manifest.json file from the image's tarball
 # and finds the image id in it
-tar -xf $tar_path "manifest.json"
+if tar -xf $tar_path "manifest.json"; then :
+else
+  echo Unable to extract manifest.json, make sure $tar_path is a valid docker image. >&2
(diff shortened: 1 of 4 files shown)


ArthurRab:
If possible, could we merge this through rebase? These two commits have fairly distinct changes.
Not a huge issue either way though.

contrib/extract_image_id.sh
```diff
@@ -20,7 +20,13 @@ test -e $tar_path
 
 # Extracts the manifest.json file from the image's tarball
 # and finds the image id in it
-tar -xf $tar_path "manifest.json"
+if tar -xf $tar_path "manifest.json"
```
xingao267: can you do the if something then: in one line?

ArthurRab: Changed

PR #435, opened 2018-06-24
Return providers from all container rules

kerinin:
* Use DefaultInfo for `files` and `runfiles`
* Provide rule-specific providers for each rule
* Retain the legacy return struct for backwards compatability

Label attributes can validate values by provider. Returning typed
providers allows rules to be explicit about which container rules they
accept.

I'm working on bazel rules to wrap Helm functionality and I'd like to be able to declare the images used by a helm chart. Restricting the allowed labels allows me to make assumptions about the label's executable (for instance, how to configure a `container_push` executable's `PYTHON_RUNFILES` env var when calling it from another executable).

### DIFF
--- a/container/layer.bzl
+++ b/container/layer.bzl
@@ -40,6 +40,7 @@ load(
     _canonicalize_path = "canonicalize",
     _join_path = "join",
 )
+load("//container:providers.bzl", "LayerInfo")
 
 def _magic_path(ctx, f, output_layer):
     # Right now the logic this uses is a bit crazy/buggy, so to support
@@ -109,16 +110,6 @@ def zip_layer(ctx, layer):
     zipped_layer = _gzip(ctx, layer)
     return zipped_layer, _sha256(ctx, zipped_layer)
 
-# A provider containing information needed in container_image and other rules.
-
-LayerInfo = provider(fields = [
-    "zipped_layer",
-    "blob_sum",
-    "unzipped_layer",
-    "diff_id",
-    "env",
-])
-
 def _impl(
         ctx,
         name = None,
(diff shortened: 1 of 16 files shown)


container/layer.bzl
```diff
@@ -150,7 +141,7 @@ def _impl(
   """
     name = name or ctx.label.name
     file_map = file_map or {}
-    files = files or ctx.files.files
+    files = files or ctx.files.files # providers?
```
nlopezgi: ?

container/providers.bzl
```diff
@@ -0,0 +1,45 @@
...
+# A provider containing information needed in container_image and other rules.
+LayerInfo = provider(fields = [
+    "zipped_layer",
+    "blob_sum",
+    "unzipped_layer",
+    "diff_id",
+    "env",
+])
+
+# A provier identifying container_push rules
```
nlopezgi: A provider containing information exposed by container_push rules

container/providers.bzl
```diff
@@ -0,0 +1,45 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Provider definitions"""
+
+# A provider containing information about a container bundle
+BundleInfo = provider(fields = ["container_images", "stamp"])
+
+# A provider identifying container_flatten rules
```
nlopezgi: Suggestion:
"A provider containing information exposed by container_flatten rules"
or
"A provider containing information about a ..."
Same for other comments - try to pick one of the two options above, or another you think is best, and use it uniformly for all comments.

container/providers.bzl
```diff
@@ -0,0 +1,45 @@
...
+FlattenInfo = provider()
+
+# A provider containing information needed in container_push and other rules
```
nlopezgi: A provider containing information exposed by image_config rules (same for other comments below?)

PR #437, opened 2018-06-27
Set use_default_shell_env to True in zip.bzl

dsjiang753:
Allow client side to override GZIP variable to control the compression speed/ratio.


dsjiang753:
I signed it!

Issue #443, opened 2018-07-04
rules_docker failing in Bazel CI


lfpino:
Hi there,

rules_docker is broken in Bazel CI because of rules_scala (https://github.com/bazelbuild/bazel/issues/5483). Example: https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/330#cf73d6ae-6ba4-4cc6-b54a-64890ec0c115

To fix this the rules_scala version needs to be updated.

PR #444, opened 2018-07-04
Update rules_scala

cushon:
Fixes bazelbuild/rules_docker#443


erain:
It looks like bzl formatter is complaining, and also `//container:image_test` test failed.

cushon:
Thanks, I reformatted and fixed `//container:image_test`.

PR #448, opened 2018-07-05
Recode extract_image_id.sh in python (as well as the compare_ids_test rule)

ArthurRab:
Changed extract_image_id.sh to extract_image_id.py, 
all uses of it will need to be converted over.


contrib/compare_ids_test.bzl
```diff
@@ -24,19 +24,15 @@ def _compare_ids_test_impl(ctx):
...
     runfiles = ctx.runfiles(files = tar_files + [ctx.file._id_extract_script])
 
     ctx.actions.expand_template(
         template = ctx.file._executable_template,
         output = ctx.outputs.executable,
         substitutions = {
             "{id}": ctx.attr.id,
-            "{tars}": tars_string,
-            "{id_script_path}": ctx.file._id_extract_script.short_path,
+            "{tars}": repr([i.short_path for i in tar_files]),
```
nlopezgi: These can be passed as args to a pythons script, no need to use a template.
You can use ctx.actions.run (e.g., https://github.com/bazelbuild/rules_docker/blob/master/contrib/passwd.bzl#L80) using as executable the py_binary target (note how the example uses the //container:build_tar target as the executable)

contrib/BUILD
```diff
@@ -19,8 +19,8 @@ exports_files(["push-all.sh.tpl"])
 
 exports_files(["structure-test.sh.tpl"])
 
-exports_files(["compare_ids_test.sh.tpl"])
+exports_files(["compare_ids_test.py.tpl"])
```
nlopezgi: - This should be a python script, not a template
- this build file should not have an exports_files target, it should have a py_binary target(s) (e.g., https://github.com/bazelbuild/rules_docker/blob/master/container/BUILD#L67)

contrib/compare_ids_test.bzl
```diff
@@ -83,12 +79,12 @@ compare_ids_test = rule(
         "_executable_template": attr.label(
             allow_files = True,
             single_file = True,
-            default = "compare_ids_test.sh.tpl",
+            default = "compare_ids_test.py.tpl",
```
nlopezgi: see https://github.com/bazelbuild/rules_docker/blob/master/contrib/passwd.bzl#L137

contrib/compare_ids_test.bzl
```diff
@@ -24,20 +24,22 @@ def _compare_ids_test_impl(ctx):
...
+        id_args = ["--id", ctx.attr.id]
 
-    ctx.actions.expand_template(
-        template = ctx.file._executable_template,
+    args = " ".join([f.short_path for f in tar_files] + id_args)
+
+    # The compare_ids_test target (produced by the py_binary rule) produces 2 files,
+    # the executable and the source code. This retrieves the executable. A better method
+    # of retrieving it would be ideal, but I could not find one.
+    compare_ids_test_exectuable = ctx.files._compare_ids_test_script[1]
```
nlopezgi: use ctx.executables https://docs.bazel.build/versions/master/skylark/lib/ctx.html#executable

tests/contrib/compare_ids_fail_test.bzl
```diff
@@ -16,6 +16,17 @@
 
 def _impl(ctx):
     test_code = """ '
+py_binary(
```
nlopezgi: why are you need this duplicated from the BUILD file above here?
If its because your test is rewriting the file contrib/BUILD, then do not re write it, write a new BUILD file in some subdirectory (e.g., contrib/fail_test/BUILD)

contrib/BUILD
```diff
@@ -19,8 +19,19 @@ exports_files(["push-all.sh.tpl"])
 
 exports_files(["structure-test.sh.tpl"])
 
-exports_files(["compare_ids_test.sh.tpl"])
+exports_files(["compare_ids_test.py"])
 
 exports_files(["compare_ids_test.bzl"])
 
-exports_files(["extract_image_id.sh"])
+exports_files(["extract_image_id.py"])
```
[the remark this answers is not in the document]
ArthurRab: I use this file in the compare_ids_fail_test.bzl as it needs to copy it into the new workspace it produces to test the failing cases of the compare_ids_test rule.

contrib/compare_ids_test.bzl
ArthurRab: Thanks!

tests/contrib/compare_ids_fail_test.bzl
ArthurRab: Changed it to copy the BUILD instead of duplicate code

contrib/compare_ids_test.py
```diff
@@ -11,21 +13,31 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import sys
+import os
```
xingao267: os and sys are not used. Can you run lint on all the py files

ArthurRab: Yup. Did pyformat, forgot lint. Fixed.

contrib/extract_image_id.py
```diff
@@ -39,10 +55,11 @@ def get_id(tar_path):
...
-  id = config_file.split(".")[0]
+  id_ = config_file.split(".")[0]
 
-  return id
+  return id_
 
 
 if __name__ == "__main__":
-  tar_path = sys.argv[1]
+  path = sys.argv[1]
```
xingao267: no need to create variable `path`

Issue #451, opened 2018-07-09
Working directory is different between py_binary and py_image


marczych:
I have a `py_binary` with some `data` files that are referenced with a relative path in code e.g. `//example:binary` references `example/binary/foo`.

I was surprised to find out that a comparable `py_image` failed to find the file. It turned out that `example/binary/foo` was indeed in the image (at `/app/example/binary.runfiles/example/binary/foo`) but the working directory was `/app` when doing `bazel run //example:binary_image`.

Shouldn't something set the working directory to `/app/example/binary.runfiles` before handing control over to the real binary?

nlopezgi:
Hi, yes I think you are correct and the runfiles directory is not being set properly (this is likely happening not only for py_image, but also for all lang _images). There have been a couple of other issues (https://github.com/bazelbuild/rules_docker/issues/355) and notably https://github.com/bazelbuild/rules_docker/pull/365. I think all issues could be resolved with setting the runfiles directory properly in image.bzl. It would be awesome if you could test the fix I suggest in my comment on https://github.com/bazelbuild/rules_docker/pull/365 on June 18, to see if that fixes your issue and send a PR with the fix.

marczych:
Thank you for the response!

I'm seeing the code using `workdir` but I can't quite make out your proposal to fix it. I'm willing to test it out a patch if you have one!

nlopezgi:
please try out now that fix is merged and let me know if this is still problematic.

marczych:
Works like a charm! Thank you!

Issue #378, opened 2018-04-09
data for a go_image doesn't get into the image


raliste:
For instance

```
go_image(
    name = "gru_docker",
    srcs = [
        "main.go",
    ],
    goarch = "amd64",
    goos = "linux",
    pure = "on",
    visibility = ["//visibility:public"],
    deps = [
        "@com_github_golang_glog//:go_default_library",
        "//gru:gru",
    ],
    data = [
        "//static:static"
    ]
)
```

The same but using the `go_binary` rules does get `static` into the runfiles, but changing to `go_image` doesn't get them into the runfiles.

PranavSathy:
Has anyone found a workaround for this, I will be relying on this functionality to successfully deploy my application.

raliste:
@PranavSathy it turns out data does get into the image, under a `.runfiles` directory.  See the data attribute at https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#go_binary. The prefix should match your binary's name, for instance:

```
go_image(
  name = "foo"
  ...
)
```

runfiles should be `foo.runfiles`. It's also helpful to use filepath.Join and getting your binary's base path:

```
base, err := filepath.Abs(filepath.Dir(os.Args[0]))
if err != nil {
	return nil, err
}

filepath.Join(base, runfiles, pathToYourArtifact)
```

PranavSathy:
@raliste Thanks! I actually managed to come to this conclusion just before as well; but this is surely helpful for anyone in the future.

Just FYI; you can also utilize `binary` in your `go_image` rule and then the runfiles path will be prefixed according to your `go_binary` rule rather than your `go_image`, so that running locally vs. running in Docker will be identical and require 0 code change.

nlopezgi:
It's possible that the workdir is not being set correctly for the go_image rules (or generally for all x_image rules). Please see my comment in #365 on June 18th (also see https://github.com/bazelbuild/rules_docker/issues/355). It would be great if you could test the fix I suggested in that PR for your failure use case. If it does also work for you, then we can be sure that the workdir fix is right for all x_image rules.
--

PR #452, opened 2018-07-11
Fix workdir and expand arg location.

### LINKED ISSUE #355: nodejs_image does not preserve data filepaths
In nodejs_binary, you can pass in data that will be accessible to the binary at runtime. 

When replaced with nodejs_image, these files are not available when executed. If you manually cd into the workspace directory under the runfiles and run the .binary file via sh ../../{binary_file}, it will work as expected.

Does this have to do with the symlinks in app_layer?

Globegitter:
This supersedes #365. And should fix #355, #378, #451

### DIFF
--- a/container/layer_tools.bzl
+++ b/container/layer_tools.bzl
@@ -100,8 +100,7 @@ def incremental_load(
         output,
         stamp = False,
         run = False,
-        run_flags = None,
-        args_len = 0):
+        run_flags = None):
     """Generate the incremental load statement."""
     stamp_files = []
     if stamp:
@@ -162,7 +161,7 @@ def incremental_load(
             # bazel automatically passes ctx.attr.args to the binary on run, so need to truncate
             # them here to not pass twice, as they are already part of the docker entrypoint
             run_statements += [
-                "docker run %s %s \"${@:%s}\"" % (run_flags, tag_reference, str(args_len + 1)),
+                "docker run %s %s \"$@\"" % (run_flags, tag_reference),
             ]
 
     ctx.template_action(
--- a/container/layer_tools.bzl
+++ b/container/layer_tools.bzl
@@ -158,8 +158,8 @@ def incremental_load(
             ),
         ]
         if run:
-            # bazel automatically passes ctx.attr.args to the binary on run, so need to truncate
-            # them here to not pass twice, as they are already part of the docker entrypoint
+            # bazel automatically passes ctx.attr.args to the binary on run, so args get passed in
+            # twice. See https://github.com/bazelbuild/rules_docker/issues/374
             run_statements += [
                 "docker run %s %s \"$@\"" % (run_flags, tag_reference),
             ]
--- a/container/layer_tools.bzl
+++ b/container/layer_tools.bzl
@@ -100,7 +100,8 @@ def incremental_load(
         output,
         stamp = False,
         run = False,
-        run_flags = None):
+        run_flags = None,
+        args_len = 0):
     """Generate the incremental load statement."""
     stamp_files = []
     if stamp:
@@ -158,8 +159,10 @@ def incremental_load(
             ),
         ]
         if run:
+            # bazel automatically passes ctx.attr.args to the binary on run, so need to truncate
+            # them here to not pass twice, as they are already part of the docker entrypoint
             run_statements += [
-                "docker run %s %s \"$@\"" % (run_flags, tag_reference),
+                "docker run %s %s \"${@:%s}\"" % (run_flags, tag_reference, str(args_len + 1)),
             ]
 
     ctx.template_action(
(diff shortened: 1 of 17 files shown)


nlopezgi:
Thanks a lot for sending this PR! Looks good overall, will review in more detail in a bit, kicked off presubmits to make sure everything is passing.

python3/image.bzl
```diff
@@ -74,6 +74,8 @@ def py3_image(name, base = None, deps = [], layers = [], **kwargs):
     # TODO(mattmoor): Consider using par_binary instead, so that
     # a single target can be used for all three.
 
+    args = kwargs.pop("args", None)
```
nlopezgi: why is this necessary for py3 image but not for others?

testdata/py3_image.py
```diff
@@ -19,6 +21,9 @@ def main():
   print('Second: %d' % py3_image_library.fn(2))
   print('Third: %d' % py3_image_library.fn(3))
   print('Fourth: %d' % py3_image_library.fn(4))
+  print(sys.argv)
```
nlopezgi: can you add short doc line somewhere indicating these two tests (py3, py) expect the third argument to be a file location

lang/image.bzl
```diff
@@ -251,9 +255,10 @@ app_layer = rule(
 
         # Override the defaults.
         "data_path": attr.string(default = "."),
-        "workdir": attr.string(default = "/app"),
+        "workdir": attr.string(default = ""),
         "directory": attr.string(default = "/app"),
         "legacy_run_behavior": attr.bool(default = False),
+        "data": attr.label_list(cfg = "data", allow_files = True),
```
nlopezgi: iiuc, this should have as value host/target? https://docs.bazel.build/versions/master/skylark/lib/attr.html#label_list

lang/image.bzl
```diff
@@ -211,21 +212,24 @@ def _app_layer_impl(ctx, runfiles = None, emptyfiles = None):
         _binary_name(ctx): _final_file_path(ctx, ctx.executable.binary),
         # Create a directory symlink from <workspace>/external to the runfiles
         # root, since they may be accessed via either path.
-        _external_dir(ctx): _runfiles_dir(ctx),
+        _external_dir(ctx): runfiles_dir(ctx),
     })
 
+    args = [ctx.expand_location(arg, ctx.attr.data) for arg in ctx.attr.args]
```
nlopezgi: add doc comment above explaining you are expanding locations of the form $(location :some_target)

java/image.bzl
```diff
@@ -215,7 +219,9 @@ jar_app_layer = rule(
         "directory": attr.string(default = "/app"),
         # https://github.com/bazelbuild/bazel/issues/2176
         "data_path": attr.string(default = "."),
+        "workdir": attr.string(default = ""),
         "legacy_run_behavior": attr.bool(default = False),
+        "data": attr.label_list(cfg = "data", allow_files = True),
```
nlopezgi: should be host/target https://docs.bazel.build/versions/master/skylark/lib/attr.html#label_list

java/image.bzl
```diff
@@ -166,12 +168,13 @@ def _jar_app_layer_impl(ctx):
 
     binary_path = layer_file_path(ctx, ctx.files.binary[0])
     classpath_path = layer_file_path(ctx, classpath_file)
+    args = [ctx.expand_location(arg, ctx.attr.data) for arg in ctx.attr.args]
```
nlopezgi: add doc comment about expanding location

container/layer_tools.bzl
```diff
@@ -158,6 +158,8 @@ def incremental_load(
             ),
         ]
         if run:
+            # bazel automatically passes ctx.attr.args to the binary on run, so need to truncate
+            # them here to not pass twice, as they are already part of the docker entrypoint
```
nlopezgi: Can you add a pointer to the Bazel issue you opened so we can track and remove once that is fixed?

python3/image.bzl
Globegitter: Sorry this is still a leftover, I also tried to fix the issue that args are being passed in twice.

container/layer_tools.bzl
Globegitter: This is also still a leftover from when I tried to also fix that issue. Yes will clean up.

lang/image.bzl
Globegitter: I have seen some doc referring that for data it should still be data, there is also a ticket for it: https://github.com/bazelbuild/bazel/issues/4316. But here: https://github.com/bazelbuild/examples/blob/5d1e5bd82e061e8e2e04bb54883fc4c79c935f76/rules/runfiles/execute.bzl#L32 there is nothing, so I will just remove it.

PR #462, opened 2018-07-19
Add ports option to go_image rule

alsutton:
We've been creating a docker image from a go application which exposes a network port for other services to communicate with it. Unfortunately it looks like the current go_image rule does not support exposed port data, so I've added it in.


alsutton:
Ok, that’s a pretty unpleasant looking workaround, so I’ll recommend we stick with our fork that has this PR in it.

Thanks for looking it over.

PR #463, opened 2018-07-19
Fix nodejs_image args.

Globegitter:
Turns out the `args` are not being forwarded to the image. Removing the `sh -c` entrypoint and passing the args via `cmd` fixes that. Further I think nodejs is fine with symlinks now as the `--preserver-symlinks` flag is being passed in by the binary rule.


Globegitter:
Yep will do as soon as I can. I already wanted to, but as it sometimes so happens other work has come inbetween.

nlopezgi:
no hurries, and thanks for your contributions!

Globegitter:
@nlopezgi Tests are now passing.

PR #465, opened 2018-07-20
Use Skylark versions of http_file and http_archive.

### LINKED ISSUE #417: Please move from native to Skylark rules for external git/http repositories
Bazel is planning to move away from native rules for external repositories to Skylark rules, mainly because bundling every version control system in bazel does not scale. As part of this effort, we plan to submit https://bazel-review.googlesource.com/c/bazel/+/55932 soon, which will cause bazel by default to insist that the skylark versions the git and http rules for external repositories be used. This can be disabled by a flag for at least the next 6 months, but we still would appreciate if you could adopt your repository to use the Skylark versions of those rules soon.

A test run of downstream projects with that flag flip in bazel enabled can be found at https://buildkite.com/bazel/bazel-with-downstream-projects-bazel/builds/286

For more background, see
- the original announcement https://groups.google.com/d/topic/bazel-discuss/dO2MHQLwJF0/discussion
- and the documentation of the incompatible changes https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#remove-native-git-repository

Thanks.

xingao267:
Change all git_repository to http_archive as the latter supports better
sha check.

Fixes #417.


xingao267:
Fixed.

xingao267:
not yet lol

PR #471, opened 2018-07-27
Properly set cmd and entrypoint to null when needed.

xingao267:
If cmd or entrypoint is set to one of {None, [], ""}, it will be set to null in the image config.

#379


siddharthab:
Thank You!

container/image.bzl
```diff
@@ -380,6 +394,13 @@ _attrs = dict(_layer.attrs.items() + {
         executable = True,
         allow_files = True,
     ),
+    # null_cmd and null_entrypoint are hidden attributes from users.
```
nlopezgi: You need to document somewhere much more visible that explicitly setting cmd or entrypoint to None, [] or "" will have this behavior (here: https://github.com/bazelbuild/rules_docker#container_image-1 , https://github.com/bazelbuild/rules_docker/blob/master/container/image.bzl#L476,  https://github.com/bazelbuild/rules_docker/blob/master/container/image.bzl#L482)

tests/docker/BUILD
```diff
@@ -39,3 +39,53 @@ container_test(
...
+)
+
+container_image(
+    name = "image_with_null_cmd_entrypoint",
+    base = ":image_with_some_cmd_entrypoint",
+    cmd = [],
+    entrypoint = None,
+)
+
+container_test(
```
nlopezgi: What happens if you have another container_image that uses image_with_null_cmd_entrypoint as base but sets a different (not null) entrypoint (i.e., can you verify the topmost layer overrides everything below)?

set_cmd_and_entrypoint_null_on_base

tests/docker/BUILD
```diff
@@ -39,3 +39,53 @@ container_test(
     configs = ["//tests/docker/configs:deb_image_with_dpkgs.yaml"],
     image = ":deb_image_with_dpkgs",
 )
+
+container_image(
```
nlopezgi: lets improve naming
remove image_with_ from all targets here here

tests/docker/BUILD
```diff
@@ -39,3 +39,53 @@ container_test(
...
+    name = "image_with_some_cmd_entrypoint",
```
nlopezgi: set_cmd_and_entrypoint

also add tests with just cmd and just entrypoint

tests/docker/BUILD
```diff
@@ -39,3 +39,53 @@ container_test(
...
+    base = "@distroless_base//image",
+    cmd = ["foo"],
+    entrypoint = ["bar"],
+)
+
+container_test(
+    name = "image_with_some_cmd_entrypoint_test",
```
nlopezgi: set_cmd_and_entrypoint_test

tests/docker/BUILD
```diff
@@ -39,3 +39,53 @@ container_test(
...
+    configs = ["//tests/docker/configs:image_with_some_cmd_entrypoint.yaml"],
+    image = ":image_with_some_cmd_entrypoint",
+)
+
+container_image(
+    name = "image_should_still_have_cmd_entrypoint",
```
nlopezgi: base_with_cmd_and_entry

tests/docker/BUILD
```diff
@@ -39,3 +39,53 @@ container_test(
...
+)
+
+container_test(
+    name = "image_should_still_have_cmd_entrypoint_test",
+    configs = ["//tests/docker/configs:image_with_some_cmd_entrypoint.yaml"],
+    image = ":image_should_still_have_cmd_entrypoint",
+)
+
+container_image(
+    name = "image_with_null_cmd_entrypoint",
```
nlopezgi: null_entrypoint_none

also add:
null_entrypoint_empty_str
null_entrypoint_empty_list

tests/docker/BUILD
xingao267: Done.

tests/docker/BUILD
xingao267: Done.

tests/docker/BUILD
xingao267: Done.

tests/docker/BUILD
xingao267: Done.

tests/docker/BUILD
xingao267: Done.

tests/docker/BUILD
xingao267: Done.

container/image.bzl
xingao267: Done.

PR #433, opened 2018-06-22
Support control.tar with xz or gzip compression

clintharrison:
Fixes #386.

When run with Python 3, this will prefer the built-in `lzma` module to extract xz-compressed tarfiles in the `.deb`. Similar to the behavior of `tools.build_defs.pkg.archive` for compatibility with Python 2, this will use `xzcat` (technically with `xz` and specific flags so it behaves as `xzcat`).

I've tested this with `.deb` files with `control.tar.gz` and `control.tar.xz` from stretch and buster, under Python 2.7 and 3.6.

PR #473, opened 2018-07-30
Use lzma to decompress .tar.xz debian files.

hrfuller:
Recent debian packages from sid can come XZ compressed. This lets you use lzma to build these .debs if it is installed and gives a helpful error message if it isn't.

### DIFF
--- a/container/build_tar.py
+++ b/container/build_tar.py
@@ -23,7 +23,6 @@ import sys
 import re
 import tarfile
 import tempfile
-import warnings
 
 from tools.build_defs.pkg import archive
 from third_party.py import gflags
--- a/container/build_tar.py
+++ b/container/build_tar.py
@@ -28,18 +28,6 @@ import warnings
 from tools.build_defs.pkg import archive
 from third_party.py import gflags
 
-lzma = None
-try:
-  import lzma
-except ImportError:
-  try:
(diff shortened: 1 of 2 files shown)


container/build_tar.py
```diff
@@ -259,7 +272,16 @@ def add_deb(self, deb):
       while current:
         parts = current.filename.split(".")
         name = parts[0]
-        ext = '.'.join(parts[1:])
+        if parts[-1] == 'xz':
```
nlopezgi: Could you try to do the import of lzma only if needed (i.e., inside this conditional, as in this PR: https://github.com/bazelbuild/rules_docker/pull/433/files)

nlopezgi:
can you also please sign the CLA

container/build_tar.py
```diff
@@ -271,7 +271,21 @@ def add_deb(self, deb):
       while current:
         parts = current.filename.split(".")
         name = parts[0]
-        ext = '.'.join(parts[1:])
+        if parts[-1].lower() == 'xz':
+          try:
```
nlopezgi: any reason why you can't just reuse _xz_decompress here?

hrfuller: No, that makes sense. Ill change it to support backports for python 2.7.

Issue #450, opened 2018-07-09
container_push fails with custom base image


marczych:
I had a working `container_push` target for a `py_image` but realized that I needed a custom base image because the code was shelling out to `git`. I got that all built, wired up, and working when running the image directly. However, my script to publish the image started failing with this error:

```
INFO: Running command line: bazel-bin/<redacted>
E0709 10:59:25.470009   19660 docker_session_.py:310] Error during upload of: <redacted>
F0709 10:59:25.470669   19660 fast_pusher_.py:171] Error publishing <redacted>: response: {'status': '400', 'content-length': '204', 'x-content-type-options': 'nosniff', 'server': 'nginx/1.9.15', 'connection': 'keep-alive', 'docker-distribution-api-version': 'registry/2.0', 'date': 'Mon, 09 Jul 2018 17:59:36 GMT', 'content-type': 'application/json; charset=utf-8'}
provided digest did not match uploaded content: {u'Reason': {}, u'Digest': u'sha256:717b092b8c86356cf976d9c83fa6f0ea67f2bf3148a5bbb7e02026a5d3245e68'}
ERROR: Non-zero return code '1' from command: Process exited with status 1
```

As far as I can tell, the only thing that changed from the perspective of the `container_push` target was the `base` argument to the `py_image` that it depends on.

I started poking around the push command and the runfiles to find that the computed digests indeed don't match:

```
[marc.zych@marc-desktop] - [~/.cache/bazel/_bazel_marc.zych/<redacted>/image] - [Mon Jul 09, 11:31:34]
[I]> sha256sum *.tar.gz
dfe544c43f03c0dbc7cc1dd0e0f0ff878010e5ca99dae949c246a8a5088c3b3d  000.tar.gz
296f72621ec779f9bec360cf247ef3baa5cdc3a374af19117b8d11cf6b5eebec  001.tar.gz
0de32c930a012885e189a14dc3fc6aed1853b23fca1d75810234f73ca2690b28  002.tar.gz
9a8c4414e2a0aa20e59beb6bd9cb92b1c56a25c52c5469b312741ee9ef11e0a9  003.tar.gz
481c50b91eb2bedac0866fced8f8588f8626d3e68d40db5a08c995edc7753472  004.tar.gz
f11797c11bb3af5d869e2a4408fa7d6fc6541c439c2d1cdba2dafde49ae4746b  005.tar.gz
[...]

Issue #298, opened 2018-01-29
Container Push with Container Load image fails.


afking:
Given my WORKSPACE, where image.tar is the `docker save` from the busybox image:

```python
container_pull(
    name = "busybox",
    registry = "index.docker.io",
    repository = "library/busybox",
    tag = "latest",
) 

container_load(
    name = "busybox_test",
    file = "//:image.tar",
)
```

And then a BUILD file:

```python
container_bundle(
    name = "bundle_test",
    images = {
        "ACCOUNT.REGION.amazonaws.com/repo:busybox": "@busybox//image",
        "ACCOUNT.REGION.amazonaws.com/repo:busybox_test": "@busybox_test//image",
    },
)

container_push(
    name = "export_test",
    bundle = ":bundle_test",
    format = "Docker",
)
```

Bazel fails with:

```sh
bazel run :export_test
INFO: Analysed target //docker:export_test (23 packages loaded).
INFO: Found 1 target...
Target //docker:export_test up-to-date:
  bazel-bin/docker/export_test
INFO: Elapsed time: 7.306s, Critical Path: 0.21s
INFO: Build completed successfully, 10 total actions

INFO: Running command line: bazel-bin/docker/export_test
ACCOUNT.REGION.amazonaws.com/repo:busybox was published with digest: sha256:f353bdf4866c5e5b5c76a97e61312cab1a788b58ce3ab7a2f31a87844a0f4093
E0129 17:08:25.815792   10712 docker_session_.py:335] Error during upload of: ACCOUNT.REGION.amazonaws.com/repo:busybox_test
Traceback (most recent call last):
  File "/private/var/tmp/_bazel_edwardmcfarlane/ad32b996cb424fa4cc44a5c86220cb2f/execroot/com_jazznetworks_repo_jazz_main/bazel-out/darwin-fastbuild/bin/docker/export_test.runfiles/com_jazznetworks_repo_jazz_main/../containerregistry/tools/fast_pusher_.py", line 149, in <module>
    main()
  File "/private/var/tmp/_bazel_edwardmcfarlane/ad32b996cb424fa4cc44a5c86220cb2f/execroot/com_jazznetworks_repo_jazz_main/bazel-out/darwin-fastbuild/bin/docker/export_test.runfiles/com_jazznetworks_repo_jazz_main/../containerregistry/tools/fast_pusher_.py", line 140, in main
    session.upload(v2_2_img)
[...]

PR #478, opened 2018-08-03
Use the zipped layer to calculate blobsums on imports

chrismcbride:
If a docker save tarball was imported with unzipped layers the blobsum
  would incorrectly be calculated off of the unzipped layer and cause
  digest mismatches when pushing.

Resolves Issues #450 and possibly issue #298


chrismcbride:
I signed it!

chrismcbride:
I typoed the second commit message title. It should say "Rehome container_import testdata *and fix* container/image_test.py". Not sure if that can be fixed when this is rebased and merged or if I should force push an amended commit.

nlopezgi:
Thanks for sending this PR! I've kicked off tests and will review code in detail in a bit. We can fix the commit message when this gets merged.

container/BUILD
```diff
@@ -150,6 +148,8 @@ TEST_DATA = [
     for t in TEST_TARGETS
 ] + [
     "//testdata:stamped_bundle_test",
+    "//testdata/import:pause_piecemeal.tar",
```
nlopezgi: why do these needed to be moved from TEST_TARGETS to TEST_DATA (and they just be left as targets)?

testdata/import/gz/BUILD
```diff
@@ -0,0 +1,11 @@
+load("//container:container.bzl", "container_import")
```
nlopezgi: Could you explain your reasoning why you think this is worth splitting off into a separate file?

testdata/import/BUILD
```diff
@@ -0,0 +1,45 @@
+load("//container:container.bzl", "container_import")
```
nlopezgi: Could you explain your reasoning why you think this is worth splitting off into a separate file?

nlopezgi: (also, if this is to be split off, it needs a license declaration)

container/BUILD
chrismcbride: because they have different paths then the rest

testdata/import/BUILD
chrismcbride: If I didnt move anything in testdata, this is the error I get:

```
ERROR: file 'testdata/da7bd81140ca921a067c604a3ba3f54d4e0d51ba5276cd9f32ad6a28e588f470.tar.gz.sha256' is generated by these conflicting actions:
Label: //testdata:pause_piecemeal_gz, //testdata:pause_piecemeal
RuleClass: container_import rule
Configuration: 7d61bd6adacde84bec39dad107f12252
Mnemonic: SHA256
Action key: cc9de884bd3f2c8c6e87d3ef4136c7ed, 205c201b5fd9a47a9f8ac944aff3562b
Progress message: SHA256 testdata/da7bd81140ca921a067c604a3ba3f54d4e0d51ba5276cd9f32ad6a28e588f470.tar.gz.sha256
PrimaryInput: File:[[<execution_root>]bazel-out/host/bin]external/bazel_tools/tools/build_defs/hash/sha256
PrimaryOutput: File:[[<execution_root>]bazel-out/k8-fastbuild/bin]testdata/da7bd81140ca921a067c604a3ba3f54d4e0d51ba5276cd9f32ad6a28e588f470.tar.gz.sha256
Owner information: //testdata:pause_piecemeal_gz BuildConfigurationValue.Key[7d61bd6adacde84bec39dad107f12252] false, //testdata:pause_piecemeal BuildConfigurationValue.Key[7d61bd6adacde84bec39dad107f12252] false
MandatoryInputs: Attempted action contains artifacts not in previous action (first 5): 
	testdata/da7bd81140ca921a067c604a3ba3f54d4e0d51ba5276cd9f32ad6a28e588f470.tar.gz
Previous action contains artifacts not in attempted action (first 5): 
	testdata/da7bd81140ca921a067c604a3ba3f54d4e0d51ba5276cd9f32ad6a28e588f470.tar.gz
Outputs: are equal
ERROR: Analysis of target '//testdata:pause_piecemeal_gz' failed; build aborted
```

testdata/import/gz/BUILD
nlopezgi: thanks for explaining, sgtm to keep them in separate folders. Could you create a single folder under testdata/ for each of these files and give them a name that is explicitly related to what is being declared (e.g., this second file could be testdata/pause_piecemeal_gz/BUILD, or something like that, as that the only thing it declares)

testdata/import/BUILD
nlopezgi: sgtm, see below. Also, please add license declaration to both new BUILD files

container/BUILD
nlopezgi: I'm not sure they need to be here just because they have different paths. Did you try to leave them where they were with the new path? (if you did and got an error, could you let me know what kind of error it was, it does not make sense to have to have these be test data dependencies as opposed to test targets.

testdata/import/BUILD
nlopezgi: Thanks for explaining, and sorry I missed the doc comment in the file below. sgtm to have them in separate folders. 
Could you please:
- create one folder for each of these files under testdata and give them a name that is more specific to what they declare (e.g., this one could be testdata/pause_piecemeal/BUILD the one below could be testdata/pause_piecemeal_gz/BUILD, or something like that)
- add license declarations doc comments to both files (see other BUILD files for reference)
- add doc comment in both BUILD files explaining why they need to be separated (right now doc is only in the file below)

container/BUILD
nlopezgi: I'm not sure they need to be here just because they have different paths. This struct is declaring data deps of the tests, the one above is declaring targets. If you add the dep here, any changes to this test would cause rerunning all other tests (as they depend on this one) if you add the dep above, changes to this test would only cause rerunning this test.

testdata/import/gz/BUILD
nlopezgi: thanks, see comment in testdata/import/BUILD

container/BUILD
```diff
@@ -76,77 +76,77 @@ py_binary(
...
+    ":py_image",
+    ":rust_image",
+    ":scala_image",
+    ":war_image",
+    ":flat",
 ]
 
 TEST_DATA = [
-    "//testdata:%s.tar" % t
+    "//testdata%s.tar" % t
```
chrismcbride: I had to change this to not add the ":" since pause_piecemeal is in a subpackage

nlopezgi: lgtm

testdata/pause_piecemeal/BUILD
```diff
@@ -0,0 +1,45 @@
...
+[genrule(
+    name = "extract_%s" % id,
+    srcs = ["//testdata:pause.tar"],
+    outs = ["%s.tar" % id],
+    cmd = ("tar xOf $(location //testdata:pause.tar) %s/layer.tar " +
+           "> $(location :%s.tar)") % (id, id),
+) for id in PAUSE_LAYERS]
+
+# this needs to be in a seperate package than the rest of the testdata because
+# it generates identically named artifacts as pause_piecemeal.
```
nlopezgi: pause_piecemeal_gz

siddharthab:
Thank you and sorry for the bug. I actually never ended up using my changes, and continued to provide compressed layers through my original implementation in https://github.com/bazelbuild/rules_docker/pull/191/commits/593e429cd388f5c5f13f80a70739722d565ae997.

It looks like the optimization I noticed in https://github.com/bazelbuild/rules_docker/pull/195#issuecomment-336547322 is no longer valid because we will need the compressed layer digest.

PR #481, opened 2018-08-04
Use pusher based on go-containerregistry

siddharthab:
This pull request is merely to share work that I had to do for avoiding python2 code paths in containerregistry.

This will still need more work, namely
1. legacy base images are not supported.
2. WORKSPACE dependency management is incomplete.
3. go-containerregistry does not do HEAD checks, so could be a performance regression for some.

Happy to take it forward if people are interested.

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -31,3 +31,8 @@ config_setting(
     name = "optimized",
     values = {"compilation_mode": "opt"},
 )
+
+load("@bazel_gazelle//:def.bzl", "gazelle")
+
+# gazelle:prefix github.com/bazelbuild/rules_docker
+gazelle(name = "gazelle")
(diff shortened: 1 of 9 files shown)


paulangton:
Would love to see this, frustrating to deal with on machines with `/usr/bin/python` as python3

siddharthab:
I agree. Internally, we have exposed the pusher attribute on the rules and have been using the Go program. I can work on the PR more if I get some buy in from the repo owners. Puller and join_layers should also be very similar eliminating all httplib2 paths.

nlopezgi:
Please see the PR that I sent (#490) that should provide a viable solution to avoid to have to make these large changes.

PR #483, opened 2018-08-07
Add test config to run against RBE.

xingao267:


.bazelci/presubmit.yml
```diff
@@ -17,4 +17,130 @@ platforms:
     - "--"
     - "..."
     - "-//tests/docker/..."
-
+  rbe_ubuntu1604:
+    build_targets:
+    - "--"
```
nlopezgi: add a comment that we are selecting to test only the targets that build on RBE because the default image does not support docker

.bazelci/presubmit.yml
```diff
@@ -17,4 +17,130 @@ platforms:
...
+    - "//docker/..."
+    - "//skylib/..."
+    - "//container:load"
```
nlopezgi: try to see if you can do //container/...

.bazelci/presubmit.yml
xingao267: Done.

.bazelci/presubmit.yml
xingao267: Done.

PR #485, opened 2018-08-10
Streamline local test on CI.

xingao267:


.bazelci/presubmit.yml
```diff
@@ -4,12 +4,22 @@ platforms:
     build_targets:
     - "..."
     test_targets:
-    - "..."
+    - "//tests/contrib:test_compare_ids_test_diff_ids_fails"
```
nlopezgi: Add note we're running these remotely only

.bazelci/presubmit.yml
```diff
@@ -4,12 +4,20 @@ platforms:
...
     - "..."
     test_targets:
+    # We are running the rest of test targets remotely only.
+    - "--"
     - "..."
+    - "-//tests/container/..."
+    - "-//tests/docker/..."
   ubuntu1604:
     build_targets:
     - "..."
```
nlopezgi: lets also remove some of the build targets, with note we're doing them remotely

xingao267: Done.

PR #491, opened 2018-08-16
adding more tests to RBE runs

nlopezgi:
* id tests were failing to pick up the right CC on RBE, adding the action_env fixes the issue


nlopezgi:
Unfortunately we need to use a container with Bazel + Docker in order to get all these tests to pass. Will abandon this PR for now.

PR #493, opened 2018-08-16
Adding Basic Support for Windows Docker Images

nowens03:
container_import
Added a parameter called “manifest”. This allows Bazel to import a Windows base image without checking in GBs of tar files. All that is needed is the config and manifest files from the base Windows image. The layers parameter should be an empty list.

container_import(
name = "windowsservercore_1803",
config = "windowsservercore.1803.config.json",
manifest = "windowsservercore.1803.manifest.json",
layers = []
)

container_image
Added an optional parameter to docker_build called “operating_system.” While most of the changes are foreign layer specific, two changes were required to address Windows specifically.

There is no support for adding registry diffs, but diffs in existing layers will be respected and continue to work.

container_image(
name = "basic_windows_image",
base = ":import_windows_base_image",
cmd = ['bar.exe'],
files = [":bar.exe"]
operating_system = "windows",
) 

#488


nlopezgi:
ugh, no idea why check for cla is still complaining. @nowens03 can you please provide written confirmation here that you accept the cla and we can then merge overriding the check. Sorry for the extra hassle.

nlopezgi:
For record. This PR was reviewed as https://github.com/bazelbuild/rules_docker/pull/488 and then moved here to solve the cla check issue.

nowens03:
I accept the CLA

PR #502, opened 2018-08-23
Fix an import bug in py3_image.py.

xingao267:


xingao267:
@nlopezgi There is no py3_image_library.py file. The only *library.py exist is py_image_library.py. I discovered this when I tried to do `bazel run //testdata:py3_image`

PR #504, opened 2018-08-24
null_entrypoint bug fix.

xingao267:


xingao267:
We do have tests to test from macros https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD#L61, but not from rules

PR #505, opened 2018-08-27
Expose docker run info in container_image rule.

### LINKED ISSUE #482: docker_run_flags not passed to docker run of lang_image
Not sure if this is intended, but I want to pass certain env variables on bazel run of the py3_image below, the docker_run_flags can only be set to a container_image and not on the py3_image. Can the docker_run_flags be added to lang_image rules, or can it be inherited from the base image?

```
container_image(
    name = "py3_image_1",
    base = "@py3_image_base//image",
    legacy_run_behavior = False,
    docker_run_flags = "-i --rm --network=host -e ABC=ABC"
)

py3_image(
    name = "python_image",
    srcs = ["main.py"],
    base = ":py3_image_1",
    layers = [
        ":main",
    ],
    main = "main.py",
)
```

xingao267:
Fix #482.

### DIFF
--- a/container/providers.bzl
+++ b/container/providers.bzl
@@ -20,7 +20,11 @@ BundleInfo = provider(fields = ["container_images", "stamp"])
 FlattenInfo = provider()
 
 # A provider containing information exposed by container_image rules
-ImageInfo = provider(fields = ["container_parts"])
+ImageInfo = provider(fields = [
+    "container_parts",
+    "legacy_run_behavior",
+    "docker_run_flags",
+])
 
 # A provider containing information exposed by container_import rules
 ImportInfo = provider(fields = ["container_parts"])
@@ -34,7 +38,7 @@ LayerInfo = provider(fields = [
     "env",
 ])
 
-# A provier containing information exposed by container_push rules
+# A provider containing information exposed by container_push rules
 PushInfo = provider(fields = [
     "registry",
     "repository",
--- a/java/image.bzl
+++ b/java/image.bzl
@@ -40,7 +40,6 @@ load(
     ":jetty.bzl",
     _JETTY_DIGESTS = "DIGESTS",
 )
-load("//container:providers.bzl", "ImageInfo")
 
 def repositories():
     # Call the core "repositories" function to reduce boilerplate.
@@ -182,14 +181,6 @@ def _jar_app_layer_impl(ctx):
         for f in unavailable + [classpath_file]
     }
 
-    # legacy_run_behavior and docker_run_flags from base override those from
-    # ctx.
-    legacy_run_behavior = ctx.attr.legacy_run_behavior
-    docker_run_flags = ctx.attr.docker_run_flags
-    if ImageInfo in ctx.attr.base:
-        legacy_run_behavior = ctx.attr.base[ImageInfo].legacy_run_behavior
-        docker_run_flags = ctx.attr.base[ImageInfo].docker_run_flags
-
     return _container.image.implementation(
         ctx,
         # We use all absolute paths.
@@ -197,8 +188,6 @@ def _jar_app_layer_impl(ctx):
         file_map = file_map,
         entrypoint = entrypoint,
         workdir = workdir,
-        legacy_run_behavior = legacy_run_behavior,
-        docker_run_flags = docker_run_flags,
     )
 
 jar_app_layer = rule(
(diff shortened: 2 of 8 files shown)


container/providers.bzl
```diff
@@ -20,7 +20,11 @@ BundleInfo = provider(fields = ["container_images", "stamp"])
 FlattenInfo = provider()
 
 # A provider containing information exposed by container_image rules
-ImageInfo = provider(fields = ["container_parts"])
+ImageInfo = provider(fields = [
+    "container_parts",
```
nlopezgi: Why not make these part of container_parts?

java/image.bzl
```diff
@@ -180,13 +182,23 @@ def _jar_app_layer_impl(ctx):
         for f in unavailable + [classpath_file]
     }
 
+    # legacy_run_behavior and docker_run_flags from base override those from
```
nlopezgi: I dont like this code being duplicated so much. This is just accessing ctx, and passing values to call of  _container. Cant we just do all this in container/image.bzl

xingao267: Good point. moved to container/image.bzl

container/providers.bzl
xingao267: I feel container_parts apply more to the static image tarball itself. But the run flags are more about the runtime?

testdata/BUILD
```diff
@@ -677,6 +677,24 @@ py3_image(
     main = "py3_image.py",
 )
 
+container_image(
+    name = "py3_image_base_with_custom_run_flags",
+    base = "@py3_image_base//image",
+    cmd = ["-V"],
+    docker_run_flags = "-i --rm --network=host -e ABC=ABC",
+    legacy_run_behavior = False,
```
nlopezgi: should we have a test that uses legacy_run_behavior=true ?

testdata/BUILD
```diff
@@ -677,6 +677,24 @@ py3_image(
     main = "py3_image.py",
 )
 
+container_image(
+    name = "py3_image_base_with_custom_run_flags",
+    base = "@py3_image_base//image",
+    cmd = ["-V"],
```
nlopezgi: why do you set command here?

testdata/BUILD
nlopezgi: Also, why is this necessary here (i.e., what happens if you just don't set it, as I see none of the examples currently use legacy_run_behavior)?

nlopezgi:
i think you might need to update the pin to bazel-toolchains?

testdata/BUILD
xingao267: Because for py3 images, for example, the entry point is python, if I don't supply something after it, it will hang.

testdata/BUILD
xingao267: The default value for legacy_run_behavior is True. docker_run_flags is only used when legacy_run_behavior is false.

testing/e2e.sh
```diff
@@ -116,19 +133,23 @@ function test_bazel_build_then_run_docker_build_clean() {
   cd "${ROOT}"
   for target in $(bazel query 'kind("container_image", "testdata/...")');
   do
-    clear_docker
-    bazel build $target
-    # Replace : with /
-    ./bazel-bin/${target/://}
+    if [[ $(SHOULD_SKIP $target) -eq 1 ]]; then
```
nlopezgi: You should do a custom bazel query that removes the targets you dont need.
e.g., I ran this successfully :
bazel query 'kind("container_image", "testdata/..." except ("//testdata:absolute_data_path_image" union "//testdata:with_user" union "//testdata:dashdash_entrypoint"))' )

xingao267: Done.

Had to duplicate some code... It's also ugly to make the additional part of the query a constant string, and I can't do simply join on a list in bash like targetsToSkip.join(" union ") either...

testing/e2e.sh
```diff
@@ -153,7 +159,10 @@ function test_bazel_run_docker_import_clean() {
 function test_bazel_run_docker_build_incremental() {
   cd "${ROOT}"
   clear_docker
-  for target in $(bazel query 'kind("container_image", "testdata/...")');
+  for target in $(bazel query 'kind("container_image", "testdata/...") except
+    ("//testdata:py3_image_base_with_custom_run_flags" union
+    "//testdata:java_image_base_with_custom_run_flags" union
+    "//testdata:war_image_base_with_custom_run_flags")');
```
nlopezgi: Lets just put this in a constant

xingao267: Done.

Issue #513, opened 2018-09-06
Windows build machine cannot produce **any** working builds


zaneclaes:
**tl;dr**: it seems that `guess_runfiles` (which is copy-pasta-ed all over the repo) makes invalid assumptions about the output directory.

Perhaps I'm missing something in the docs, but it appears that the `[OUT].runfiles` directory yields only a `manifest` file, which subsequently causes anything after the `build` image step to fail. Note that everything tried here works fine on OSX. And the `build` step succeeds; I just can't `run` to turn anything into an image (or tag it, for that matter).

Assuming a simple container `BUILD` (n.b., `@filebeat` is just a pull of `docker.elastic.co/beats/filebeat:6.4.0`).

```
container_image(
  name = "DEV",
  base = "@filebeat//image",
)
```

Attempt 1: Running Directly
----
`bazel run //service/logpipe:DEV`

```
INFO: Analysed target //service/logpipe:DEV (25 packages loaded).
INFO: Found 1 target...
Target //service/logpipe:DEV up-to-date:
  C:/users/zane/_bazel_zaneclaes/eojxvnhj/execroot/__main__/bazel-out/x64_windows-fastbuild/bin/service/logpipe/DEV-layer.tar
INFO: Elapsed time: 306.810s, Critical Path: 13.48s
INFO: 21 processes: 21 local.
INFO: Build completed successfully, 43 total actions
INFO: Running command line: C:/users/zane/_bazel_zaneclaes/eojxvnhj/execroot/__main__/bazel-out/x64_windows-fastbuildINFO: Build completed successfully, 43 total actions
/bin/bash: C:/users/zane/_bazel_zaneclaes/eojxvnhj/execroot/__main__/bazel-out/x64_windows-fastbuild/bin/service/logpipe/DEV: No such file or directory
```

If I run the actual output script (`DEV`) from the shell it complains more explicitly about the files missing from the `DEV.runfiles`. Inspecting this directory, I see only the `MANIFEST` file.

Attempt 2: Exporting a Tar
----
`bazel run //service/logpipe:DEV.tar`

```
INFO: Analysed target //service/logpipe:DEV.tar (0 packages loaded).
INFO: Found 1 target...
Target //service/logpipe:DEV.tar up-to-date:
  C:/users/zane/_bazel_zaneclaes/eojxvnhj/execroot/__main__/bazel-out/x64_windows-fastbuild/bin/service/logpipe/DEV.tar
[...]

PR #514, opened 2018-09-06
Allow incremental_load to work on Windows

zaneclaes:
@nlopezgi @enriched I've fixed the issue described in https://github.com/bazelbuild/rules_docker/issues/513

**note** this still doesn't execute in "one step." If I `bazel run` I get a `/bin/bash: C:\XXXXX: No such file or directory` (even if I include `-- --norun`). I have to execute the script manually after the build. I'm also not an expert with (ba)sh, so I'm not proud of the syntax.

This is "good enough" for me, in that executing `[the_output_file]` now works. I can simply execute the `bash` script, i.e., type `DEV` in Mingw64 and it will produce the expected `loaded` and `tagging` text, which results in a working Docker image attached to my local Windows daemon. I've tested it with `container_bundle` and successfully launched all 3 of the containers it produced. Test it yourself with the archive:
```
http_archive(
  strip_prefix = "rules_docker-runfiles",
  sha256 = "a6504be695e5458d9274dcc445e71dc5952e61f5753eeead99f17e07684613e3",
  urls = ["https://github.com/zaneclaes/rules_docker/archive/runfiles.tar.gz"],
)
```

Note, though, I haven't fixed the other uses of `guess_runfiles` which continue to proliferate. This only addresses the `incremental_load` file, which unlocks the essential build features.

### DIFF
--- a/container/incremental_load.sh.tpl
+++ b/container/incremental_load.sh.tpl
@@ -19,10 +19,27 @@ set -eu
 # This is a generated file that loads all docker layers built by "docker_build".
 
 function guess_runfiles() {
-    pushd ${BASH_SOURCE[0]}.runfiles > /dev/null 2>&1
-    pwd
-    popd > /dev/null 2>&1
+  pushd ${BASH_SOURCE[0]}.runfiles > /dev/null 2>&1
+  pwd
+  popd > /dev/null 2>&1
 }
+
+function get_runfile() {
+  if [[ -f "$RUNFILES/$1" ]]; then
+    echo "$RUNFILES/$1"
+  else
+    rel="__main__/../"
+    abs="__main__/external/"
+    fn=${1/$rel/$abs}
+    echo "${MANIFEST[$fn]}"
+  fi
+}
+
+declare -A MANIFEST
+while IFS='' read -r line || [[ -n "$line" ]]; do
+  IFS=$' ' read -r from to <<< "$line"
+  MANIFEST["$from"]="$to"
+done < "${BASH_SOURCE[0]}.runfiles/MANIFEST"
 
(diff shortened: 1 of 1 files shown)


nlopezgi:
Thanks for sending this PR. Given the changes you needed to make to get this working, my impression is that rules_docker has deep issues that currently prevent it from working on windows.
However, there is upcoming work by the Bazel team that will help make this easier to fix (https://github.com/bazelbuild/bazel/issues/4319 - https://docs.google.com/document/d/e/2PACX-1vRCSB_n3vctL6bKiPkIa_RN_ybzoAccSe0ic8mxdFNZGNBJ3QGhcKjsL7YKf-ngVyjRZwCmhi_5KhcX/pub). I do have plans to have our team do some of the work (related to how tools and bash is being used in the rules) which is a prerequisite for these rules to work on windows (planned for the upcoming weeks). I will keep you informed. At the moment, however, the changes you make are not the right solution to support windows builds. I will remove the statement regarding Windows support from the docs and replace it with a note regarding our upcoming work to get this supported properly.

zaneclaes:
That makes perfect sense. I hacked together the script so that my CI would pass. My situation is a bit interesting. Right now I'm building with Windows and only creating Docker images... but in the future, I will move it to Linux (AWS, really) and have more (non-Docker) targets. So I wanted to integrate and familiarize myself with Bazel early-on.

All this to say: @nlopezgi please tag me whenever any significant updates occur. I'm going to re-focus on my core needs for now, but would like to bring everything in-line with best practices as they become available.

nlopezgi:
thanks for your understanding and for telling me more about your usage of rules_docker (I need to know more about our users!). I'll keep you informed once we have a more detailed plan of our upcoming efforts to solve the problems related to windows.

Issue #511, opened 2018-09-03
container_push from contrib/push-all.bzl fails on empty images dict


stevewolter:
When run on an empty images dict, the container_push fails on some shells with:

line 36: PIDS[@]: unbound variable

I realize that this is a contrib/ file and it's easy enough to work around, so I'm posting this for tracking and don't expect a fix from the maintainers. Thanks for maintaining rules_docker, it's awesome!

drigz:
Specifically, it fails with bash 4.3 but works with bash 4.4. It's not clear to me whether bash 4.3 or 4.4 behaves as expected, as the docs in 4.4 suggest that the error should still occur. For more discussion, [see the comments on this SO answer](https://stackoverflow.com/a/7577209).

xingao267:
It looks like the push-all.sh.tpl should probably be updated to adapt this.

smukherj1:
I locally reproduced this on Ubuntu 16.04 with bash 4.3.48. Verified #516 fixes the issue and does not break on bash 4.4. Also added a test for the empty bundle case to //tests/contrib. Closing this issue. Please  reopen if the fix doesn't work for you

PR #516, opened 2018-09-07
Issue511

smukherj1:
Add a test target to create an empty container and attempt to push it using docker_push which internally calls docker push. Verified bazel run passes on bash 4.4 and gives the unbound variable error on bash 4.3. Verified both versions pass after the fix


nlopezgi:
https://github.com/bazelbuild/rules_docker/issues/511

Issue #506, opened 2018-08-28
java_image doesn't work with runfiles


bjchambers:
A `java_binary` target with `data` dependencies (either directly or indirectly) doesn't seem to work when converted to a `java_image` target. If used with `com.google.devtools.build.runfiles.Runfiles` to access the runfiles, it gives the following error:

```
Caused by: java.io.IOException: Cannot find runfiles: $RUNFILES_DIR and $JAVA_RUNFILES are both unset or empty
```

It looks like the `java_image` rule doesn't set the environment parameters. Is there some other mechanism for accessing data dependencies from within a `java_image`?

nlopezgi:
Thanks for opening this issue. 

Could you provide a bit more info about how to reproduce the error you are getting? 

My initial thought is that $RUNFILES_DIR and $JAVA_RUNFILES are make variables that should be properly expanded in the rules, and we might not be expanding them correctly atm (i.e., we recently had changes to expand args and location https://github.com/nlopezgi/rules_docker/commit/f5ed963f7411f07efd6b0bd8667fe89d6a05077b
https://github.com/nlopezgi/rules_docker/commit/115d891e359ea3097cd0bc1b32e1aaf8c580c455). 

If you could also try to send a PR with a fix (if these PRs are enough to get you on the right path) it would be awesome! (otherwise will try to get to take a detailed look as soon as I can once you send repro instructions)

bjchambers:
They are environment variables (https://github.com/bazelbuild/bazel/blob/master/tools/java/runfiles/Runfiles.java#L174) set by the Java wrapper (https://github.com/bazelbuild/bazel/blob/master/src/main/java/com/google/devtools/build/lib/bazel/rules/java/java_stub_template.txt#L140).

I'll try to put together a minimal repro and send it.

bjchambers:
Added two entries to testdata in this branch: https://github.com/bjchambers/rules_docker/tree/java-data-runfiles

When running `bazel run //testdata:java_runfiles_image` or `bazel run //testdata:java_runfiles_as_lib_image` the following error is printed:

```
Exception in thread "main" java.io.IOException: Cannot find runfiles: $RUNFILES_DIR and $JAVA_RUNFILES are both unset or empty
	at com.google.devtools.build.runfiles.Runfiles.getRunfilesDir(Runfiles.java:180)
	at com.google.devtools.build.runfiles.Runfiles.create(Runfiles.java:114)
[...]

PR #517, opened 2018-09-07
Load data files in java_image 

nlopezgi:
UPDATE: this CL now adds all transitive data files in a java_image (feature guarded by env variable). Activating this feature enables a java_image rule to depend on a java_library rule and have files for the java_library made available inside the container. 
To achieve this, however, we are getting **all** transitive dependencies included in the image. Note, since the transitive deps of a java_xx rule include the java toolchain it is HEAVILY RECOMMENDED that build/run/test commands that will build a java_image rule do not use the @local_jdk.
To enable easy overriding of the @local_jdk, this PR also includes a repository rule that creates a remote repo @local_config_java//:jdk. This repo rule will point to the JAVA_HOME, overriding the @local_jdk. To use this target your build must pass flag --javabase=@local_config_java//:jdk (see bazelrc file added to this repo in this commit). Note: if feature is not enabled, @local_config_java//:jdk will point to the @local_jdk to avoid any breakages.

**To enable this feature the env variable EXPERIMENTAL_TRANSITIVE_JAVA_DEPS=1 must be set, JAVA_HOME must be set and --javabase=@local_config_java//:jdk must be passed )via bazel rc or directly). The easiest way to do this is to just export these env variables but its also feasible to do something like:

`bazel build --define=EXPERIMENTAL_TRANSITIVE_JAVA_DEPS=1 --action_env=EXPERIMENTAL_TRANSITIVE_JAVA_DEPS=1 --javabase=@local_config_java//:jdk  <some target>`

(note both define and action_env must be used as this env var will be read in repository rule and skylark rule, sorry for this nastiness)
Note until java rules support selection of platform/toolchain on a per-target basis it is not possible to solve https://github.com/bazelbuild/rules_docker/issues/506 in any other way that I can think of (i.e., the main regression this feature has is that java_home must be present in the env, otherwise builds will fail, thus guarding it with the additional EXPERIMENTAL_TRANSITIVE_JAVA_DEPS env var).

This PR also adds the JAVA_RUNFILES env variable to java_image to get behavior on par with java_binary targets that require use of this env variable (not guarded by flag).

Previous description (for record):
https://github.com/nlopezgi/rules_docker/commit/f5ed963f7411f07efd6b0bd8667fe89d6a05077b added data files to lang_image (and transitively to all xx_image rules that use app_layer). 
java_image does not use app_layer so the change did not work in that case. 

Also setting workdir for java_image to "/app/io_bazel_rules_docker/" (experimental, please report any issues with this workdir change)

This fix partially addresses https://github.com/bazelbuild/rules_docker/issues/506

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -14,6 +14,12 @@ platforms:
     - "..."
     - "-//tests/container/..."
     - "-//tests/docker/..."
+    build_flags:
+    - "--define=EXPERIMENTAL_TRANSITIVE_JAVA_DEPS=1"
+    - "--action_env=EXPERIMENTAL_TRANSITIVE_JAVA_DEPS=1"
+    test_flags:
+    - "--define=EXPERIMENTAL_TRANSITIVE_JAVA_DEPS=1"
+    - "--action_env=EXPERIMENTAL_TRANSITIVE_JAVA_DEPS=1"
   ubuntu1604:
     build_targets:
     # We are running the skipped targets remotely only.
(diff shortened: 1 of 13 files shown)


nlopezgi:
Abandoning in favor of https://github.com/bazelbuild/rules_docker/pull/530

PR #522, opened 2018-09-15
Fix runfiles discovery

scele:
The launcher scripts were not able to find runfiles when executed
deeper inside a runfile tree generated by another rule.  This
happens when another rule depends on the container image target
and wants to execute the scripts from its own runfiles tree.

Fix discovery to also account for this possibility, using similar
heuristics as FindModuleSpace in the python launcher.

### DIFF
--- a/container/incremental_load.sh.tpl
+++ b/container/incremental_load.sh.tpl
@@ -19,9 +19,14 @@ set -eu
 # This is a generated file that loads all docker layers built by "docker_build".
 
 function guess_runfiles() {
-    pushd ${BASH_SOURCE[0]}.runfiles > /dev/null 2>&1
-    pwd
-    popd > /dev/null 2>&1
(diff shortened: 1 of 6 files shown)


xingao267:
Hi @scele thanks for you contribution. Can you provide some test cases for this?

xingao267:
Here of places where you can add the tests:
- Test targets (e.g. container_image) can be added under https://github.com/bazelbuild/rules_docker/blob/master/testdata/BUILD
- Checks of runfiles can be added under https://github.com/bazelbuild/rules_docker/blob/master/container/image_test.py

And it would be nice if you can provide a bit more comments in the code.

Thanks again for your contribution!

nlopezgi:
Hi @scele, please let me know if you are having issues creating a test for this so we can help you out.

PR #530, opened 2018-09-25
Java data runfiles

bjchambers:
Add examples using data dependencies with Java the runfiles library. These demonstrate the problem described in #506.

Fix the java_image rule so that these examples work. This required an additional environment variable being passed when executing the binary, and adding the data dependencies to the files that are part of the image.

This in turn required adding the `_jdk` attribute, so the files that appear to be data dependencies but come from the JRE may be excluded by marking them as already available.


bjchambers:
I signed it!

testdata/BUILD
```diff
@@ -796,6 +796,53 @@ java_image(
     runtime_deps = [":java_bin_as_lib"],
 )
 
+java_binary(
```
nlopezgi: can you remove this target, though useful for your testing we dont need it here.

testdata/BUILD
```diff
@@ -796,6 +796,53 @@ java_image(
...
+    ],
+)
+
+java_image(
+    name = "java_runfiles_as_lib_image",
+    main_class = "examples.images.Runfiles",
+    runtime_deps = [":java_runfiles_as_lib"],
+)
+
+java_binary(
```
nlopezgi: same for this one

bjchambers:
Added the tests to image_test.py. Is there a way to add a test that the container actually runs and behaves properly? Specifically, the tests as they exist don't verify that the `env` is properly set to locate the runfiles, so I was thinking it would be helpful to either (1) verify that in a unit test similar to existing tests or (2) to actually try running the image...

bjchambers:
Also had to make some changes, since the "available" files were added to the classpath, to treat the "jdk_files" as a separate set that is also removed from "unavailable", but not added to the classpath.

java/image.bzl
```diff
@@ -140,21 +140,22 @@ def _jar_app_layer_impl(ctx):
...
     for jar in ctx.attr.deps + ctx.attr.runtime_deps:
         unavailable += java_files(jar)
 
     unavailable += java_files(ctx.attr.binary)
     unavailable = [x for x in unavailable if x not in available]
 
+    # Mark all files that are available from the JDK as available.
+    # This will prevent them from being part of the generated image.
+    jdk_files = depset()
+    jdk_files += list(ctx.files._jdk)
```
nlopezgi: cant you just inline?
jdk_files = depset(list(ctx.files._jdk))

java/image.bzl
```diff
@@ -140,21 +140,22 @@ def _jar_app_layer_impl(ctx):
...
     # out what it available.
-
     unavailable = depset()
     for jar in ctx.attr.deps + ctx.attr.runtime_deps:
         unavailable += java_files(jar)
 
     unavailable += java_files(ctx.attr.binary)
     unavailable = [x for x in unavailable if x not in available]
 
+    # Mark all files that are available from the JDK as available.
```
nlopezgi: Could you update the comment to reflect better what you are now doing
"removing from unavailable all files from the JDK as the image already contains the required Java toolchain" (or something like that)

testdata/BUILD
```diff
@@ -796,18 +796,6 @@ java_image(
...
-        "foo",
-    ],
-    main_class = "examples.images.Runfiles",
-    deps = [
-        "@bazel_tools//tools/runfiles:java-runfiles",
-    ],
-)
-
 java_image(
     name = "java_runfiles_image",
```
nlopezgi: could you please add a test for java_runfiles_image in container/image_test.py

bjchambers: In the process of working on this.

I'm running into some problems with layer 1 -- it looks like it passes java_files in as lambda, and that is also picking up the data files. I'm going to fix it by having a version of java_files that doesn't include the data dependencies, but it raised a question:

*should* the java_runfiles_as_lib case have the runfiles in the top layer, or the dependency layer?

nlopezgi: good question. I personally do not have enough strong arguments in favor of either. From a conceptual point of view it might seem to make sense to put in the dependency layer (iiuc). Do you have any arguments to put it in top or dep layer or can imagine any trade-offs this would bring?

bjchambers: I agree that it makes the most sense to put it in the dependency layer. However, trying to make that work I ran into problems, since that makes use of functionality from lang/image.bzl, which wasn't easily extendable to support excluding the JDK files.

My inclination is to leave it as is -- data files of dependencies in the top layer -- to get things working for the time being. The disadvantage is that a change in dependencies may require rebuilding both the dependency layer and the top layer (I think), but I'm not sure of the cost of that. It also means that every time the top layer changes, the data dependencies need to be re-tar'd, etc. Again, likely not a huge problem, and something that seems better to fix in the future by adding appropriate exclusion mechanisms to the rules that build the dep layer.

nlopezgi: sgtm, could you leave a comment in the code about this decision so that we dont loose track it.

bjchambers: Done

nlopezgi:
Tests are failing (specifically //container:image_test).
Let me know when you have fixed so I can kick off presubmits

bjchambers:
I made a change that I think will fix them. However, they are passing locally. It looks like the paths are slightly different `bazel_tools/src/tools` on my machine and `bazel_tools/tools` in the continuous build. Hopefully changing these paths will make the tests pass, but I'm not sure why the paths locally are different.

nlopezgi:
image test still seems to be failing, let me know if you're running into issues debugging

bjchambers:
Ok, I think I was able to reproduce locally. As far as I can tell, the problem was that the location of the runfiles library moved in 0.17. It didn't cause build errors because the old path remained around as an alias, but was producing a warning it was going away. Locally I had been using 0.16.2.

I think this should fix the tests.

Issue #526, opened 2018-09-21
container_push with (basic) auth


pcj:
The `container_push` rule works if the `pusher.par` can find an adequate `$HOME/.docker/config.json`  for authentication (`DOCKER_CONFIG` works too).  It would be nice to be able to provide the docker config.json file as an input to the `container_push` rule.  Any plans for this?  I can work on a PR if people are amenable to that.

nlopezgi:
Hi @pcj I think it would be great if we could provide a feature to override this default behavior. Do you have any thoughts on how to implement this properly (before going to a PR), my concern mainly is where this info is defined (once its defined somewhere passing it to pusher.par seems trivial). My initial thought is the proper way to do this is:
1. Via a repository rule (https://docs.bazel.build/versions/master/skylark/repository_rules.html - e.g., https://github.com/bazelbuild/rules_docker/blob/master/container/pull.bzl) which produces as output a new "docker_toolchain" rule target (following advice in https://docs.bazel.build/versions/master/toolchains.html) which should have (among other attrs) the location of the auth file.

There are some other ways to do this that are not acceptable, imo, particularly:
2. Via an attr in container_push that takes in a single file. The auth file has to be in the source tree. The main issue with this approach is the file has to be checked in, which will not work for an auth file.
3. Add a string attr to container_push that should take an absolute path to the auth file. This is unacceptable because it breaks Bazel hermeticity (i.e., changing the contents of the auth file might result in undefined / undesired behavior).

Doing 1 will require significant effort. If you're willing to put in some time to get this working I can help you get started. We do have plans to do work so that these rules can start using toolchain rules (lots more than just for auth for container_push) properly in the upcoming months, so it might also be an option to delay working on this issue until we have at least created the first parts of the toolchain rule which would allow you to just work on hooking that up to also work for the auth bit.

If you can think of other options to implement this let me know.

pcj:
I don't like (3) either.  I'd have to do more research about (1).

I was imagining (2) [which you blacklisted :smile:].  Here's how we do the auth file:

[...]

PR #531, opened 2018-09-26
Implement registry_credential for basic authentication

pcj:
This is a WIP/prototype for #526 that works in conjunction with https://github.com/google/containerregistry/pull/112.

### DIFF
--- a/container/credentials.bzl
+++ b/container/credentials.bzl
@@ -1,74 +1,28 @@
 load("//container:providers.bzl", "RegistryCredentialInfo")
 
-# def write_docker_config_file(ctx, registry, credential, config_file):
-#     auths = {}
-#     if cred.type != "basic":
-#         fail("Unsupported docker credential type: %s" % cred.type)
-#     auths[registry] = struct(
-#         username = cred.username,
-#         password = cred.password,
-#     )
-#     config = struct(
-#         auths = auths,
-#     )
-#     if not config_file:
-#         config_file = ctx.actions.declare_file(".docker/config.json")
-#     ctx.actions.write(config_file, config.to_json())
-#     return config_file
+def write_docker_config_json_action(ctx, registry, credential, config_json = None):
+    """Create an action that writes a docker config file.
 
-def write_docker_config_fileOLD(ctx, registry, credential, config_file = None):
+    Args:
+      ctx: the rule ctx object
+      registry: the name of the registry to which the credential should be
+      associated
+      credential: a RegistryCredentialInfo struct
+      config_json: path to the file where the config.json should be written.
+      Defaults to %{credential.name}/.docker/config.json.
+
+    Return: The config_json file 
+    """
     if credential.type != "basic":
         fail("Unsupported docker credential type: %s" % credential.type)
-    env = ctx.configuration.default_shell_env
-    print("DEFAULT? CONFIGURATION ENV: %s" % ctx.configuration.default_shell_env)
-    print("HOST CONFIGURATION ENV: %s" % ctx.host_configuration.default_shell_env)
-    username = env.get(credential.uservar)
-    if not username:
-        fail("registry auth '%s': required environment variable '%s' is not defined (should contain basic auth username)\n%r" % (registry, credential.uservar, env))
-    password = env.get(credential.passvar)
-    if not password:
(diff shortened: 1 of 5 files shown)


container/credentials.bzl
```diff
@@ -0,0 +1,118 @@
...
+    Provides the RegistryCredentialInfo object from given rule attributes.
+    """
+    return [RegistryCredentialInfo(
+        name = ctx.label.name,
+        user_env = ctx.attr.user_env,
+        pass_env = ctx.attr.pass_env,
+        type = ctx.attr.type,
+    )]
+
+registry_credential = rule(
```
nlopezgi: I'm not sure if this should be a repository rule or a skylark rule as it is now as this is a bit problematic.
Specifically, there are two issues:
- this rule will not work for container_pull, as container_pull is a repository rule and cannot access outputs of a skylark rule
- env vars like this one should be read by repository rules, not skylark rules, particularly because if the user changes the values of the env variables, Bazel will not regenerate the outputs unless you do a Bazel clean

However, I'm also not sure this should be a repository_rule. Specifically because:
- repo rules cannot have providers, so the elegant solution that you have to avoid users from checking in their secret values cannot be used here.

Thoughts?

container/credentials.bzl
```diff
@@ -0,0 +1,118 @@
...
+    implementation = _registry_credential_impl,
+    attrs = {
+        "user_env": attr.string(
```
nlopezgi: if this was a repo rule, env variables to be read would be declared in a way that Bazel knows to regenerate outputs if values change.
example:
https://github.com/bazelbuild/bazel/blob/master/tools/cpp/cc_configure.bzl#L59

container/credentials.bzl
```diff
@@ -0,0 +1,118 @@
...
+
+    Return: The config_json file 
+    """
+    if credential.type != "basic":
+        fail("Unsupported docker credential type: %s" % credential.type)
+    if not config_json:
+        config_json = ctx.actions.declare_file("%s/.docker/config.json" % credential.name)
+    # NOTE: ctx.actions.write does not work here as we need the
+    # use_default_shell_env feature of the run_shell action.
+    ctx.actions.run_shell(
```
nlopezgi: you should use a template file + expand_template to avoid using shell to create a file (which would only work on environments that have bash)  https://docs.bazel.build/versions/master/skylark/lib/actions.html#expand_template

container/credentials.bzl
pcj: You're right about `container_pull` would need a different mechanism.  I'll have to think about how to reconcile this.

Regarding changing of environment variables.  The implementation of `--action_env` includes the env vars into the SkyKey (or whatever the action hash is called), so in fact it does re-run the rule if you change the value of the environment variable.

container/credentials.bzl
pcj: See comment above.  It doesn't have to be a repository rule.

container/credentials.bzl
pcj: Yes that was my initial thought.  However, in order to use `--action_env` you MUST use a `run` or `run_shell` action, which has the `use_default_shell_env` feature.

container/credentials.bzl
nlopezgi: You're right about env vars, did not remember the action key does include those values. 
let me know what you think about container_pull

container/credentials.bzl
nlopezgi: I really dislike having this code here. We're in process of fixing any instances of usage of tools that will block these rules from working with remote execution or with windows. The way to avoid this hard restriction (--action_env you MUST use a run or run_shell), I think we could use ctx.var (https://docs.bazel.build/versions/master/skylark/lib/ctx.html#var). The main problem with that is that we would not get the env var to be part of the action key. In any case, it should at least be possible to use templates to generate the file with the contents you are generating here using format, and then use run_shell to execute the file produced by the template function.
However, I think we need to hash out the issue about how to do this for container_pull and that will likely require change how were doing this with --action_env. I don't want there to be two different paths to achieve the same thing for container_pull/push.

nlopezgi:
Closing this PR as this feature will be implemented in https://github.com/bazelbuild/rules_docker/pull/531 via toolchain rules

pcj:
SGTM

PR #535, opened 2018-10-04
Add support for print_progress to container_pull

scele:
Add minimal prints to indicate puller progress if container_pull
has print_progress=True attribute.  The intention is that the pull
operation would provide some feedback to the user, instead of bazel
getting seemingly stuck for 10 minutes when pulling an image that is
many gigabytes in size.

Pending pull request to upstream puller.par is here:
https://github.com/google/containerregistry/pull/66

### DIFF
--- a/container/pull.bzl
+++ b/container/pull.bzl
@@ -80,6 +80,10 @@ container_import(
     if "PULLER_TIMEOUT" in repository_ctx.os.environ:
         kwargs["timeout"] = int(repository_ctx.os.environ.get("PULLER_TIMEOUT"))
 
+    if repository_ctx.attr.print_progress:
+        kwargs["quiet"] = False
+        args += ["--print-progress"]
+
     result = repository_ctx.execute(args, **kwargs)
     if result.return_code:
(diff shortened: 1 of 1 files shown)


scele:
@nlopezgi any thoughts on this? We've had problems with developers reporting "bazel is stuck", and turns out it was just pulling a big image for 10 minutes. There are other ways to expose this, for example adding `quiet=True|False` and `puller_args=[string]` flags. Or ideally workspace rules could support `mnemonic`, like actions, but that would need to go into bazel core..

If you agree something like this is a good idea, perhaps you could vouch for the related puller.par change that hasn't gotten much attention here: https://github.com/google/containerregistry/pull/66

ittaiz:
+100 this is a big problem for us as well

ittaiz:
Is it worth while adding the attribute and quiet=false for now? It will bring better info for failure cases I think

nlopezgi:
is this PR still relevant now that we're moving to go containerregistry? let me know if so, otherwise i'll close.

scele:
Not in its current form at least. Closing.

Issue #273, opened 2017-12-22
Puller not working with (custom) system certs


abergmeier:
We have custom certs at my company. These are installed in `/usr/local/share/ca-certificates` and work fine with Browsers and most tools (on Ubuntu 16.04.3).
Sadly _puller_ is the exception. It fails with:

    File "/usr/lib/python2.7/runpy.py", line 174, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
    File "/usr/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
    File "/home/user/.cache/bazel/_bazel_user/39b2d6f0d213152194c5e70c47d440c6/external/puller/file/puller.par/__main__.py", line 93, in <module>
    File "/home/user/.cache/bazel/_bazel_user/39b2d6f0d213152194c5e70c47d440c6/external/puller/file/puller.par/__main__.py", line 80, in main
    File "/home/user/.cache/bazel/_bazel_user/39b2d6f0d213152194c5e70c47d440c6/external/puller/file/puller.par/containerregistry/client/v2_2/docker_image_.py", line 324, in __enter__
    File "/home/user/.cache/bazel/_bazel_user/39b2d6f0d213152194c5e70c47d440c6/external/puller/file/puller.par/containerregistry/client/v2_2/docker_http_.py", line 191, in __init__
    File "/home/user/.cache/bazel/_bazel_user/39b2d6f0d213152194c5e70c47d440c6/external/puller/file/puller.par/containerregistry/client/v2_2/docker_http_.py", line 217, in _Ping
    File "/home/user/.cache/bazel/_bazel_user/39b2d6f0d213152194c5e70c47d440c6/external/puller/file/puller.par/containerregistry/transport/transport_pool_.py", line 62, in request
    File "/home/user/.cache/bazel/_bazel_user/39b2d6f0d213152194c5e70c47d440c6/external/puller/file/puller.par/httplib2/__init__.py", line 1659, in request
    File "/home/user/.cache/bazel/_bazel_user/39b2d6f0d213152194c5e70c47d440c6/external/puller/file/puller.par/httplib2/__init__.py", line 1399, in _request
    File "/home/user/.cache/bazel/_bazel_user/39b2d6f0d213152194c5e70c47d440c6/external/puller/file/puller.par/httplib2/__init__.py", line 1319, in _conn_request
[...]

PR #539, opened 2018-10-10
Proposal/WIP: adding support for external cacerts.txt file

roblg:
**Note: Work based on (and depends on) https://github.com/google/containerregistry/pull/89**

Addresses  #273 

Wanted to optimistically start the ball rolling on a possible implementation to take advantage of google/containerregistry#89 providing a `--cacert` flag. _I don't believe this is a complete implementation._

Proposal: manage cacerts.txt like any other external dependency: in the WORKSPACE.

Non-working example for illustration:

```
# WORKSPACE
http_file(
    name='cacerts',
    urls=['https://example.com/cacerts.txt'],
    sha256='<sha256 here>',
)

container_pull(
    name = "favorite",
    registry = "private.myregistry.com",
    repository = "favorites/container",
    digest = "sha256:<digest>",
    cacerts = "@cacerts//file:downloaded",
)
```

I was able to make this work and resolve the issue described in #273 using this approach. (Due to `container_pull` dependency on a prebuilt `puller` artifact from google/containerregistry, in order to _really_ make this work, you also have to separately `bazel build //:puller.par` in containerregistry and add it to your workspace as `puller`.)

I'm open to alternative suggestions as well, this just seemed like a reasonable and reasonably quick route to get from the changes proposed to containerregistry and me having a working build :)

Feedback welcome!

### DIFF
--- a/container/pull.bzl
+++ b/container/pull.bzl
@@ -56,6 +56,12 @@ container_import(
         repository_ctx.path("image"),
     ]
 
+    if repository_ctx.attr.cacerts:
+        args += [
+            "--cacert",
+            repository_ctx.path(repository_ctx.attr.cacerts),
+        ]
+
     # If a digest is specified, then pull by digest.  Otherwise, pull by tag.
     if repository_ctx.attr.digest:
         args += [
@@ -90,6 +96,10 @@ container_pull = repository_rule(
         "repository": attr.string(mandatory = True),
         "digest": attr.string(),
         "tag": attr.string(default = "latest"),
+        "cacerts": attr.label(
+            allow_single_file = True,
+            mandatory = False,
+        ),
         "_puller": attr.label(
             executable = True,
             default = Label("@puller//file:downloaded"),
--- a/container/push-tag.sh.tpl
+++ b/container/push-tag.sh.tpl
@@ -28,4 +28,4 @@ function guess_runfiles() {
 
 RUNFILES="${PYTHON_RUNFILES:-$(guess_runfiles)}"
 
-%{container_pusher} %{format} --name=%{tag} %{stamp} %{image} "$@"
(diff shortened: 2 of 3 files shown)


nlopezgi:
Thanks for sending this PR. Will look at it in detail next week (in a conference all this week), in the meantime, can you check out https://github.com/bazelbuild/rules_docker/pull/531 and let us know if these two issues are related (seems to me) and we can have a conversation with @pcj to figure out how to address these use cases? Thanks!

roblg:
@nlopezgi no worries; no rush on my end.

I checked out #531, and I'll want that functionality too, eventually :) It feels related, but separate from the issue I'm trying to tackle here. This PR is specifically about being able to override the (incomplete) cacerts.txt file that's bundled with httplib2 down in the depths of containerregistry with one that includes our root CA (or, in others' cases, their internal/private CA).

nlopezgi:
thanks for the clarifications. This does seem like it would be a useful feature. Is the cacerts something you envision setting for each container_pull rule or is this more of a global setting that you would want all your rules_docker rules to run with. If the latter, then I dont think this is the best approach. We need to create a toolchain rule for docker and this looks like a piece of information that could be stored there. In any case, you probably do need the upstream changes in containerregistry first. We plan on working on a toolchain rule for rules_docker soon, so I'll keep you posted on that.

roblg:
>  Is the cacerts something you envision setting for each container_pull rule or is this more of a global setting that you would want all your rules_docker rules to run with.

Definitely would prefer a single, global setting. I wasn't sure how to achieve that with my bazel knowledge, but the toolchain thing sounds promising if it provides that. I'll go read about toolchains.

Going to close this since it seems like the fix will end up being radically different. Thanks for reviewing!

PR #542, opened 2018-10-12
Print digest mismatch on failure in compare_ids_test

smukherj1:
Helps update shas for container reproduction tests for to debug them
when run on machines where remote access isn't an option

### DIFF
--- a/contrib/compare_ids_test.py
+++ b/contrib/compare_ids_test.py
@@ -38,24 +38,24 @@ def compare_ids(tars, id_=None):
     tars: list of str paths to the image tarballs
     id_: (optional) the id we want the images to have
           if None, just makes sure they are all the same
-
+  Raises:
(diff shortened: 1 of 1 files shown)


contrib/compare_ids_test.py
```diff
@@ -38,24 +38,24 @@ def compare_ids(tars, id_=None):
...
-
   for image in tars:
     current_id = get_id(image)
     if id_ is None:
       id_ = current_id
     elif current_id != id_:
-      exit(1)
-
-  exit(0)
+      raise RuntimeError("Digest mismatch: actual {} vs expected {}".format(
```
xingao267: Why not just print a message before exit(1). But I don't have a strong preference between the two.

smukherj1: Personally I prefer exceptions because it gives a full stack when if I need to debug further

PR #544, opened 2018-10-13
container_pull: return the resolved digest to bazel and support overriding platform

### LINKED ISSUE #543: container_pull implementation should return the digest
Bazel 0.19 will have an experimental `bazel sync` command, which can be used to resolve floating dependencies. For example, as described in https://blog.bazel.build/2018/07/09/bazel-sync-and-resolved-file.html, this is used to pin a git branch to a specific commit.

It'd be useful to have similar functionality for `container_pull`. That way, I could write something in WORKSPACE like
```python
container_pull(
    name = "example",
    registry = "gcr.io",
    repository = "example/image",
    tag = "latest",
)
```
but then be able to run `bazel sync` to pin `gcr.io/example/image:latest` to a specific digest.

From talking to @aehlig and reading the commit that implemented this functionality for `git_repository` (https://github.com/bazelbuild/bazel/commit/b42734f4213e505d9582def4cd25bd510f0909a9), I believe that we'd mostly just need to make sure that the implementation of `container_pull` returns a tuple of its args, additionally including the resolved digest. 

I'm not sure what plumbing or changes are necessary in google/containerregistry to support this.

Another open question: if a tag resolves to a manifest list, should we return the digest of the manifest list, or the resolved platform-specific image digest? (I'd probably vote for the manifest list.)

ixdy:
This potentially fixes #543.

It's roughly modeled after the fix for `git_repository` in https://github.com/bazelbuild/bazel/commit/b42734f4213e505d9582def4cd25bd510f0909a9.

It depends on https://github.com/google/containerregistry/pull/113 (and a corresponding update of that dependency in `rules_docker`), so this should not be merged as-is.


container/pull.bzl
```diff
@@ -84,20 +99,15 @@ container_import(
...
-)
+    updated_attrs = {
+        k: getattr(repository_ctx.attr, k)
+        for k in _container_pull_attrs.keys()
+    }
+    digest_result = repository_ctx.execute(["cat", repository_ctx.path("image/digest")])
+    if digest_result.return_code:
+        fail("Failed to read digest: %s" % digest_result.stderr)
+    updated_attrs["digest"] = digest_result.stdout
+    return updated_attrs
```
xingao267: How is this used? "bazel sync" will pick this up?

container/pull.bzl
```diff
@@ -31,6 +31,19 @@ def python(repository_ctx):
     fail("rules_docker requires a python interpreter installed. " +
          "Please set BAZEL_PYTHON, or put it on your path.")
 
+_container_pull_attrs = {
+    "registry": attr.string(mandatory = True),
+    "repository": attr.string(mandatory = True),
+    "digest": attr.string(),
+    "tag": attr.string(default = "latest"),
+    "platform": attr.string(default = "linux/amd64"),
```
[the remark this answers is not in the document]
ixdy: I chose platform to match what `docker pull` uses, but I agree that it's confusing in a bazel context.

The problem is that the "platform" is really a tuple of OS and architecture, so "arch" is a bit misleading. Would it be clearer if we split this into two attributes, `os` and `arch`?

container/pull.bzl
ixdy: Yes, that is my understanding. Future versions of bazel add the `bazel sync` command as well as experimental flags which save the results in a specified `.bzl` file, including the return values of any repository rules; other flags would let you tell bazel to read workspace dependencies from this resolved `.bzl` file instead of from `WORKSPACE`. [This blog post](https://blog.bazel.build/2018/07/09/bazel-sync-and-resolved-file.html) gives a better explanation.

xingao267: Cool. Thanks for the explanation and pointer.

nlopezgi:
Thanks for sending this PR. It looks good overall. Lets get the change in containerregistry submitted and then return to this.

container/pull.bzl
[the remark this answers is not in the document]
ixdy: changed to `os` + `arch`.

container/pull.bzl
```diff
@@ -84,20 +103,15 @@ container_import(
...
-            cfg = "host",
-        ),
-    },
-    implementation = _impl,
-)
+    updated_attrs = {
+        k: getattr(repository_ctx.attr, k)
+        for k in _container_pull_attrs.keys()
+    }
+    digest_result = repository_ctx.execute(["cat", repository_ctx.path("image/digest")])
```
nlopezgi: lets avoid using execute in repository rules as much as possible (particularly for things like this) as this will not be platform agnostic

ixdy: we already use `execute` to call the puller, but yeah, I'm a bit unhappy about having to shell out to `cat`.

I couldn't find another way to read a file into a starlark variable, though. If we don't like `cat`, I guess we could write a small python wrapper, though is that better or worse than using `cat`?

nlopezgi: ugh, using a python wrapper properly is too much effort, alright, lets leave as is then. This PR looks good overall then as is. Lets try to get the upstream change. Also, could you try to add a test (we can wait till upstream change is in if you want)

fejta:
testing 1 2 3

fejta:
gubernator confirmed - https://k8s-gubernator.appspot.com/pr/ixdy

fejta:
/unassign @erain 
Prow commands confirmed!

ixdy:
google/containerregistry update now included. this should be good to go.

nlopezgi:
hi @ixdy , do you think you can try to set up tests for this PR?

ixdy:
Yeah, I was starting to work on tests, then ran into https://github.com/google/containerregistry/issues/128, and then I got busy with some other project work. I hope to get back to this by the end of the week, though we'll need a new containerregistry release, too.

ixdy:
er, wrong issue, I meant https://github.com/google/containerregistry/issues/127.

ixdy:
(though https://github.com/google/containerregistry/issues/128 means that windows images aren't supported currently)

nlopezgi:
right, forgot about google/containerregistry#127, sgtm, if you need any help with extra reviews for containerregistry changes let me know. And bummer about google/containerregistry#128, but imo its not a priority to get windows working. (I'm not in a hurry to see this in either, just wanted to check up on status, so take your time)

container/pull.bzl
```diff
@@ -31,6 +31,24 @@ def python(repository_ctx):
...
+    "registry": attr.string(mandatory = True),
+    "repository": attr.string(mandatory = True),
+    "digest": attr.string(),
+    "tag": attr.string(default = "latest"),
+    "os": attr.string(default = "linux"),
+    "os_version": attr.string(),
+    "os_features": attr.string_list(),
+    "architecture": attr.string(default = "amd64"),
+    "cpu_variant": attr.string(),
+    "platform_features": attr.string_list(),
```
ixdy: I'm not particularly happy about the name of this attribute, but `features` is a reserved attribute.

The [docker manifest list spec](https://docs.docker.com/registry/spec/manifest-v2-2/#manifest-list-field-descriptions) refers to this as CPU features, but the [OCI image index spec](https://github.com/opencontainers/image-spec/blob/master/image-index.md) just says it's a reserved field.

ixdy: I haven't found any images that actually use `features` yet.

WORKSPACE
```diff
@@ -43,6 +43,46 @@ container_load(
...
+    architecture = "ppc64le",
+    registry = "index.docker.io",
+    repository = "library/alpine",
+    tag = "3.8",
+)
+
+container_pull(
+    name = "k8s_pause_amd64",
+    registry = "k8s.gcr.io",
+    repository = "pause",
```
nlopezgi: should we set a fixed sha here so test can be guaranteed to only fail due to resolution of the digest and not due to a change in the image being pulled?

WORKSPACE
```diff
@@ -43,6 +43,46 @@ container_load(
...
+    tag = "3.1",
+)
+
+container_pull(
+    name = "k8s_pause_arm64",
+    architecture = "arm64",
+    registry = "k8s.gcr.io",
+    repository = "pause",
```
nlopezgi: same here, add sha?

WORKSPACE
ixdy: I changed this one to a digest pointing at the manifest list to verify that the resolved digest points at the amd64 image.

WORKSPACE
ixdy: I used the kubernetes pause images here because
a) they're small
b) they're unlikely to be updated
c) if we do update them, we have a policy of not overwriting existing tags

I guess I'll have to eat my hat if this changes. :)

nlopezgi: sgtm

WORKSPACE
nlopezgi: thanks

WORKSPACE
nlopezgi: sgtm

PR #549, opened 2018-10-17
Run update-deps.sh

dprotaso:
This includes the commit https://github.com/GoogleContainerTools/distroless/commit/f93e9b5c88a0ac6b98bc4ff94d206702cfb4e7e3
It adds the full text for various OSS licenses

### DIFF
--- a/cc/cc.bzl
+++ b/cc/cc.bzl
@@ -18,8 +18,8 @@
 # git repository.
 
 DIGESTS = {
-    # "gcr.io/distroless/cc:debug" circa 2018-10-15 14:08 +0200
-    "debug": "sha256:d8f6bf07bc352cf691f1de3d17d42282f007e631dbad7c973d2c478167afa302",
-    # "gcr.io/distroless/cc:latest" circa 2018-10-15 14:08 +0200
-    "latest": "sha256:923564f1d33ac659c15edf538b62f716ed436d7cc5f6a9d64460b8affba9ccd9",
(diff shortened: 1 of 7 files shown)


dprotaso:
Curious what's the best _independent_ way to determine what this digest should be https://github.com/bazelbuild/rules_docker/blob/6d9fdf2ca948ba4cfe3da778bb5afae71a3cf8dd/tests/docker/BUILD#L174-L178

It's causing the failure here:
https://source.cloud.google.com/results/invocations/b915ecc0-2769-47bd-a26e-95a9a3b0bf37/targets/%2F%2Ftests%2Fdocker:test_digest_output2/log

nlopezgi:
hmm, @ash2k recently added this test; could you comment on how you determined this value?

ash2k:
@nlopezgi IIRC I built the digest target and also ran the push target and they both produced the same value. Then I took that value and put into the test.

ash2k:
The digest is expected to be different if the base image is different.

dprotaso:
I pushed the necessary changes 

> The digest is expected to be different if the base image is different.

Should the base image for these tests remain static? ie. pin the sha?

xingao267:
We can create a new container_pull target named maybe distroless_base_test and pin to a specific version.

xingao267:
Oh wait. distroless_base is created for testing purpose, so we can just pin this to a specific sha.

nlopezgi:
+1 to using distroless_fixed_id and distroless_fixed_id_2 for these tests

dprotaso:
rebased and re-ran the update deps script

PR #551, opened 2018-10-19
cleanup: use the new Python octal integer literals notation consistently

pawelz:
Python supports two notations for octal integer literals: 0X and 0oX. According to PEP-3127, the former is deprecated and will be dropped in Python 3. Update rules_docker to use the new notation consistently.

https://www.python.org/dev/peps/pep-3127/#grammar-specification

### DIFF
--- a/README.md
+++ b/README.md
@@ -544,7 +544,7 @@ passwd_file(
 pkg_tar(
     name = "passwd_tar",
     srcs = [":passwd"],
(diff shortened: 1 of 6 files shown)


xingao267:
Thanks for the PR. I'll send a PR to fix the test and you can rebase onto it.

pawelz:
Sounds good, thanks!

PR #559, opened 2018-10-24
Use docker toolchain rule in incremental_load

smukherj1:
Add the toolchain_type as a toolchain field to import, bundle
and image that use incremental_load under the hood


tools/BUILD
```diff
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-package(default_visibility = ["//visibility:private"])
+package(default_visibility = ["//visibility:public"])
```
nlopezgi: can we just increase visibility of the toolchain_type_docker target only

smukherj1: Done

toolchains/docker/BUILD
```diff
@@ -0,0 +1,71 @@
...
+toolchain_type(
+    name = "toolchain_type",
+)
+
+# Default docker toolchain that expects the 'docker' executable
+# to be in the PATH
+docker_toolchain(
+    name = "default_toolchain_impl",
+    tool_path = "docker",
+    visibility = ["//visibility:public"],
```
nlopezgi: can we restrict the visibility of this target?

smukherj1: Nope. This is actually the rule that is being accessed by rules that use this toolchain_type. Making this private will give visibility errors

nlopezgi: got it, sgtm

PR #560, opened 2018-10-24
Support stamp and creation_time in go_image

fejta:
ref https://github.com/bazelbuild/rules_docker/issues/155

* Allow `go_image(creation_time="1540414141")` and `go_image(stamp=True)`
* Basic unit tests for `go_image()` that validates the tar file

/assign @nlopezgi

### DIFF
--- a/go/image.bzl
+++ b/go/image.bzl
@@ -62,7 +62,15 @@ DEFAULT_BASE = select({
     "//conditions:default": "@go_image_base//image",
 })
 
-def go_image(name, base = None, deps = [], layers = [], binary = None, **kwargs):
+def go_image(
+    name,
+    base = None,
+    deps = [],
+    layers = [],
+    binary = None,
+    stamp = None,
+    creation_time = None,
+    **kwargs):
     """Constructs a container image wrapping a go_binary target.
 
   Args:
@@ -94,4 +102,6 @@ def go_image(name, base = None, deps = [], layers = [], binary = None, **kwargs)
         tags = tags,
         args = kwargs.get("args"),
         data = kwargs.get("data"),
+        stamp = stamp,
+        creation_time = creation_time,
     )
--- a/tests/go/BUILD.bazel
+++ b/tests/go/BUILD.bazel
@@ -1,0 +1,29 @@
+package(default_visibility = ["//visibility:public"])
+
+load("@io_bazel_rules_go//go:def.bzl", "go_binary")
+load("//go:image.bzl", "go_image")
+
+go_binary(
+    name = "app",
+    srcs = ["main.go"],
+)
+
+go_image(
+    name = "image",
+    binary = ":app",
+    pure = "on",
+    stamp = True,
+)
+
+go_image(
+    name = "image_build_binary",
+    srcs = ["main.go"],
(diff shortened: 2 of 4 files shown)


nlopezgi:
Hi @fejta 
the xx_image rules have attributes that match exactly the corresponding xx_binary rule, so adding these attrs to go_image is not possible.
The preferred solution is either 1. create a base with the stamp/creation time and extend that in a go_image rule or 2. extend a go_image rule with a container_image that adds the stamp/creation time. If none of those two alternatives currently work then we should change the code to make them work (w/o having to add the property to go_image directly)

fejta:
This feedback doesn't make sense. 

`creation_time` and `stamp` are not valid arguments to send to `go_binary()` and we do not send them there, just like we do not send `base`, `layers` and `binary` to the `go_binary()` target.

These attrs are already present in `go_image()`:
* `go_image()` is a macro for [calling](https://github.com/bazelbuild/rules_docker/blob/f43915a64a46c8cdd30e9ee7155f400972d03cab/go/image.bzl#L97) `app_layer()`
* `app_layer()` is a rule that [includes](https://github.com/bazelbuild/rules_docker/blob/f43915a64a46c8cdd30e9ee7155f400972d03cab/lang/image.bzl#L202) all the attributes of `_container.image.attrs.items()`
* `container.image` [equals](https://github.com/bazelbuild/rules_docker/blob/f43915a64a46c8cdd30e9ee7155f400972d03cab/container/container.bzl#L31) `container_image`
* `container_image` [includes](https://github.com/bazelbuild/rules_docker/blob/f43915a64a46c8cdd30e9ee7155f400972d03cab/container/image.bzl#L464) the `stamp` and `creation_time` attributes

So all the work is done here, we just need to ensure we do not send the irrelevant `stamp` argument to `go_binary()` by moving it out of `**kwargs` in `go_image()`

fejta:
The reason `xx_image` and `xx_binary` have matching attributes is because the `xx_image()` macro calls `xx_binary(name = name.binary, deps = deps + layers, **kwargs)` but if we pull `stamp` out of `kwargs` (like we do for `layers`) then there's no problem

nlopezgi:
Sorry I was not clear,
the reason why xx_image and xx_binary have the same attr is that _by design_ we want them to have the same api. That is, we want for users who have a xx_binary rule to be able to translate to xx_image and back with a minimum overhead (in terms of having to add or remove attrs when replacing one for the other). 
So its not at all for technical reasons that we don't have attributes like creation time and stamp in xx_image rules. 
That is, I understand that its technically possible to add this attribute to go_image and that it does not make sense to add it to go_binary, but we don't want the APIs to diverge.

My suggestion is that for all requests like this one (and we do often get requests to add more attributes to xx_image rules) is that we need to make the functionality in container_image work in such a way that using a container_image target either as base or overlaying it on top of an xx_image rule allows for the properties (in this case the creation time / stamp) to be passed through (either from the base up or from the top layer down) to the xx_image rule. Please let me know if this makes more sense otherwise I can try to explain it with a more detailed example.

fejta:
Makes sense, thanks!

PR #566, opened 2018-11-06
Add basic RPM support

### LINKED ISSUE #203: RFE support building container images with rpms
for the `container_image` (or `docker_build`) there is currently no support for rpm based distributions. This is a blocker for a number of use-cases.

rmohr:
Hi,

I tried to implement installing RPMS inside the containers similar to debs.
It would be great to get some initial feedback.

An example rule, which is also used for testing in the PR looks like this:

```
container_image(
    name = "tar_with_rpm",
    base = ":tar_base",
    rpms = ["iptables-1.6.2-3.fc27.i686.rpm"],
)
```

When reading through the comments in #203 I guess I would have to improve the following things:

 * don't use `cpio` and `rpm2cpio`, or would that be ok?
 * Create an rpm via python, or use a smaller rpm for tests

If anyone has an idea on how to best proceed with the mentioned points, I would be happy to hear that.

Fixes #203.


container/build_tar.py
```diff
@@ -284,6 +290,34 @@ def add_pkg_metadata(self, metadata_tar, deb):
...
+    be added to the output.
+
+    Args:
+      rpm: the tar file to add
+
+    Raises:
+      RPMError: if the format of the rpm archive is incorrect.
+    """
+    try:
+      p1 = subprocess.Popen(["rpm2cpio", rpm], stdout=subprocess.PIPE)
```
cynepco3hahue: I do not sure if it good to assume that we had `rpm2cpio` by default.

container/layer.bzl
```diff
@@ -149,6 +153,7 @@ def _impl(
     env: str Dict, overrides ctx.attr.env
     operating_system: operating system to target (e.g. linux, windows)
     debs: File list, overrides ctx.files.debs
+    rpms: File list, overrides ctx.files.debs
```
cynepco3hahue: `s/overrides ctx.files.debs/overrides ctx.files.rpms/`

container/build_tar.py
```diff
@@ -284,6 +290,34 @@ def add_pkg_metadata(self, metadata_tar, deb):
...
 
+  def add_rpm(self, rpm):
+    """Extract an rpm package in the output tar.
+
+    All files presents in that rpm package will be added to the
+    output tar under the same paths. No user name nor group names will
+    be added to the output.
+
+    Args:
+      rpm: the tar file to add
```
cynepco3hahue: I think more correct to say `the rpm package to add`

testdata/BUILD
```diff
@@ -104,6 +104,12 @@ container_image(
     tars = ["one.tar"],
 )
 
+container_image(
+    name = "tar_with_rpm",
+    base = ":tar_base",
+    rpms = ["iptables-1.6.2-3.fc27.i686.rpm"],
```
cynepco3hahue: IIUC I firstly will need to download `rpm` under the directory with the bazel target, does it correct?

cynepco3hahue:
> thanks for your contribution. I will review your PR in detail later this week On quick inspection it seems your PR adds functionality to transform an rpm package to a tar file and then adds the tar file to the container image. If that is the case, then perhaps this does not need to be upstreamed in this repo, and you can just rely on a standalone rule to transform your packages to tars and then put that in the container_image? let me know what you think.

I tend to agree, but in this case, `deb` variables also must be dropped for consistency.

rmohr:
@nlopezgi thank

> On quick inspection it seems your PR adds functionality to transform an rpm package to a tar file and then adds the tar file to the container image.

Yes that's exactly the current state.

> and you can just rely on a standalone rule to transform your packages to tars and then put that in the container_image? let me know what you think.

As it is right now, definitely. Here a few additional thoughts which I had:
 * I wanted it to be on an equal level like `deb` and make it equally easy to use.
 * I was also thinking about making sure that a query-able rpm database is inside the container, so that e.g. [container-diff](https://github.com/GoogleContainerTools/container-diff) would work. There it would be beneficial to have full access to the container, so that rpm database can be updated and that all this happens automatically without extra bazel configuration or rules.

What do you think?

testdata/BUILD
rmohr: Yes. Correct. An example download rule would look like this:

```
http_file(
   name = "glibc",
   url = "https://dl.fedoraproject.org/pub/fedora/linux/releases/28/Everything/x86_64/os/Packages/g/glibc-2.27-8.fc28.x86_64.rpm",
   sha256 = "573ceb6ad74b919b06bddd7684a29ef75bc9f3741e067fac1414e05c0087d0b6"
)
```

and it can then be referenced like this:

```
container_image(
    name = "glibc_only_fedora",
    rpms = ["@glibc//file"],
)
```

container/build_tar.py
rmohr: Yes. I think that we would need a pure python solution. My main issue was that I could find all the needed pieces written in pure python, but not in a reasonable shape or one place, so that I could just rely on one or more pip dependencies.

nlopezgi:
debs will likely be removed in the short term as it does not properly install arbitrary packages. We're not interested in providing more functionality in this direction. I suggest you write separate rules to deal with rpm packages that either produce outputs that can be consumed by container_image or that extend these rules to provide the functionality you need.

rmohr:
@nlopezgi understood. thank you for looking into this.

container/build_tar.py
vbatts: pure-python is a bit of a wormhole i think. and if there is `rpm` package installed, it provides `rpm2cpio`

PR #567, opened 2018-11-06
Merge toolchain rules into one

smukherj1:
Can't use same impl function in multiple rules until
https://github.com/bazelbuild/bazel/issues/6613 is fixed


smukherj1:
Unnecessary as the issue has been resolved and the fix will be included in the next bazel release. https://github.com/bazelbuild/bazel/issues/6613 has a workaround

PR #573, opened 2018-11-08
Remove exec constraint for docker toolchain

smukherj1:
The docker toolchain is only meant to be used in the target platform.
Thus the host environment is irrelevant and the registered toolchains
shouldn't include constraints from those

See https://github.com/bazelbuild/rules_k8s/issues/215 which lead to the discovery of this issue. Specifically https://github.com/bazelbuild/rules_k8s/issues/215#issuecomment-437059049 explains why.

https://github.com/bazelbuild/rules_k8s/pull/218 makes the same fix for rules_k8s


smukherj1:
Seems like the mac machines are having some issues downloading external artifacts
`
ERROR: /private/var/tmp/_bazel_buildkite/0976794005e2791285a96758be9dc21d/external/rust_default_toolchains/BUILD.bazel:43:1: no such package '@rust_darwin_x86_64//': Error downloading [https://static.rust-lang.org/dist/rust-1.24.1-x86_64-apple-darwin.tar.gz] to /private/var/tmp/_bazel_buildkite/0976794005e2791285a96758be9dc21d/external/rust_darwin_x86_64/rust-1.24.1-x86_64-apple-darwin.tar.gz: Broken pipe (Write failed) and referenced by '@rust_default_toolchains//:rust-darwin-x86_64_impl'
`

Issue #555, opened 2018-10-23
container image build fails when env variable contains comma


Weiss91:
I encountered a bug where the container build fails if an environment variable declared in the `env` parameter contains a comma.
E.g.:
```
container_image(
    name = "image_test_bug",
    base = ":layer_test_bug",
    cmd = ["run"],
    env = {
        "BUGGY_ENV_VAR": "This does not work with a comma ,",
    },
    ports = ["12345"],
    stamp = True,
    visibility = ["//visibility:public"],
)
```

Minimal working example: 
https://github.com/Weiss91/bazelDockerRulesCommaBug

To see the bug, run `bazel build image_test_bug`

output: 
```$ bazel build image_test_bug
INFO: Analysed target //:image_test_bug (1 packages loaded).
INFO: Found 1 target...
ERROR: /mnt/c/Users/matthias.weiss/go/src/github.com/weiss91/bazelDockerEnvBug/BUILD.bazel:21:1: ImageConfig image_test_bug.0.config failed (Exit 1)
Traceback (most recent call last):
  File "/home/matthias/.cache/bazel/_bazel_matthias/876a68a91054abbeffa7e92ef6acf28e/sandbox/processwrapper-sandbox/1/execroot/test_env/bazel-out/host/bin/external/io_bazel_rules_docker/container/create_image_config.runfiles/io_bazel_rules_docker/container/create_image_config.py", line 208, in <module>
    main()
  File "/home/matthias/.cache/bazel/_bazel_matthias/876a68a91054abbeffa7e92ef6acf28e/sandbox/processwrapper-sandbox/1/execroot/test_env/bazel-out/host/bin/external/io_bazel_rules_docker/container/create_image_config.runfiles/io_bazel_rules_docker/container/create_image_config.py", line 184, in main
    for (k, v) in six.iteritems(KeyValueToDict(args.env))
  File "/home/matthias/.cache/bazel/_bazel_matthias/876a68a91054abbeffa7e92ef6acf28e/sandbox/processwrapper-sandbox/1/execroot/test_env/bazel-out/host/bin/external/io_bazel_rules_docker/container/create_image_config.runfiles/io_bazel_rules_docker/container/create_image_config.py", line 95, in KeyValueToDict
    (k, v) = kv.split('=', 1)
ValueError: need more than 1 value to unpack
Target //:image_test_bug failed to build
Use --verbose_failures to see the command lines of failed build steps.
[...]

PR #577, opened 2018-11-09
added support for commas in env and labels #555

alexbumbacea:


alexbumbacea:
I signed it!

alexbumbacea:
hi @nlopezgi,
Added a test case for csv in env

tests/docker/BUILD
```diff
@@ -176,3 +176,18 @@ file_test(
...
+container_image(
+    name = "set_env_csv",
+    base = "@distroless_fixed_id//image",
+    cmd = ["foo"],
+)
+
+container_test(
+    name = "set_env_csv_test",
+    configs = ["//tests/docker/configs:set_env_csv.yaml"],
+    env = {
```
nlopezgi: container test does not allow env attr, it only receives the configs to validate with

nlopezgi: (the env should be set in the container_image above)

nlopezgi:
also //container:image_test is failing, you should be able to run that locally and debug

nlopezgi:
ugh, mac builds are timing out pulling images, will have to figure out tomorrow what the problem is

PR #578, opened 2018-11-09
Move more java image tests to container image

nlopezgi:
* moving more tests to use container image
* refactoring tests to java directory
* adding env variables check for all java image tests

### DIFF
--- a/tests/docker/BUILD
+++ b/tests/docker/BUILD
@@ -19,7 +19,6 @@ load(
     "@bazel_tools//tools/build_rules:test_rules.bzl",
     "file_test",
 )
-load("//java:image.bzl", "java_image")
 
 container_test(
     name = "structure_test",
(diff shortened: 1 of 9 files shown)


container/image_test.py
```diff
@@ -706,18 +706,6 @@ def test_java_runfiles_image(self):
...
-          '/app/io_bazel_rules_docker/../bazel_tools/tools/java/runfiles/librunfiles.jar',
-          '/app/io_bazel_rules_docker/testdata/java_runfiles_image.binary.jar',
-          '/app/io_bazel_rules_docker/testdata/java_runfiles_image.binary',
-          '/app/io_bazel_rules_docker/testdata/foo',
-        ]),
-        'examples.images.Runfiles',
-      ])
-
-      
   def test_java_runfiles_as_lib_image(self):
```
xingao267: can we change this to also use images in tests/docker/java/BUILD, and remove those targets in testdata/BUILD?

nlopezgi: will be doing that in an upcoming PR, I want to migrate things progressively to start getting everything out of testdata/BUILD that is a big mess

xingao267: sg.

PR #579, opened 2018-11-09
migrating more tests to tests/docker/java using container_test

nlopezgi:


testing/e2e.sh
```diff
@@ -315,18 +315,6 @@ function test_war_image_with_custom_run_flags() {
   EXPECT_CONTAINS "$(cat bazel-bin/testdata/war_image_with_custom_run_flags)" "-i --rm --network=host -e ABC=ABC"
 }
 
-function test_java_runfiles_image() {
-  cd "${ROOT}"
-  clear_docker
-  EXPECT_CONTAINS "$(bazel run "$@" testdata:java_runfiles_image)" "asdf"
```
xingao267: are we still testing this?

nlopezgi: its overkill, we have sufficient tests that run a simple java image and check its expected output, this is not adding any more value

PR #584, opened 2018-11-26
Flip legacy_run_behavior default value

Globegitter:
I think it is about time to flip the default - it amounts to quite a better user experience.


Globegitter:
As far as I could see this was actually an undocumented attribute - but yes probably makes sense to update the docs anway - will do.

xingao267:
The attribute was added in https://github.com/bazelbuild/rules_docker/pull/112. I think the default should still be kept false. But the naming can probably be changed to be more clear. If the image doesn't have any entrypoint or cmd set, the `bazel run` (which triggers the `docker run`) will fail. Note that the default value of this for the lang_images are set to False as those images always have the entrypoint and cmd set properly.

PR #585, opened 2018-11-26
Fix not contains in end-to-end tests

mjduijn:
The function NOT CONTAINS would output success if any line in the output did not contain substring.
The new version outputs success if no line in the output contains substring.

Not sure if this was intentional, naming seems to suggest it wasn't.


mjduijn:
Yes my mistake, always seem to mess up bash booleans

PR #587, opened 2018-11-27
remove unnecessary imports

nlopezgi:


smukherj1:
Actually ignore my last comment. I realized unused imports are still available as symbols that import the module with the unused imports. This is used by rules_go and other repos to export all their rules from a single "defs.bzl" file. I wonder if this every symbol is exported by default is a good idea in starlark in general

PR #591, opened 2018-12-02
Add 'digest' field to PushInfo provider.

ceason:
This PR adds a 'digest' field to PushInfo provider. The field's value is a `<File>` containing the to-be-published image digest.

I rearranged the `container_pusher` implementation function a bit, to emphasize that the digester needs to receive the same arguments as the pusher.

PR #595, opened 2018-12-03
add test for container_push

nlopezgi:
Add a simple test for container_push by starting up a local registry and using that to push the image. We are using a hardcoded port, not sure if that will work on travis. Lets let it run...

fyi @ceason this PR should enable you to test https://github.com/bazelbuild/rules_docker/pull/591 easily

### DIFF
--- a/testing/e2e.sh
+++ b/testing/e2e.sh
@@ -372,7 +372,15 @@ function test_d_image() {
 function test_nodejs_image() {
   cd "${ROOT}"
   clear_docker
-  EXPECT_CONTAINS "$(bazel run "$@" testdata:nodejs_image)" "Hello World!"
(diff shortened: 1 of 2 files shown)


testing/e2e.sh
```diff
@@ -372,7 +372,15 @@ function test_d_image() {
...
   cd "${ROOT}"
   clear_docker
-  EXPECT_CONTAINS "$(bazel run "$@" testdata:nodejs_image)" "Hello World!"
+  EXPECT_CONTAINS "$(bazel run testdata:nodejs_image)" "Hello World!"
+}
+
+function test_container_push() {
+  cd "${ROOT}"
+  clear_docker
+  run -d -p 5000:5000 --name registry registry:2
```
smukherj1: Doesn't this need to be "docker run"?

nlopezgi: fixed

Issue #598, opened 2018-12-04
Issue in README


chrislovecnm:
```
Alternatively, you can build a docker load compatible bundle with: bazel build my/image:helloworld.tar. This will produce the file: bazel-genfiles/my/image/helloworld.tar, which you can load into your local Docker client by running: docker load -i bazel-genfiles/my/image/helloworld.tar. Building this target can be expensive for large images.
```

This does not seem to be correct

tars seem to be landing in `bazel-bin`

Issue #597, opened 2018-12-04
README is ahead of release


chrislovecnm:
The readme lists using

```
http_archive(
    name = "io_bazel_rules_docker",
    sha256 = "29d109605e0d6f9c892584f07275b8c9260803bf0c6fcb7de2623b2bedc910bd",
    strip_prefix = "rules_docker-0.5.1",
    urls = ["https://github.com/bazelbuild/rules_docker/archive/v0.5.1.tar.gz"],
)

# Call this to override the default docker toolchain configuration. This call
# should be placed BEFORE the call to "container_repositories" below to actually 
# override the default toolchain configuration
load("@io_bazel_rules_docker//toolchains/docker:toolchain.bzl",
    docker_toolchain_configure="toolchain_configure"
)
docker_toolchain_configure(name = "docker_config")
```

But this is incorrect.  Version v0.5.1.tar.gz does not have the toolchain in it yet.

```
http_archive(
    name = "io_bazel_rules_docker",
    sha256 = "29d109605e0d6f9c892584f07275b8c9260803bf0c6fcb7de2623b2bedc910bd",
    strip_prefix = "rules_docker-0.5.1",
    urls = ["https://github.com/bazelbuild/rules_docker/archive/v0.5.1.tar.gz"],
)

load(
    "@io_bazel_rules_docker//container:container.bzl",
    "container_pull",
    container_repositories = "repositories",
)

container_repositories()
```

I believe is correct

/cc @nlopezgi

nlopezgi:
the latest release does not include the toolchain changes, please look at the state of README at the release for instructions for that release (https://github.com/bazelbuild/rules_docker/tree/v0.5.1)

Issue #317, opened 2018-02-09
do not flatten file paths in container_image


jmhodges:
See https://github.com/bazelbuild/bazel/issues/2176

This is still happening and has now gotten weirder: https://github.com/bazelbuild/bazel/issues/2176#issuecomment-364352176

It would be good for container_image's `files` to not just be the basename files including in the directory, but to, by default, maintain the full relative file paths.

pcj:
I find this troubling also.  Is this a bug or a feature?

pcj:
Friendly ping.  This is a problem for adoption of rules_docker.

rutsky:
Setting `strip_prefix` in tar_pkg workaround helped in my case (not to `.`, but to specific directory, as in my layout I needed to strip some prefix):

```
pkg_tar(
    name = "webfrontend-tar",
    srcs = ["//webfrontend:build"],
    mode = "0o644",
    package_dir = "/usr/share/nginx/html/",
    # Otherwise all directories are flattened:
    # <https://github.com/bazelbuild/rules_docker/issues/317>
    strip_prefix = "webfrontend/build/",
)

container_image(
    name = "webfrontend-image",
    tars = [":webfrontend-tar"],
    base = "@nginx_base//image",
    visibility = ["//visibility:public"],
)
```

Without `strip_prefix = "webfrontend/build/",` all files are flattened in `webfrontend-tar`.

nlopezgi:
is this still an issue? please reopen if so.

jmhodges:
It is! I can’t reopen this issue, though

nlopezgi:
Is the workaround you found using pkg_tar noy working or otherwise problematic?

jmhodges:
It’s problematic because it’s undocumented and unexpected. It’s really hard to understand why this happens or how to fix it without lots of googling. It’s another weird thing that you have to tell people about when they just want to try out Bazel (and there’s plenty of them, you know?)

nlopezgi:
I have updated the docs (See PR updating readme) to explain this issue. I'm closing this again as I dont think the fix for this issue is on this repo  (i.e., it seems to me the issue is with pkg_tar rule). Please let me know if anyone feels there is more we can do here to reopen.

njlr:
I find it very strange that an extra rule is required to place files in specific folders in  a `container_image`. It seems like an extremely noisy API to me. 

Perhaps something like this should be possible?

```python
container_image(
  name = "my_image",
  base = ":nodejs",
[...]

PR #599, opened 2018-12-04
update readme

nlopezgi:
* fix https://github.com/bazelbuild/rules_docker/issues/597
* fix https://github.com/bazelbuild/rules_docker/issues/598
* provide docs for recommended fix for https://github.com/bazelbuild/rules_docker/issues/317
* Improve documentation for xx_image rules to make more explicit that args such as env and symlinks will never be added to xx_image and container_image must be used in conjunction with xx_image for these use cases.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ https://docs.bazel.build/versions/master/be/java.html#java_library))
 * [scala_image](#scala_image) ([signature](
 https://github.com/bazelbuild/rules_scala#scala_binary))
 * [groovy_image](#groovy_image) ([signature](
-https://github.com/bazelbuild/rules_groovy#groovy_binary))
+https://github.com/bazelbuok, not sure if you want to publish ild/rules_groovy#groovy_binary))
 * [cc_image](#cc_image) ([signature](
 https://docs.bazel.build/versions/master/be/c-cpp.html#cc_binary))
 * [go_image](#go_image) ([signature](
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -45,7 +45,7 @@ https://docs.bazel.build/versions/master/be/python.html#py_binary))
 https://docs.bazel.build/versions/master/be/python.html#py_binary))
 * [nodejs_image](#nodejs_image) ([usage](
 https://github.com/bazelbuild/rules_nodejs#usage))
-* [java_image](#java_image) ([signature](
+* [java_image](#java_image) ([signature](see <a href=#go_image-custom-base>go_image (custom base)</a> example.
```
xingao267: copy/paste error?

README.md
```diff
@@ -118,6 +134,15 @@ container_pull(
...
 ```
 
+Note: Bazel does not deal well with diamond dependencies. If the repositories that
+are imported by `container_repositories()` have already been imported (at a
+different version) by other rules you called in your `WORKSPACE`, which are
+placed above the call to `container_repositories()`, arbitrary errors might
+ocurr. If you get errors related to external repositories, you will likely
+not be able to use `container_repositories()` and will have to import
+directly in your `WORKSPACE` all the required dependencies (see the most up
+to date impl of `container_repositories()` for details).
```
xingao267: This is happening? container_repositories should check for if the repo has been loaded or not.

Issue #601, opened 2018-12-07
rules_docker broken on Bazel's downstream projects


buchgr:
rules_docker tests don't work with Bazel HEAD, probably because --incompatible_strict_action_env has been flipped to true. Some tests fail because the `bazel` command is not on the PATH.
```
+++ bazel test --test_output=all //test_folder:test_for_failure
++ out='/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/5b031cc74a6f71160756712e3d8d076d/sandbox/linux-sandbox/568/execroot/io_bazel_rules_docker/bazel-out/k8-fastbuild/bin/tests/contrib/test_compare_ids_test_diff_ids_fails.runfiles/io_bazel_rules_docker/tests/contrib/test_compare_ids_test_diff_ids_fails: line 73: bazel: command not found'
```

See https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/676#f72520bc-0aa7-44d6-bdbf-ceabb76715d1

We have disabled the project in our downstream testing until this issue is resolved. Thank you very much!

meteorcloudy:
/cc @xingao267 @nlopezgi
Can you add `--action_env=PATH` in the bazelci config file?

nlopezgi:
@buchgr this should be fixed, can you please re-enable in upstream and let me know if you see issues.

meteorcloudy:
https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/1#_
@nlopezgi Looks like it's still failing on macos, can you look into it?

mhlopko:
Looking at https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/8 looks like this is properly fixed. Will reenable rules_docker now. Thanks!

PR #602, opened 2018-12-07
adding --action_env=PATH to fix tests/contrib/test_compare_ids_test_* tests

nlopezgi:
fixes https://github.com/bazelbuild/rules_docker/issues/601


.bazelci/presubmit.yml
```diff
@@ -34,7 +46,7 @@ platforms:
     test_targets:
     - "--"
     - "//tests/..."
-    - "-//tests/contrib:test_compare_ids_test_diff_ids_fails"
+    - "-//tests/contrib:test_compare_ids_test_*"
```
smukherj1: Do we need the following lines after changing this pattern to a wildcard?

- "-//tests/contrib:test_compare_ids_test_invalid_tar_fails"
- "-//tests/contrib:test_compare_ids_test_invalid_tar_fails_multi_regex"
 - "-//tests/contrib:test_compare_ids_test_no_images_with_id_fails"

nlopezgi: ugh, this was a blanket replace error

PR #605, opened 2018-12-08
Automatically infer stamping for container_{push,bundle}

ceason:
This PR deprecates the `stamp` attr for the `container_push` and `container_bundle` rules. Given that the stamp format (ie python format syntax) is mutually exclusive with a valid image name (per the [image specification](https://github.com/moby/moby/blob/master/image/spec/v1.md) `{` and `}` are not valid characters), I believe we can safely infer a user's intent to stamp without a separate 'stamp' attr.

### DIFF
--- a/README.md
+++ b/README.md
@@ -200,7 +200,7 @@ at present for doing this.
 
 ### Stamping
 
-The first option is to use `stamp = True`.
+The first option is to use stamping.
 
 ```python
 # A common pattern when users want to avoid trampling
@@ -214,10 +214,6 @@ container_push(
   registry = "gcr.io",
   repository = "my-project/my-image",
   tag = "{BUILD_USER}",
-  creation_time = "{BUILD_TIMESTAMP}",
-
-  # Trigger stamping.
-  stamp = True,
 )
 ```
 
--- a/README.md
+++ b/README.md
@@ -200,7 +200,8 @@ at present for doing this.
 
 ### Stamping
 
-The first option is to use stamping.
+The first option is to use stamping. Stamping is enabled when a supported
+attribute contains a python format placeholder (eg `{BUILD_USER}`).
 
 ```python
 # A common pattern when users want to avoid trampling
--- a/container/bundle.bzl
+++ b/container/bundle.bzl
@@ -43,6 +43,7 @@ def _container_bundle_impl(ctx):
     for unresolved_tag in ctx.attr.images:
         # Allow users to put make variables into the tag name.
         tag = ctx.expand_make_variables("images", unresolved_tag, {})
+
         # If any tag contains python format syntax (which is how users
         # configure stamping), we enable stamping.
         if "{" in tag:
--- a/container/bundle.bzl
+++ b/container/bundle.bzl
@@ -37,9 +37,14 @@ def _container_bundle_impl(ctx):
 
     images = {}
     runfiles = []
+    if ctx.attr.stamp:
+        print("Attr 'stamp' is deprecated; it is now automatically inferred. Please remove it from %s" % ctx.label)
+    stamp = False
     for unresolved_tag in ctx.attr.images:
         # Allow users to put make variables into the tag name.
         tag = ctx.expand_make_variables("images", unresolved_tag, {})
+        if "{" in tag:
+            stamp = True
 
         target = ctx.attr.images[unresolved_tag]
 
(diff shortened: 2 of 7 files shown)


container/bundle.bzl
```diff
@@ -37,9 +37,14 @@ def _container_bundle_impl(ctx):
 
     images = {}
     runfiles = []
+    if ctx.attr.stamp:
+        print("Attr 'stamp' is deprecated; it is now automatically inferred. Please remove it from %s" % ctx.label)
+    stamp = False
     for unresolved_tag in ctx.attr.images:
         # Allow users to put make variables into the tag name.
         tag = ctx.expand_make_variables("images", unresolved_tag, {})
+        if "{" in tag:
```
nlopezgi: can you please add docstring indicating in detail why we use a check for "{" to infer stamping (dont want that knowledge to be lost)

README.md
```diff
@@ -200,7 +200,7 @@ at present for doing this.
 
 ### Stamping
 
-The first option is to use `stamp = True`.
+The first option is to use stamping.
```
nlopezgi: can you update these docs to indicate explicitly how the new inferred stamping works

ceason:
I've updated the PR to include better/more explicit documentation on how stamping is inferred.

README.md
```diff
@@ -200,7 +200,8 @@ at present for doing this.
 
 ### Stamping
 
-The first option is to use `stamp = True`.
+The first option is to use stamping. Stamping is enabled when a supported
+attribute contains a python format placeholder (eg `{BUILD_USER}`).
```
nlopezgi: iiuc things like {BUILD_USER} and {BUILD_TIMESTAMP} get replaced by Bazel before the image name is fully resolved, so I'm not quite sure the assumption that "Given that the stamp format (ie python format syntax) is mutually exclusive with a valid image name (per the image specification `{` and `}` are not valid characters)" applies when users are passing here things like {BUILD_USER}. Am I not understanding correctly your assumption or am I misunderstanding something else?

ceason: Bazel provides stamp variables to rules via the `ctx.info_file` (which includes `BUILD_USER`) and `ctx.version_file` (which includes `BUILD_TIMESTAMP`) files (`bazel-out/stable-status.txt` and `bazel-out/volatile-status.txt` respectively). Users can add additional variables via the `--workspace_status_command` flag.

It's up to individual rules, however, whether and how to incorporate these variables into the built artifact. `rules_docker` uses `@containerregistry//:pusher` for pushing, [which uses `str.format()`](https://github.com/google/containerregistry/blob/master/tools/fast_pusher_.py#L93-L105) to substitute these stamp variables into the provided tag name.

nlopezgi: I do think that its a stretch though to say that because someone is using `{ }` they want to use stamping.

What benefit do you see in inferring this automatically?

ceason: I guess my reasoning was that because `{ }` are invalid characters for a docker image tag:
```
% docker build -t {BUILD_USER}:latest - <<<"FROM busybox"
invalid argument "{BUILD_USER}:latest" for "-t, --tag" flag: invalid reference format: repository name must be lowercase

% docker build -t {build_user}:latest - <<<"FROM busybox"
invalid argument "{build_user}:latest" for "-t, --tag" flag: invalid reference format

% docker build -t ceason:{BUILD_TIMESTAMP} - <<<"FROM busybox"
invalid argument "ceason:{BUILD_TIMESTAMP}" for "-t, --tag" flag: invalid reference format

% docker build -t ceason:latest - <<<"FROM busybox"
Sending build context to Docker daemon  2.048kB
Step 1/1 : FROM busybox
 ---> 59788edf1f3e
Successfully built 59788edf1f3e
Successfully tagged ceason:latest
```
..then their only possible purpose is for stamping.  `rules_k8s` [makes the same assumption](https://github.com/bazelbuild/rules_k8s/blob/master/k8s/object.bzl#L94) (and is actually where I got the idea). When I started using `container_push` (after using `k8s_object` first, which infers stamping) I was confused why I needed to set an additional setting to get stamping to work. 

Put another way, I think the benefit here is better adherence to the "principle of least surprise" (as the only reason for `{` to be in any of these fields is for stamping) and a more ergonomic user interface (one less attr for the user to think about).

Does that explanation make sense?

ceason:
There appear to be. When I remove `stamp = True` from [`testdata/BUILD`](https://github.com/bazelbuild/rules_docker/pull/605/files#diff-b6691c5b532994f9a2e981bc9305720a) all tests still pass.

PR #606, opened 2018-12-08
Add 'launcher' attr to container_image & lang_image

ceason:
This PR adds two attributes to `container_image` and each `lang_image` rule:
- `launcher`: Optional. If present, prefix the image's ENTRYPOINT with this file.
- `launcher_args`: Optional arguments for the launcher attribute. Only valid when launcher is specified.

### Motivation
The lang_image rules make containerizing *_binary rules very easy, which is great. However, when deploying to some container orchestrators (eg ECS), a common practice is to use a "launcher" like [chamber](https://github.com/segmentio/chamber#exec) to securely provision secrets and configuration into the environment (because ECS does not have a mechanism to securely inject secrets or config files). The launcher then `exec`s the actual application and gets out of the way.

The current lang_image rules do not play nice with this pattern. It's certainly possible to use container_image rules and manually construct an image, but then we'd be giving up all the smooth user experience of the lang_image rules.

This is my take at enabling the use of a launcher with the lang_image rules; does it seem reasonable? (If it does, I'd like to think through what a good test would look like and get this thing merged!)

### DIFF
--- a/README.md
+++ b/README.md
@@ -920,6 +920,10 @@ and see <a href=#go_image-custom-base>go_image (custom base)</a> example.
 > attribute.  If specified, they will be appended directly after the
 > container ENTRYPOINT binary name.
 
+> NOTE: all application image rules support the `launcher` label and
+> `launcher_args` string_list attributes. If specified, they will be
+> prepended to the container ENTRYPOINT.
+
 ### container_bundle
 
 ```python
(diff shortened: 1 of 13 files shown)


nlopezgi:
hi @ceason 
In principle, the `xx_image` rules should not expose any not expose any docker related attributes (see readme). This is because we want to maintain the API of `xx_binary` as close as possible as `xx_image`. 

Lets first discuss if/how to add the `launcher` / `launcher_args` to `container_image` (and ideally get that in in a separate PR), and then lets try to figure out a way to make these available to users of `xx_image` w/o exposing them directly in `xx_image` (I've been thinking about how this could be achieved, but also want to hear your thoughts).

ceason:
I'm happy to discuss `launcher`/`launcher_args` for `container_image` in a separate PR if necessary. In fact, that should be enough to make these rules available to `xx_image` users (albeit with the additional step of creating an extra `container_image(..., base=":xx_image", launcher="...")` target)

I do think having the attributes directly available on the `xx_image` rules makes for a nicer user experience, though. Doing so still feels consistent with the top level readme too, because:
> The idea behind these rules is to make containerizing an application built via a `lang_binary` rule as simple as changing it to `lang_image`.

would not change, plus all `xx_image` rules are already have additional attributes (`{cc,go,rust,d}_image` all have the `binary` attr and `{py,py3,java,scala,war,nodejs}_image` all have the `layers` attr).

nlopezgi:
Hi @ceason to help you get this moving, lets just do the PR with changes to container_image and once that is (properly tested) and in we can talk about xx_image.

ceason:
I've created #616 which has just the changes to `container_image`, along with a test.

nlopezgi:
closing as we now have this supported in `container_image` and we're not looking to support it in `lang_image`.

PR #624, opened 2018-12-17
Properly set `cmd` from `args` in lang images.

### LINKED ISSUE #618: Regression in nodejs_image
a nodejs_image target without args set gets a default cmd in docker container produced is set to:
```
                "/bin/sh",
                "-c",
                "/bin/bash"
```
if args are set to empty string
`args = [""],`
cmd is set to empty string (as is expected when no args are passed)

xingao267:
Fixes #618.

### DIFF
--- a/lang/image.bzl
+++ b/lang/image.bzl
@@ -196,6 +196,9 @@ def app_layer_impl(ctx, runfiles = None, emptyfiles = None):
         # we should use the "exec" (list) form of entrypoint.
         entrypoint = entrypoint,
         cmd = args,
+        # If `args` is set to [], None or not set, Docker config will have
+        # a null `Cmd` value.
+        null_cmd = args == [],
     )
 
 _app_layer = rule(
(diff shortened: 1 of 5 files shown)


clintharrison:
This seems to conflict with the recent change made to `args` and whether they are passed in the `docker run` command in #586. Since we are now setting `args` and _not_ `cmd`, this means we can revert the change to pass `$@` in the run statements put in the incremental load script without duplicating them, right?

xingao267:
I'm trying to understand this and also crossing referencing issue
https://github.com/bazelbuild/rules_docker/issues/637.

For lang_images, setting `args` is setting the `cmd` of the resulting
image. #624 <https://github.com/bazelbuild/rules_docker/pull/624> fixes the
behavior that if `args` is not specified, then the cmd from the base image
will be used in the current image. The expected behavior should be, if I
don't specified any args for a lang_image, then the cmd should be set to
Null. This means the args/cmd is directly set in the image config, and #586
<https://github.com/bazelbuild/rules_docker/pull/586>, which removes the
duplicated args in `docker run` statement, is needed and works as intended.

For java_images, since args are appended to the entrypoint of the image, so
#586 <https://github.com/bazelbuild/rules_docker/pull/586> is also needed
so that the args are not specified again in the `docker run` statement.

Maybe the issue here is just the inconsistency of how args are treated for
java_images and other lang_images as described in
https://github.com/bazelbuild/rules_docker/issues/637?

PR #630, opened 2018-12-20
Include custom runfile symlinks in layers

scele:
Include any symlinks in *_image layers that were added to dependent
runfiles with ctx.runfiles(symlinks=...).

The new feature only works for bazel >= 0.21.0 which has
https://github.com/bazelbuild/bazel/issues/3880 fixed, and for older
versions such symlinks will silently be dropped (like today).

### DIFF
--- a/container/BUILD
+++ b/container/BUILD
@@ -139,6 +139,7 @@ TEST_TARGETS = [
     ":nodejs_image",
     ":py3_image",
     ":py_image",
+    ":py_image_with_symlinks_in_data",
     ":py_image_complex",
     ":rust_image",
     ":scala_image",
--- a/container/image_test.py
+++ b/container/image_test.py
@@ -499,6 +499,51 @@ class ImageTest(unittest.TestCase):
         './app/io_bazel_rules_docker/testdata/py_image_library.py',
       ])
 
+  def test_py_image_with_symlinks_in_data(self):
+    with TestImage('py_image_with_symlinks_in_data') as img:
+      # Check the application layer, which is on top.
+      self.assertTopLayerContains(img, [
+        '.',
+        './app',
+        './app/testdata',
+        './app/testdata/py_image_with_symlinks_in_data.binary.runfiles',
+        './app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker',
+        './app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/testdata',
+        './app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/testdata/py_image.py',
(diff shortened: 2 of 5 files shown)


scele:
@nlopezgi, thoughts about this change?

For the new test to pass, the CI would need to migrate to bazel 0.21.0. Is there something I can do to help with that? The transition to 0.21.0 was easy for me locally, the only thing I had to do was to install java on my system, since 0.20.0 introduced the following compat break:

> If the --javabase flag is unset, it Bazel locates a JDK using the JAVA_HOME environment variable and searching the PATH. If no JDK is found --javabase will be empty, and builds targeting Java will not
be supported. Previously Bazel would fall back to using the embedded JDK as a --javabase, but this is no longer default behaviour. A JDK should be explicitly installed instead to enable Java development

Older versions of bazel can still be used with rules_docker, but this particular new feature won't work there and that one test will fail when ran with older bazel.

nlopezgi:
travis Ci failed due to buildifier issues, Buldkite is indeed still using 0.20.0

scele:
> travis Ci failed due to buildifier issues

Ran buildifier in the latest patch.

nlopezgi:
Looks like you have an error still
```
'/app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/external'

  ['.',
   './app',
   './app/testdata',
   './app/testdata/py_image_with_symlinks_in_data.binary.runfiles',
   './app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker',
   './app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/testdata',
   './app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/testdata/py_image.py',
   './app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/testdata/py_image_with_symlinks_in_data.binary',
   './app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/testdata/foo.txt',
   './app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/testdata/__init__.py',
   '/app',
   '/app/testdata',
+  '/app/testdata/py_image_with_symlinks_in_data.binary',
   '/app/testdata/py_image_with_symlinks_in_data.binary.runfiles',
   '/app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker',
-  '/app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/foo-symlink.txt',
-  '/app/testdata/py_image_with_symlinks_in_data.binary',
   '/app/testdata/py_image_with_symlinks_in_data.binary.runfiles/io_bazel_rules_docker/external']
```

nlopezgi:
Kicking off tests again to see if BuildKite is now using 0.21.0

PR #631, opened 2018-12-20
Support '=' character in path names

scele:
Equals sign is a valid character in paths, so it should not be used
as a delimiter when passing file map to build_tar.py.  For example,
pyspark python package contains directories like 'cls=kittens'
(see https://github.com/apache/spark/tree/master/data/mllib/images/partitioned).

Pass contents as a json manifest file to build_tar.py to resolve the
issue.

### DIFF
--- a/container/build_tar.py
+++ b/container/build_tar.py
@@ -16,6 +16,7 @@
 from contextlib import contextmanager
 import gzip
 import io
+import json
 import os
 import os.path
 import subprocess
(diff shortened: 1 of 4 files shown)


testdata/BUILD
```diff
@@ -31,7 +31,7 @@ exports_files(["Runfiles.java"])
 
 genrule(
```
nlopezgi: could we instead add a target like this one in https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD + a target like gen_image (also add to https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD) and then add a container_image test that verifies the resulting file is present in the container (example config for container_test: https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/configs/test.yaml#L3)

scele: Oh that's sweet! Added new-style tests.

PR #632, opened 2018-12-21
Allow container_layer's data_path to work on directory output target names

EdSchouten:
There is a tiny bug in pkg_tar's strip_prefix handling that prevents it
from stripping the full name of a file. Though this sounds weird at
first, it is pretty useful when creating tarballs from output
directories. A pull request for this has been sent to Bazel:

https://github.com/bazelbuild/bazel/pull/6980

The same issue applies to container_layer's data_path argument, which is
actually where we noticed this at first. Also apply this fix here.

### DIFF
--- a/container/build_tar.py
+++ b/container/build_tar.py
@@ -393,7 +393,7 @@ def main(unused_argv):
   # Add objects to the tar file
   with TarFile(FLAGS.output, FLAGS.directory, FLAGS.compression, FLAGS.root_directory) as output:
     def file_attributes(filename):
-      if filename[0] == '/':
+      if filename.startswith('/'):
         filename = filename[1:]
       return {
           'mode': mode_map.get(filename, default_mode),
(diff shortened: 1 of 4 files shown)


xingao267:
@EdSchouten, thanks for the PR. Could you make a simple test for this?

EdSchouten:
> @EdSchouten, thanks for the PR. Could you make a simple test for this?

Done!

nlopezgi:
Thanks for adding the tests. Looks good. Will kick off Buildkite tests, but travis CI failed due to buildifier.
Can you run buildifier on ./tests/docker/BUILD

nlopezgi:
Thanks again, one question though, 
Do we need to update anything in the README after this PR is in? Specifically, does this PR change the statement in this section (https://github.com/bazelbuild/rules_docker/#container_image) - the bit that starts with "Hint: ..." ?

EdSchouten:
Hi there,

> Thanks again, one question though,

And thank you for merging this so quickly!

> Do we need to update anything in the README after this PR is in? Specifically, does this PR change the statement in this section (https://github.com/bazelbuild/rules_docker/#container_image) - the bit that starts with "Hint: ..." ?

I think that part of the documentation remains unaffected. It's just that archiving of directory outputs and stripping prefixes already worked, except when they were combined and stripped the entire path prefix.

PR #634, opened 2018-12-22
Use toolchains to resolve "docker" path in container_test

scele:


scele:
I needed this for container_image_tests to work at least with bazel 0.21.0 on macOS.

PR #636, opened 2018-12-22
Set path to xz in toolchain when possible

clintharrison:
The `xz` tool may be used in `build_tar.py` when Python 2 is used (which does not have a built-in `lzma` module) and `backports.lzma` is not installed. Bazel 0.21 flipped the default of `--incompatible_strict_action_env` to true, which means `which xz` is unlikely to work. This can be done in the repository rule to autoconfigure the docker toolchain.

I don't totally love that this is in the _docker_ toolchain, but it felt too heavyweight to separate it into its own, since other users of this toolchain can just ignore the new field.

### DIFF
--- a/container/layer.bzl
+++ b/container/layer.bzl
@@ -77,6 +77,7 @@ def build_layer(
         tars = None,
         operating_system = None):
     """Build the current layer for appending it to the base layer"""
+    toolchain_info = ctx.toolchains["@io_bazel_rules_docker//toolchains/docker:toolchain_type"].info
     layer = output_layer
     build_layer_exec = ctx.executable.build_layer
     args = [
@@ -84,6 +85,9 @@ def build_layer(
         "--directory=" + directory,
         "--mode=" + ctx.attr.mode,
     ]
+
+    if toolchain_info.xz_path != "":
+        args += ["--xz_path=%s" % toolchain_info.xz_path]
 
     # Windows layer.tar require two separate root directories instead of just 1
     # 'Files' is the equivalent of '.' in Linux images.
@@ -232,6 +236,7 @@ layer = struct(
     attrs = _layer_attrs,
     outputs = _layer_outputs,
     implementation = _impl,
+    toolchains = ["@io_bazel_rules_docker//toolchains/docker:toolchain_type"],
 )
 
 container_layer_ = rule(
@@ -239,6 +244,7 @@ container_layer_ = rule(
     executable = False,
     outputs = _layer_outputs,
     implementation = _impl,
+    toolchains = ["@io_bazel_rules_docker//toolchains/docker:toolchain_type"],
 )
 
 def container_layer(**kwargs):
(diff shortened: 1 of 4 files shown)


nlopezgi:
thanks for sending this PR. Overall looks good. Please resolve the conflicts. I wish there was a way to add a simple test for this. Will kick off tests.

container/layer.bzl
```diff
@@ -85,6 +86,9 @@ def build_layer(
         "--mode=" + ctx.attr.mode,
     ]
 
+    if toolchain_info.xz_path:
```
nlopezgi: `and toolchain_info.xz_path != "":`
(as this will be set to empty string if its not found on the path?)

clintharrison: just curious, is this a style difference between Python and Starlark? in Python it's common to rely on empty strings being falsy

nlopezgi: no, its what we need to check for as we're using the template to set the value, and the template includes the "{...}" so we cannot replace its value for None. Its part of our experimenting with toolchain rules + repository rules. You're probably right that this is not idiomatic and we should change how the template creation currently works in the repository rule. We'll make sure to refactor it all homogeneously when we do.

clintharrison: ahhh gotcha, thanks!

tsuna:
Hi, sorry for the question on this PR but can you explain how to ultimately set the new `xz_path` flag in `build_tar.py`? With this change I can't build kops out of the box anymore (which I first reported here: bazelbuild/homebrew-tap#36 but maybe I barked up the wrong tree).

clintharrison:
> Hi, sorry for the question on this PR but can you explain how to ultimately set the new `xz_path` flag in `build_tar.py`? With this change I can't build kops out of the box anymore (which I first reported here: [bazelbuild/homebrew-tap#36](https://github.com/bazelbuild/homebrew-tap/issues/36) but maybe I barked up the wrong tree).

@tsuna this PR would actually fix your issue, I believe. It looks like kops is still on v0.5.1 of rules_docker, which does not include the `xz_path` in the toolchain. Try bumping the version of rules_docker in order to build, or explicitly set `--action_env=PATH=...` to include your local `xz`, or build with `--noincompatible_strict_action_env`, since you're using Bazel 0.21.

Here's a quick diff that worked for me to get kops on rules_docker 0.7.0:
```diff
diff --git a/WORKSPACE b/WORKSPACE
index 336fa270b..f60c615c2 100644
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -34,13 +34,17 @@ gazelle_dependencies()
 git_repository(
     name = "io_bazel_rules_docker",
     remote = "https://github.com/bazelbuild/rules_docker.git",
-    tag = "v0.5.1",
+    tag = "v0.7.0",
+)
+
+load(
+    "@io_bazel_rules_docker//repositories:repositories.bzl",
+    container_repositories = "repositories",
 )
 
 load(
     "@io_bazel_rules_docker//container:container.bzl",
     "container_pull",
-    container_repositories = "repositories",
 )
 
 container_repositories()
```

tsuna:
Confirmed, that fixes it, thanks!

PR #639, opened 2019-01-03
Use incremental loader in container_test

scele:
When using the docker driver to run container_tests, use the
fast incremental loader instead of "docker loading" the tarball.
Producing the tarball can be very expensive for large images.

Also rewrite container_test runfiles assembly to avoid calls to
depset.to_list().

### DIFF
--- a/contrib/structure-test.sh.tpl
+++ b/contrib/structure-test.sh.tpl
@@ -6,4 +6,4 @@ set -ex
 
 %{test_executable} version
 
-%{test_executable} test --driver %{driver} --image %{image} %{configs} %{quiet}
+%{test_executable} %{args}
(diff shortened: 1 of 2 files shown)


xingao267:
@scele thanks for the PR. Could you update existing container_test targets in this repo?

xingao267:
It looks like it breaks the existing container_test targets in bazel-toolchains repo https://buildkite.com/bazel/bazel-toolchains/builds/1096#5923b00f-ddaa-480e-8658-57b70212d338

xingao267:
But it's possible that the rule in bazel-toolchains needs a fix. Still looking.

xingao267:
It looks like it's because we are missing the legacy base tarball in the runfiles so the loader cannot load the base.

xingao267:
It's the container_bundle that doesn't prepare the runfiles properly. I sent https://github.com/bazelbuild/rules_docker/pull/645

scele:
Thanks for root causing and fixing the issue! :)

PR #642, opened 2019-01-04
Make obtainer build user in tests remote execution compatible

smukherj1:
image_test.py currently uses the environment variable USER to obtain the
build user. While this matches the BUILD_USER when bazel is invoked
localy, it can be different when the test action is executed in a remote
machine. This commit makes the Bazel stamp variables available to
image_test.py so that it can obtain the BUILD_USER as set by Bazel.

### DIFF
--- a/container/image_test.py
+++ b/container/image_test.py
@@ -25,6 +25,8 @@ from containerregistry.client.v2_2 import docker_image as v2_2_image
 TEST_DATA_TARGET_BASE='testdata'
 DIR_PERMISSION=0o700
 PASSWD_FILE_MODE=0o644
+# Dictionary of key to value mappings in the Bazel stamp file
+STAMP_DICT = {}
 
 def TestData(name):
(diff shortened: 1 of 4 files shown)


container/image_test.py
```diff
@@ -830,6 +832,30 @@ def test_nodejs_image_args(self):
...
+  stamp_file = TestData("stamp_info_file.txt")
+  with open(stamp_file) as stamp_fp:
+    for line in stamp_fp:
+      # The first column in each line in the stamp file is the key
+      # and the second column is the corresponding value.
+      split_line = line.strip().split()
+      if len(split_line) == 0:
+        # Skip blank lines.
+        continue
+      key = ""
```
xingao267: or
```
key = split_line[0]
value = ""
if len(split_line) > 1
  value = " ".join(split_line[1:])
```
Both are fine to me. Your current implementation is better in readability.

PR #644, opened 2019-01-07
Migrate more tests to tests/docker

alex1545:
Migrating more tests to tests/docker

File existence tests:
* cc

Metadata tests:
* cc
* py
* py3
* go
* rust
* scala
* groovy

### DIFF
--- a/container/BUILD
+++ b/container/BUILD
@@ -151,6 +151,7 @@ TEST_DATA = [
     for t in TEST_TARGETS
 ] + [
     "//testdata:stamped_bundle_test",
+    "//testdata:stamp_info_file.txt",
 ]
 
 py_test(
--- a/container/BUILD
+++ b/container/BUILD
@@ -130,18 +130,11 @@ TEST_TARGETS = [
     ":cc_based_warning",
     "/pause_piecemeal:image",
     "/pause_piecemeal_gz:image",
-    ":cc_image",
     # Re-enable once https://github.com/bazelbuild/rules_d/issues/14 is fixed.
     # "d_image",
-    ":go_image",
-    ":groovy_image",
     ":java_image",
-    ":nodejs_image",
-    ":py3_image",
     ":py_image",
     ":py_image_complex",
-    ":rust_image",
-    ":scala_image",
     ":war_image",
     ":flat",
 ]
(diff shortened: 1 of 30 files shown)


container/BUILD
```diff
@@ -130,18 +130,11 @@ TEST_TARGETS = [
     ":cc_based_warning",
     "/pause_piecemeal:image",
     "/pause_piecemeal_gz:image",
-    ":cc_image",
```
nlopezgi: Is it possible to fully remove the cc_image target and some of the below targets now from https://github.com/bazelbuild/rules_docker/blob/master/testdata/BUILD#L756?

tests/docker/groovy/configs/groovy_image.yaml
```diff
@@ -0,0 +1,13 @@
+schemaVersion: 2.0.0
+
+metadataTest:
+  entrypoint: [
+    '/usr/bin/java',
+    '-cp',
+    '/app/io_bazel_rules_docker/tests/docker/groovy/libgroovy_image_library-impl.jar:/app/io_bazel_rules_docker/../com_google_guava_guava/jar/guava-18.0.jar:/app/io_bazel_rules_docker/../groovy_sdk_artifact/groovy-2.4.4/lib/groovy-2.4.4.jar:/app/io_bazel_rules_docker/tests/docker/groovy/libgroovy_image.binary-lib-impl.jar:/app/io_bazel_rules_docker/tests/docker/groovy/groovy_image.binary.jar:/app/io_bazel_rules_docker/tests/docker/groovy/groovy_image.binary:/app/io_bazel_rules_docker/tests/docker/groovy/BUILD',
```
xingao267: Can this be broken into multilines?

alex1545: I couldn't find a way to. All the ways to break a string into multilines in a YAML file will either insert a space or a new line char into the string.

xingao267: Ack

xingao267:
Check with @smukherj1 before merging. @smukherj1 is doing something with copybara.

smukherj1:
Suggest waiting until my copybara CL is then. Then I suggest we use CaaS to sync a merged PR to internal codebara before merging any other PRs

nlopezgi:
I think we can now fix the issues in this PR and get it submitted

PR #646, opened 2019-01-08
Convert dict.items() calls to use dicts.add.

joshuamorton:


container/bundle.bzl
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 """Rule for bundling Container images into a tarball."""
 
+load("//skylib:dicts.bzl", "dicts")
```
nlopezgi: did you forget to add skylib/dicts.bzl to this PR?

xingao267: It should be load("@bazel_skylib//skylib:dicts.bzl", "dicts")

xingao267: The other ones doesn't have @bazel_skylib because they locate in this repo, under the skylib folder. But dicts.bzl is not. It's in https://github.com/bazelbuild/bazel-skylib/blob/master/lib/dicts.bzl. That has to be loaded as an external dependency. And we have already loaded that repo here https://github.com/bazelbuild/rules_docker/blob/master/container/container.bzl#L218

joshuamorton: There still looks to be an ordering issue. I can play around with that, or we can just vendor the one function. Opinions?

xingao267: It's because rules_docker use to repositories() in container/container.bzl to add all its dependencies including the bazel_skylib. Now since we are imposing a new dependency of bazel_skylib to the container_{pull, image, push, ...} rules, so container/container.bzl cannot be loaded unless bazel_skylib is imported before that. It's kinda a cyclic dependency. We have some options here:

1. Manually import the bazel_skylib repo at the very top of the WORKSPACE file. But this would require all the users of rules_docker to do the same, i.e. they would need to add the http_archive target of bazel_skylib before the http_archvie target of rules_docker in order for the container/container.bzl of rules_docker to be loaded properly. This sounds not very preferable to me.

2. Separate the current container/container.bzl file into two files. One that only contains a repositories def which only imports Bazel native workspace rules, i.e. no external dependencies required so that this can be safely loaded at the top of any workspace file. And the other one would contain the load statements of container_{pull, image, push, ...}. The second file has to be loaded after the repositories def from the first file is loaded (which contains the bazel_skylib). This would also cause a change in behavior of how people are currently use rules_docker.

3. Keep the current .item() way, which does not seem like it is not planned to be deprecated. By not using the dicts.bzl file, we remove one dependency.

4. Fork the dicts.bzl file in this repo. Not good.

xingao267: To me, I would prefer option 3 as it will not break the current behavior.

xingao267:
Close in favor of #652.

PR #648, opened 2019-01-09
Fix Bazel incompatible issues.

xingao267:
Some refs:
- https://github.com/bazelbuild/bazel/issues/5817


xingao267:
@laurentlb could you help take a look at this PR which resolve various Bazel incompatible issues?

java/image.bzl
```diff
@@ -98,15 +102,15 @@ def java_files(f):
     files = []
     if java_common.provider in f:
         java_provider = f[java_common.provider]
-        files += list(java_provider.transitive_runtime_jars)
+        files += java_provider.transitive_runtime_jars.to_list()
     if hasattr(f, "files"):  # a jar file
-        files += list(f.files)
+        files += f.files.to_list()
     return files
```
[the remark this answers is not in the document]
xingao267: Thanks for the suggestion. I'll do do that in a separate PR.

container/image.bzl
```diff
@@ -229,8 +229,8 @@ def _assemble_image_digest(ctx, name, image, image_tarball, output_digest):
 
     ctx.actions.run(
         outputs = [output_digest],
-        inputs = [image["config"]] + blobsums + blobs +
-                 ([image["legacy"]] if image.get("legacy") else []),
+        tools = [image["config"]] + blobsums + blobs +
```
ash2k: @xingao267 what is the reason for this change? Just trying to understand. I thought `tools` is for "tools" that used to perform the action and the `inputs` are for inputs i.e. files that are processed by the tools. Here all these things are inputs, not files, I think. Same in other files below.

xingao267: I agree. I think I made that change because Bazel complains that some of these files were executable, so I put them into tools according to https://docs.bazel.build/versions/master/skylark/backward-compatibility.html#disallow-tools-in-action-inputs. Let me try to change it back to inputs to see if the CI complains about it.

ash2k: I wonder why they are executable... From that page:

> In the rare case that your action requires a tool as input, but does not actually run the tool and therefore does not need its runfiles, the safety check will fail even though the action would have succeeded. In this case, you can bypass the check by adding a (possibly empty) tools argument to your action.

xingao267: Yea. I was surprising too. I sent https://github.com/bazelbuild/rules_docker/pull/716/files to changed it back and I think it's fine. Maybe I mistakenly changed it.

xingao267: Thanks for pointing out!

PR #651, opened 2019-01-10
Fix container bundle pusher broken by #591

### LINKED ISSUE #609: //contrib/push-all.bzl `container_push` fails
Pushes with that rule fail with an error as of 5eb0728594013d746959c4bd21aa4b0c3e3848d8:

```
usage: fast_pusher_.py [-h] --name NAME [--tarball TARBALL] [--config CONFIG]
                       [--manifest MANIFEST] [--digest DIGEST] [--layer LAYER]
                       [--stamp-info-file STAMP_INFO_FILE] [--oci]
                       [--client-config-dir CLIENT_CONFIG_DIR]
                       [--stderrthreshold STDERRTHRESHOLD]
fast_pusher_.py: error: argument --name is required
```

The args being used for `push-tag-sh.tpl` are no longer correct, so the pusher call is invalid.

smukherj1:
This fixes #609. Also added end to end test to verify bundle pusher and
regular pusher upload legal images.


tests/contrib/BUILD
```diff
@@ -154,3 +154,17 @@ docker_push(
     name = "test_docker_push_empty_bundle",
     bundle = ":create_empty_bundle",
 )
+
+container_bundle(
```
nlopezgi: move this to https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD

tests/docker/BUILD
```diff
@@ -252,10 +252,10 @@ container_push(
 container_push(
     name = "push_stamped_test",
     format = "Docker",
-    image = ":set_cmd",
+    image = "@io_bazel_rules_docker//testdata:files_in_layer_with_files_base",
```
nlopezgi: why do you need to use this one as base?
also, you should not need the @io_bazel_rules_docker

tests/docker/BUILD
```diff
@@ -252,10 +252,10 @@ container_push(
...
     registry = "localhost:5000",
-    repository = "docker/test/{BUILD_USER}",
-    tag = "{BUILD_USER}",
+    repository = "docker/test/{TEST_STAMP}",
```
nlopezgi: why not use a fixed repository name / tag and avoid having to use print-workspace-status.sh altogether?
i.e., you probably can create a new container_push that is simpler than this one to use for your e2e test.

tests/contrib/BUILD
smukherj1: Done

tests/docker/BUILD
smukherj1: I wanted an image that has multiple layers. The e2e tests push and then use docker pull on this image. This is expected to catch the scenario where the number of layers in the manifest and the image config disagree which would result in the docker pull failing

smukherj1: Removed @io_bazel_rules_docker

tests/docker/BUILD
smukherj1: Revert this test to only test the stamp expansion

PR #659, opened 2019-01-13
Match the full package name.

jo2y:
\w does not include hyphen, so packages with - in their name generate
the same filenames. A random one is included into the final image.

Testing suggests that the files put into the image based on these are how https://cloud.google.com/container-registry/docs/container-analysis determines what is installed. Having non-overlapping filenames means that all the metadata is present for the analysis.

### DIFF
--- a/container/build_tar.py
+++ b/container/build_tar.py
@@ -104,7 +104,7 @@ class TarFile(object):
   class DebError(Exception):
     pass
 
-  PKG_NAME_RE = re.compile(r'Package:\s*(?P<pkg_name>\w+).*')
+  PKG_NAME_RE = re.compile(r'Package:\s*(?P<pkg_name>[^\s]+).*')
   DPKG_STATUS_DIR = '/var/lib/dpkg/status.d'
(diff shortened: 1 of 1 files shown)


nlopezgi:
Thanks for sending this PR. 
Would it be possible for you to add a test that highlights the error scenario that you are fixing?
You could add a new subpackage under `https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/` that has '-' in the name and one w/o and use that in a new target you create in `https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD`?

jo2y:
I think I have a handle on what you're requesting that I add. Should I create a test of the file generation in a different PR and then expand it for the hyphen use-case in this one or just add everything here?

nlopezgi:
thanks for your response @jo2y ,
adding it here would be fine.

nlopezgi:
Hi @jo2y ,
do you think you'll get the chance to finish this PR or should we close?

PR #661, opened 2019-01-14
Remove --all_incompatible_changes from presubmit.yml

meteorcloudy:
Adding this flag will cause rules_docker to be red in bazel downstream when new incompatible changes come.
FYI @hlopko @laurentlb @nlopezgi


meteorcloudy:
closing as a duplicate of #662

PR #662, opened 2019-01-14
Do not add --all_incompatible_changes to presubmit

hlopko:
The incompatible change process is a little bit more involved than using `--all_incompatible_changes`. We allow creating new incompatible flags which we don't expect users to migrate for right away. The principled place to look for incompatible changes is to go to https://github.com/bazelbuild/bazel/issues?q=is%3Aissue+label%3Aincompatible-change+label%3Amigration-ready, ideally also filtered with the bazel version (e.g. for flags that should be migrated in bazel 0.22.0, add the label 'migration-0.22'.

On the buildkite, this is the pipeline that tells rule owners which incompatible flags they have to migrate for in the current bazel release: https://buildkite.com/bazel/bazel-at-release-plus-incompatible-flags

This PR fixed the failure on the Bazel downstream pipeline: https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/745#477cd902-ce69-4aa3-9f07-137bbc40649b

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -18,7 +18,6 @@ platforms:
     # tests/contrib/test_compare_ids_test_* expect 'bazel' on path
     test_flags:
     - "--action_env=PATH"
-    - "--all_incompatible_changes"
     # TODO(xingao): Remove after https://github.com/bazelbuild/rules_scala/issues/644 is addressed.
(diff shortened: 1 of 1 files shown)


hlopko:
CC @meteorcloudy

xingao267:
Thanks for the fix!

Issue #670, opened 2019-01-18
Support make variables in env field


syml:
I would like to specify some env vars at build time. Make variables are not currently supported in the env field.

nlopezgi:
Hi @syml ,
Thanks for opening this issue.
Supporting make variables for the env field in container_image (and other related rules) should not be hard. We currently support this for, e.g., tags in the images attr of container_bundle (doc: https://github.com/bazelbuild/rules_docker#container_bundle-1 / impl of make variable substitution:
https://github.com/bazelbuild/rules_docker/blob/master/container/bundle.bzl#L46). It should be fairly straightforward to add code in image.bzl that does make variable substitution somewhere around here: https://github.com/bazelbuild/rules_docker/blob/master/container/image.bzl#L304 + adding tests somewhere around here: https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD#L206. If you're willing to volunteer a PR I'd be happy to review it and guide you through any issues you have.

PR #672, opened 2019-01-23
Expand make variables in env field

syml:
This addresses https://github.com/bazelbuild/rules_docker/issues/670
I haven't figured out how to write a test that can set Make variables. Please let me know what is a good way to test this.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -65,6 +65,9 @@ platforms:
     - "--platforms=@bazel_toolchains//configs/ubuntu16_04_clang/1.1:nosla_xenial_docker"
     # TODO(xingao): Remove after https://github.com/bazelbuild/rules_scala/issues/644 is addressed.
     - "--incompatible_disallow_legacy_javainfo=false"
+    # For tests/docker:set_env_make_vars
+    - "--define=ENV_KEY=my_key"
+    - "--define=ENV_VALUE=my_value"
     test_targets:
     - "--"
     - "//:structure_test_at_workspace_root"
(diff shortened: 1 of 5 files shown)


nlopezgi:
Hi @syml ,
Thanks for putting this PR together.
To test you can add a new target to tests/docker/BUILD like this one:
```
container_image(
    name = "with_env_make_var_substitution",
    base = ":base_with_volume",
    env = {
        "$(ENV_KEY)": "$(ENV_VALUE)",
    },
)
```

To verify this works locally, you can run something like:
`bazel build --define=ENV_KEY=my_key --define=ENV_VALUE=my_value  //tests/docker/BUILD:with_env_make_var_substitution`

To test this locally, you will need to add a container_test similar to this one: 
https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD#L215
With a corresponding yaml file like this one https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/configs/set_env_csv.yaml
(note the yaml file should contain the expanded values, i.e.:
```
env:
  - key: my_key
    value: my_value
```

To test this on CI, you will then need to modify the presubmit yaml  (the config for `rbe_ubuntu1604` - https://github.com/bazelbuild/rules_docker/blob/master/.bazelci/presubmit.yml#L78)
to add the new define flags to both build and test flags
```
    - "--define=ENV_KEY=my_key"
    - "--define=ENV_VALUE=my_value"
```
(with a short doc comment indicating which tests are depending these flags)

Please let me know if you run into issues or have questions, and thanks again.

syml:
Added tests. Thanks!

PR #691, opened 2019-02-14
container_push: added attribute 'tag_file'

excavador:
I need ability to specify tag of image for push as content of file.
I added `tag_file` attribute to `container_push` rule.

### DIFF
--- a/container/push.bzl
+++ b/container/push.bzl
@@ -88,6 +88,10 @@ def _impl(ctx):
     registry = ctx.expand_make_variables("registry", ctx.attr.registry, {})
     repository = ctx.expand_make_variables("repository", ctx.attr.repository, {})
     tag = ctx.expand_make_variables("tag", ctx.attr.tag, {})
+    runfiles_tag_file = []
+    if ctx.file.tag_file:
+        tag = "$(cat {})".format(_get_runfile_path(ctx, ctx.file.tag_file))
+        runfiles_tag_file = [ctx.file.tag_file]
     if ctx.attr.stamp:
         print("Attr 'stamp' is deprecated; it is now automatically inferred. Please remove it from %s" % ctx.label)
 
@@ -120,7 +124,7 @@ def _impl(ctx):
         output = ctx.outputs.executable,
         is_executable = True,
     )
-    runfiles = ctx.runfiles(files = [ctx.executable._pusher] + image_files + stamp_inputs)
+    runfiles = ctx.runfiles(files = [ctx.executable._pusher] + image_files + stamp_inputs + runfiles_tag_file)
     runfiles = runfiles.merge(ctx.attr._pusher.default_runfiles)
 
     return [
@@ -166,6 +170,10 @@ container_push = rule(
         "tag": attr.string(
             default = "latest",
             doc = "(optional) The tag of the image, default to 'latest'.",
+        ),
+        "tag_file": attr.label(
+            allow_single_file = True,
+            doc = "(optional) The label of the file with tag value. Overrides 'tag'.",
         ),
         "_digester": attr.label(
             default = "@containerregistry//:digester",
--- a/tests/docker/BUILD
+++ b/tests/docker/BUILD
@@ -269,6 +269,24 @@ container_push(
     tags = ["manual"],
 )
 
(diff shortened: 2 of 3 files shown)


container/push.bzl
```diff
@@ -167,6 +171,10 @@ container_push = rule(
             default = "latest",
             doc = "(optional) The tag of the image, default to 'latest'.",
         ),
+        "tag_file": attr.label(
+            allow_single_file = True,
+            doc = "(options) The label of the file with tag value. Overrides 'tag'.",
```
xingao267: (optional)

container/push.bzl
```diff
@@ -99,6 +99,10 @@ def _impl(ctx):
         "--stamp-info-file=%s" % _get_runfile_path(ctx, f)
         for f in stamp_inputs
     ]
+    runfiles_tag_file = []
```
xingao267: Move this new section to be after line 90 so it's more clear how tag is set.

excavador:
Will do, thank you!

excavador:
> One that is similar to https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD#L262 and https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e.sh#L390

Done

> And another file_test (example: https://github.com/GoogleContainerTools/base-images-docker/blob/master/tests/package_managers/BUILD#L32) that validates that the generated executable file has the desired content (e.g. the tag that reads from a file)

Should I add this test to rules_docker or to https://github.com/GoogleContainerTools/base-images-docker/?

xingao267:
In rules_docker as well. Probably in file
https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD
as well. Similar to other file_test.

excavador:
> In rules_docker as well. Probably in file https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/BUILD as well. Similar to other file_test.

I see. Done.

excavador:
@xingao267 fixed PR comments

excavador:
rebased to latest master

excavador:
@xingao267 everything fixed, tests added, PR rebased to latest master.

tests/docker/BUILD
```diff
@@ -285,6 +304,12 @@ file_test(
     file = ":push_test.digest",
 )
 
+file_test(
+    name = "test_push_tag_file_digest_output",
```
xingao267: Don't check the digest output, check the default output executable that it contains the right tag.

tests/docker/BUILD
```diff
@@ -269,6 +269,25 @@ container_push(
...
+    name = "push_test_tag_file",
+    format = "Docker",
+    image = ":set_cmd",
+    registry = "localhost:5000",
+    repository = "docker/test",
+    tag_file = ":test.tag",
+    tags = ["manual"],
+)
+
+
```
xingao267: extra line

excavador:
@xingao267 can you please point me to specific example of the test, where tag checked?
I re-read all rules_docker completely and did not understand what the right way to do it.
I have only single idea: in shell script I should execute bazel run <mytarget>, than try to pull container by tag and compare digests

xingao267:
@excavador So when you build the new container_push target //tests/docker:push_test_tag_file you created, it should create several outputs, one is `bazel-bin/tests/docker/push_test_tag_file.digest` and one is `bazel-bin/tests/docker/push_test_tag_file`. So what I was saying in the comment is do not use the `:push_test_tag_file.digest` output, but use `:push_test_tag_file` instead. That file should contain a pusher command that includes the tag of the container. You should make sure that tag is correct.

So try to do this on your command line and you will see:

```
bazel build //tests/docker:push_test_tag_file
cat bazel-bin/tests/docker/push_test_tag_file
```
Then in the `content` you want to validate in the `file_test` rule, you need to check that file containers a regex that has the tag in it.

excavador:
@xingao267 thank you so much, now got it!
I knew about the executable script file but did not imagine to use it for verification.

Tests added, rebased to latest master, seems like now I fixed everything

xingao267:
Thanks! Looks good!

PR #694, opened 2019-02-20
Set default_mtime=0 on TarFileWriter so that elements default to the epoch

aiuto:
Context: https://github.com/bazelbuild/bazel/pull/7276

### DIFF
--- a/container/build_tar.py
+++ b/container/build_tar.py
@@ -126,7 +126,8 @@ class TarFile(object):
     self.tarfile = archive.TarFileWriter(
         self.output,
         self.compression,
-        self.root_directory
+        self.root_directory,
+        default_mtime=0
     )
     return self
 


aiuto:
This is waiting on a PR to bazelbuild/bazel which adds default_mtime.

nlopezgi:
Hi @aiuto ,
thanks for sending this PR. Its failing on CI as the archive.py version does not include yet your change.

I think we'll need to import bazel head like we did recently for a similar change:
https://github.com/bazelbuild/rules_docker/commit/e9e2809b7099f92e46f9e8336cda041ab78fcf35#diff-4b8375c5820da1f50741dad71db73d2cR94

and then import that bazel head archive in build_tar.py:
https://github.com/bazelbuild/rules_docker/commit/e9e2809b7099f92e46f9e8336cda041ab78fcf35#diff-ffe0f7bdca587bf6daaded6123f45b70R27

nlopezgi:
I've prepared https://github.com/bazelbuild/rules_docker/pull/695 to replace this PR

PR #695, opened 2019-02-20
Set default_mtime=0 on TarFileWriter so that elements default to the epoch

nlopezgi:
Replaces https://github.com/bazelbuild/rules_docker/pull/694 
* adds import of bazel at commit at which mtime was added

### DIFF
--- a/container/BUILD
+++ b/container/BUILD
@@ -79,8 +79,8 @@ py_library(
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [
+        "@bazel_source//tools/build_defs/pkg:archive",
         "@bazel_tools//third_party/py/gflags",
-        "@bazel_tools//tools/build_defs/pkg:archive",
     ],
 )
 
--- a/container/build_tar.py
+++ b/container/build_tar.py
@@ -25,7 +25,7 @@ import re
 import tarfile
 import tempfile
 
-from tools.build_defs.pkg import archive
+from bazel_source.tools.build_defs.pkg import archive
 from third_party.py import gflags
 
 gflags.DEFINE_string('output', None, 'The output file, mandatory')
@@ -126,7 +126,8 @@ class TarFile(object):
     self.tarfile = archive.TarFileWriter(
         self.output,
         self.compression,
-        self.root_directory
+        self.root_directory,
+        default_mtime=0
     )
     return self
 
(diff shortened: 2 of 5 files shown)


nlopezgi:
I'm reluctant to merge this PR as there is a change in the SHA's since (i think) we cannot change the value of mtime here:https://github.com/bazelbuild/bazel/commit/45e68f2f4344afa32f91b835888123cda3a77dc6#diff-002c180b932b0dbae88c2559af105850R125) . We can alterantively wait until the PR that adds the default_mtime option is in Bazel head and then use that.

aiuto:
The change to archive.py went in here:
https://github.com/bazelbuild/bazel/commit/1cf75b9dab3e0bb2ffa3014d6e657aa486b1c7bf

I will still have to put it behind an incompatible change flag, but that should not impact any code in rules_docker.

nlopezgi:
Awesome, updated to pin to https://github.com/bazelbuild/bazel/commit/1cf75b9dab3e0bb2ffa3014d6e657aa486b1c7bf and that did not require changes to the shas in the tests! However, we have 2 tests still failing, which rely on the build timestamp being set to a specific value. I think that I should be able to resolve this issue by piping the desired timestamp to archive.py, but don't want to block you while I do that. Will disable the test temporarily in this PR and send another PR later to fix / re-enable.

nlopezgi:
@smukherj1 pointed out that those timestamp tests are flaky and sometimes require a full bazel clean. So looks like this PR is all good and there are no pending tests for me to fix. Will merge once CI passes.

aiuto:
As I dig into this. I notice that bazel/tools/pkg/build_tar.py is almost identical to rules_docker/contain/build_tar.py. In fact, the comments in the bazel version say "build a docker layer", which is not really true, but it does indicate that Bazel lifted if from rules_docker sometime in the past.
So, I'm not sure how you are picking
I'm also not fond of this change, because of the sha change you mentioned.

What I have working now is a change to the Bazel pkg_* rules which does not require any changes to rules docker. It will only do the epoch change for direct users of pkg_tar().

smukherj1:
build_tar.py was originally forked to rules_docker from Bazel, not the other way around. I believe the reason was that rules_docker needed updates to build_tar.py faster than Bazel's release cadence. Given that Bazel now has automated sync using copybara from internal codebase to Github, maybe it makes sense to combine them again? ie use the one on Bazel as the source of truth.

nlopezgi:
We did not have to end up doing any changes to shas in this PR.
We could consider switching back to using build_tar.py from Bazel repo, but I think the release cadence will keep being an issue unless we want to keep pulling Bazel head in this repo to get to build_tar.py / archive.py. But let me ask, @aiuto, do you know if there is a reason why do the pkg rules get shipped with bazel binary and not live in a separate repo so that we can pin/depend on that?

PR #696, opened 2019-02-24
Use internally downloaded @six and disable legacy_create_init

aaliddell:
This moves the `py_binary` rules to use the internally downloaded `@six` dependency, rather than the one from `@bazel_tools`. Previously, both versions of six were in the runfiles, causing odd issues when the incorrect one was resolved first when six was imported.

Additionally, to fix the `__init__.py` issue, this also toggles `legacy_create_init`, which has been done for all targets that support it. The `//container:build_tar` target is the only one that still needs this enabled, due to the dependency on `@bazel_source` that appears temporary.

Alternatively, the internally downloaded six could be removed and you could depend on the `@bazel_tools` one.

### DIFF
--- a/container/BUILD
+++ b/container/BUILD
@@ -50,7 +50,7 @@ py_binary(
     visibility = ["//visibility:public"],
     deps = [
         ":utils",
-        "@six",
+        "@six//:six",
         "@containerregistry",
     ],
     legacy_create_init = False,
@@ -62,12 +62,13 @@ py_binary(
     visibility = ["//visibility:public"],
     deps = [
         ":utils",
-        "@six",
+        "@six//:six",
         "@containerregistry",
     ],
     legacy_create_init = False,
 )
 
+# TODO(xingao): Flip legacy_create_init once dependency on @bazel_source removed
 py_binary(
     name = "build_tar",
     srcs = ["build_tar.py"],
--- a/container/BUILD
+++ b/container/BUILD
@@ -33,6 +33,7 @@ py_binary(
     deps = [
         "@containerregistry",
     ],
+    legacy_create_init = False,
 )
 
 py_library(
@@ -52,6 +53,7 @@ py_binary(
         "@six",
         "@containerregistry",
     ],
+    legacy_create_init = False,
 )
 
 py_binary(
@@ -63,6 +65,7 @@ py_binary(
         "@six",
         "@containerregistry",
     ],
+    legacy_create_init = False,
 )
 
 py_binary(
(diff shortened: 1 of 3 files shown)


aaliddell:
56a3d15 rolls back the removal of renaming `six.py` to `__init__.py` (305c4b480157c3ff6fb1e34877c5c48debd8c520), due to `@six` being shared with containerrigistry which imports `six.moves...`, requiring it be named `__init__.py`.

If desired I can remove these two commits from the history to get the same result?

xingao267:
I'm leaning towards use the @bazel_tools//third_party/py/six one, remove the @six in rules_docker and make oauth2client to depends on @bazel_tools//third_party/py/six as well.

xingao267:
I don't see `//container:build_tar` still depends on `@bazel_source` though.

aaliddell:
Ok, I'll add a commit removing the internal six.

`build_tar` uses `build_tar_lib`, which has a dep on `@bazel_source//tools/build_defs/pkg:archive`
Whether @bazel_source is still needed here is not something I can comment on, as there appears to be a note about it elsewhere in the code.

xingao267:
My bad. Saw that now. Added in https://github.com/bazelbuild/rules_docker/commit/2078cb21a8f7e47ce06c69e380f723a689333d26. I was looking at an older state of the repo.

container/BUILD
```diff
@@ -71,6 +74,7 @@ py_binary(
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [":build_tar_lib"],
+    legacy_create_init = True,
```
xingao267: You can add a TODO assigned to me (xingao) to try to flip this to False once we remove the dependency on @bazel_source (if that's the only reason that blocks flipping this bit.)

xingao267: Use Format

# TODO(xingao): .....

aaliddell: Ok, added in latest commit. Yes, it appears that it's only the dependency on @bazel_source that requires that flag.

aaliddell:
Interesting, builds are failing because @containerregistry was depending on the six downloaded by rules_docker, which is somewhat insane. I think this can be fixed by adding defs.bzl from containerregistry to WORKSPACE

xingao267:
We need to change containerregistry as well. https://github.com/google/containerregistry/blob/master/def.bzl#L38 and https://github.com/google/containerregistry/blob/master/BUILD.bazel#L29

aaliddell:
That may not be possible as apparently containerregistry is no longer maintained: https://github.com/google/containerregistry/issues/114

Perhaps it is better to stay with internal six as at 
7f2f876 until the dependency on containerregistry is removed, as per https://github.com/bazelbuild/rules_docker/issues/580

xingao267:
ok. Using @six everywhere sounds good to me.

PR #697, opened 2019-02-25
Forward new provider for python in filter_layer

aaliddell:
In upcoming Bazel 0.23.0, the legacy `py` provider [has been changed to `PyInfo`](https://github.com/bazelbuild/bazel/issues/7298), with both being allowed to work currently. However, in a future release the use of `py` will be not be permitted, which can be simulated with `--incompatible_disallow_legacy_py_provider`.

This change forwards the old `py` provider in `filter_layer` in older Bazel, and `PyInfo` in newer Bazel and has been tested to work correctly when `--incompatible_disallow_legacy_py_provider` is enabled. Prior to this change, `filter_layer` would fail when run in this mode.

### DIFF
--- a/lang/image.bzl
+++ b/lang/image.bzl
@@ -287,21 +287,21 @@ def _filter_layer_rule_impl(ctx):
             runfiles = runfiles.merge(dep.target.default_runfiles)
             filtered_depsets.append(dep.target_deps)
 
-    # Forward correct provider, depending on Bazel version, so that the filter_layer() can be
+    # Forward correct provider, depending on availability, so that the filter_layer() can be
     # used as a normal dependency to native targets (e.g. py_library(deps = [<filter_layer>])).
     if hasattr(ctx.attr.dep, "py"):
-        # Forward legacy builtin provider
+        # Forward legacy builtin provider and PyInfo provider
         return struct(
             providers = [
                 FilterLayerInfo(
                     runfiles = runfiles,
                     filtered_depset = depset(transitive = filtered_depsets),
                 ),
-            ],
+            ] + ([ctx.attr.dep[PyInfo]] if PyInfo in ctx.attr.dep else []),
             py = ctx.attr.dep.py if hasattr(ctx.attr.dep, "py") else None,
         )
     else:
-        # Forward the PyInfo provider in Bazel >= 0.23.0
+        # Forward the PyInfo provider only
         return struct(
             providers = [
                 FilterLayerInfo(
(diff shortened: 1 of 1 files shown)


aaliddell:
Ah, unfortunately Starlark behaves slightly differently from what I expected from full Python when using identifiers that are not defined in a not-taken branch. I can't see an obvious way around this to allow picking the correct provider when `PyInfo` is not defined in Bazel < 0.23.0 and for the time being the `py` provider is allowed.

Perhaps instead of proxying an externally created `py_library` in `filter_layer`, the rule should instead create and return a `native.py_library` internally after doing the deps filtering, which would itself have the correct python provider, rather than trying to forward it?

xingao267:
Hi @aaliddell, thanks for sending the PR. We can wait until Bazel 0.23.0 is out before making this change. And once the --incompatible_disallow_legacy_py_provider flag is flipped in either Bazel 0.24.0 or 0.25.0, we can remove forwarding legacy py provider at all.

aaliddell:
Ok. If you're going to require Bazel 0.23.0 anyway, you might as well always forward the PyInfo provider to match the behaviour of the built-in python rules, such as (not tested):

```
    # Forward correct provider, depending on availability, so that the filter_layer() can be
    # used as a normal dependency to native targets (e.g. py_library(deps = [<filter_layer>])).
    if hasattr(ctx.attr.dep, "py"):
        # Forward legacy builtin provider and PyInfo provider
        return struct(
            providers = [
                FilterLayerInfo(
                    runfiles = runfiles,
                    filtered_depset = depset(transitive = filtered_depsets),
                ),
            ] + ([ctx.attr.dep[PyInfo]] if PyInfo in ctx.attr.dep else []),
            py = ctx.attr.dep.py if hasattr(ctx.attr.dep, "py") else None,
        )
    else:
        # Forward the PyInfo provider only
        return struct(
            providers = [
                FilterLayerInfo(
                    runfiles = runfiles,
                    filtered_depset = depset(transitive = filtered_depsets),
                ),
            ] + ([ctx.attr.dep[PyInfo]] if PyInfo in ctx.attr.dep else []),
)
```

Then in 0.24.0 or 0.25.0, the branch containing py can be dropped.

xingao267:
Yes. We can hold on this until Bazel 0.23.0 is out.

aaliddell:
I've implemented the proposed change assuming 0.23.0+, which was just released. Once things have settled and the CI version has been updated, this can be retested.

xingao267:
CI has switched to 0.23.1 and the tests passed. Can you fix the buildifier issue?

aaliddell:
The buildifier issue appears to have come in with the merge from master and the output on buildkite contains no actual output. Any ideas what it might be complaining about?

xingao267:
It's because Buildifier on Buildkite CI is upgrade today https://github.com/bazelbuild/continuous-integration/commit/6d48796af439f8f75c042ed3befacd49e3cc8197. Hold on a little on this PR and we will fix it.

PR #701, opened 2019-02-28
nodejs_image: Allow passing in custom binary.

Globegitter:


smukherj1:
Hi @Globegitter, thanks for this PR. Could you please let us know what's the specific use case you have that requires using a custom binary with nodejs_image? If you are using a given binary in multiple images, would it make sense to have something like the following:-

1. nodejs_image(name="A") produces a binary target ":A.binary"
2. Use :A base the base image in another nodejs_image where you originally wanted to include the custom binary.
3. Use :A.binary if you were reusing the binary you wanted to inject into the nodejs_binary

Also, if we do decide to go with this modification, we would also want some tests for it. But let's hold off on adding tests before we figure out if we can satisfy your use case with what we have already.

Globegitter:
@smukherj1 thanks for the response, this is partially to get parity with some of the other image rules, e.g. `go_image` allows for exactly the same thing and we have generally two use cases for this, one we sometimes want a different name for the binary target than what the image macro creates; I know we could use an alias but that also just adds extra config and an extra target. The second one is that we needed to wrap the `nodejs_binary` with another custom rule that adds some custom setup that needs to run before the binary, e.g. setting env vars with access to `pwd` etc and then we want to pass that wrapped binary into the nodejs_image.

smukherj1:
@Globegitter Thanks for explanation! This change SGTM. Could you please add some tests for this? Recommend adding them in https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/nodejs

Globegitter:
Will do :)

tests/docker/nodejs/BUILD
```diff
@@ -14,6 +14,7 @@
 
 load("//contrib:test.bzl", "container_test")
 load("//nodejs:image.bzl", "nodejs_image")
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
```
nlopezgi: run buildifier to order these loads properly

Globegitter: Done.

Issue #665, opened 2019-01-15
--no-run does not stop image from running when image build contains `args` attribute


Jdban:
Tested in 0.5.1 and 0.7.0

Here's a repository with a simple example to demonstrate the issue:

https://github.com/Jdban/rules_docker_bug_example

The `bazel run //:java_image_fails_with_args -- --norun` command does indeed create a docker image with the arg appended to the` EntryPoint`, but it should not run it immediately because the --norun flag is appended

alex1545:
Hello @Jdban,

I was able to reproduce this issue and it looks that the introduction of passing the args to the binary (in a `java_image` target in this case) turns out to be incompatible with the `--norun` option. We apologize for this inconvenience as this was something we overlooked.

Can you please describe your use case where the `--norun` flag is important to you, so we could still try and help?

Jdban:
Our use case was that we wanted to add args to the entrypoint of a docker image and build the image. 

We're building the image with ```bazel run :target -- --norun```

We will be fine for now, we ended up not needing this yet. It'd definitely be nice for this to be fixed in the future though.

nlopezgi:
The main issue is that we are using the same mechanism (args passed to bazel run) for two different features and we have no way of generically recognizing if an arg is to be passed to an image or is to be interpreted by the script to mean 'norun' is desired. 

One viable alternative would be to provide an additional output for the rules which can be executed via `bazel run :target.norun`. The output (e.g., norun_output) would need to be created with the same info as the output here (https://github.com/bazelbuild/rules_docker/blob/master/container/layer_tools.bzl#L181), and likely a small variation in the template would be sufficient to enable layer_tools.bzl to produce both files. You'd need to then declare the output in the rules that execute incremental_load (and pass the norun_output to incremental_load).

Making such a change would not be hard, and tests would be relatively simple as well. If you're willing to volunteer a PR we would be happy to guide you through more details and review it!

enigmatix:
@alex1545 @nlopezgi , given the underlying issue is now 2 years old and no other pull requests have come forward, perhaps the solution from https://github.com/bazelbuild/rules_docker/pull/703 can be implemented?

PR #703, opened 2019-03-01
Fix --norun option when args are specified for _image rules

mterring:
When an _image rule has args specified these are automatically passed through when running via `bazel run`, which breaks the "single argument" test to skip running the container. In our case this broke CI.

### DIFF
--- a/container/incremental_load.sh.tpl
+++ b/container/incremental_load.sh.tpl
@@ -234,8 +234,8 @@ function read_variables() {
 %{tag_statements}
 
 # An optional "docker run" statement for invoking a loaded container.
-# This is not executed if the single argument --norun is passed.
-if [ "a$*" != "a--norun" ]; then
+# This is not executed if the argument --norun is passed.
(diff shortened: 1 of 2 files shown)


xingao267:
Hi @mterring thanks for the PR. Can you add a test case for this?

mterring:
Test added

alex1545:
Hi @mterring. Thank you for sending this PR. An issue related to the same problem was opened earlier here #665. Your solution and the one that checks if all the args end with "--norun" were considered, but are not desirable. Such solutions will misbehave if your lang_image target specifies args that include "--norun". It will also make it harder in the future if we decide to support additional flags. Could you please instead try and implement what's suggested in this [comment](https://github.com/bazelbuild/rules_docker/issues/665#issuecomment-455259228).

mterring:
No problem, not sure I'll have time to provide the alternate implementation as we will likely stop using the language images for other reasons (the directory structure).

PR #707, opened 2019-03-04
 update bazel-toolchains and add rbe_autoconfig target

nlopezgi:
For now we will have the digests here, but left a TODO to move them to some pin file (probably in bazel-toolchains repo) so that its easier to update.


WORKSPACE
```diff
@@ -278,11 +278,21 @@ load(
...
+  name = "bazel_toolchains",
+  sha256 = "4b1468b254a572dbe134cc1fd7c6eab1618a72acd339749ea343bd8f55c3b7eb",
+  strip_prefix = "bazel-toolchains-d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4",
+  urls = [
+    "https://mirror.bazel.build/github.com/bazelbuild/bazel-toolchains/archive/d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4.tar.gz",
+    "https://github.com/bazelbuild/bazel-toolchains/archive/d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4.tar.gz",
+  ],
+)
+
+# TODO(nlopezgi): use pinned versions for digests once the container is made public
```
xingao267: use versions from a pin file once the container ....

nlopezgi: fixed

.bazelci/presubmit.yml
```diff
@@ -63,10 +63,15 @@ platforms:
...
+    - "-//testdata:pkg_control_xz"
+    - "-//tests/docker:deb_image_with_dpkgs"
+    - "-//tests/docker:deb_image_with_dpkgs_test"
+    - "-//tests/docker:test_digest_output1"
     build_flags:
-    - "--extra_execution_platforms=@bazel_toolchains//configs/ubuntu16_04_clang/1.1:nosla_xenial_docker"
-    - "--host_platform=@bazel_toolchains//configs/ubuntu16_04_clang/1.1:nosla_xenial_docker"
-    - "--platforms=@bazel_toolchains//configs/ubuntu16_04_clang/1.1:nosla_xenial_docker"
+    - "--extra_execution_platforms=@bazel_toolchains//configs/ubuntu16_04_clang/1.2:nosla_xenial_docker"
+    - "--host_platform=@bazel_toolchains//configs/ubuntu16_04_clang/1.2:nosla_xenial_docker"
```
xingao267: You can probably use the latest docker_platform and toolchain alias here as well https://github.com/bazelbuild/bazel-toolchains/blob/master/.bazelci/presubmit.yml#L20

WORKSPACE
```diff
@@ -279,10 +279,23 @@ _nodejs_image_repos()
...
+    strip_prefix = "bazel-toolchains-d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4",
     urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/bazel-toolchains/archive/ad7f0157e13d5d3a4b893652a375b03ee9e032b4.tar.gz",
-        "https://github.com/bazelbuild/bazel-toolchains/archive/ad7f0157e13d5d3a4b893652a375b03ee9e032b4.tar.gz",
+        "https://mirror.bazel.build/github.com/bazelbuild/bazel-toolchains/archive/d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4.tar.gz",
+        "https://github.com/bazelbuild/bazel-toolchains/archive/d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4.tar.gz",
     ],
 )
+
+load("@bazel_toolchains//rules:rbe_repo.bzl", "rbe_autoconfig")
```
xingao267: load statements should be at top

.bazelci/presubmit.yml
nlopezgi: fixed

WORKSPACE
nlopezgi: Not in WORKSPACE; I dont think you can load before the http_archive rule that imports bazel_toolchains?

xingao267: got it

PR #710, opened 2019-03-05
Fix buildifier 0.22.0 warnings in files matching *image.bzl

smukherj1:
Still get

./container/image.bzl:342: uninitialized: Variable "layer" may not have been initialized. (https://github.com/bazelbuild/buildtools/blob/master/WARNINGS.md#uninitialized)

which seems to be a bug in buildifier because the loop variable in a
list comprehension is being reported as potentially uninitialized.

Fixes involve adding function, argument, return statement and module
docs.

### DIFF
--- a/container/layer_tools.bzl
+++ b/container/layer_tools.bzl
@@ -17,6 +17,10 @@ load(
     "//skylib:path.bzl",
     _get_runfile_path = "runfile",
 )
+load(
+    "//container:providers.bzl",
+    "ImageInfo",
+)
 
 def _extract_layers(ctx, name, artifact):
     config_file = ctx.actions.declare_file(name + "." + artifact.basename + ".config")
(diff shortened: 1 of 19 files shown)


contrib/rename_image.bzl
```diff
@@ -13,6 +13,8 @@
 # limitations under the License.
 
 # Migrated from https://github.com/GoogleContainerTools/base-images-docker/blob/4f2fc8da248a61c3f8e13bbb43e9db6c0ed44ba3/util/run.bzl#L264
+"""A rule to rename the image from a <lang>_image or container_image target.
+"""
```
xingao267: Can you also put the "Migrated from ..." to be in the body of the docstring?

container/layer_tools.bzl
```diff
@@ -17,6 +17,10 @@ load(
     "//skylib:path.bzl",
     _get_runfile_path = "runfile",
 )
+load(
```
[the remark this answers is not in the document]
smukherj1: Removed

contrib/rename_image.bzl
smukherj1: Done

PR #711, opened 2019-03-06
Publish an image.digest file for container_pull.

stevewolter:
The container_build rules publish "foo.digest" for a container "foo".
container_pull names the container "image" and only publishes "digest",
not "image.digest", so far.


stevewolter:
@nlopezgi Thanks for the pointers, they made it a lot simpler! Done, could you please take another look?

PR #724, opened 2019-03-09
Cleanups

ash2k:
See individual commits.

### DIFF
--- a/contrib/test.bzl
+++ b/contrib/test.bzl
@@ -40,8 +40,8 @@ def _impl(ctx):
         args += ["--image", ctx.attr.loaded_name]
         runfiles = ctx.runfiles(
             files = [ctx.executable._structure_test, ctx.executable.image] + ctx.files.configs,
-            transitive_files = ctx.attr.image.files,
-        ).merge(ctx.attr.image.data_runfiles)
+            transitive_files = ctx.attr.image[DefaultInfo].files,
+        ).merge(ctx.attr.image[DefaultInfo].data_runfiles)
 
     if not ctx.attr.verbose:
         args += ["--quiet"]
--- a/contrib/test.bzl
+++ b/contrib/test.bzl
@@ -61,9 +61,7 @@ def _impl(ctx):
         is_executable = True,
     )
 
-    return struct(
-        runfiles = runfiles,
-    )
+    return [DefaultInfo(runfiles = runfiles)]
 
 _container_test = rule(
     attrs = {
--- a/java/image.bzl
+++ b/java/image.bzl
@@ -114,18 +114,31 @@ def java_files(f):
     Returns:
         Depset of java source files.
     """
-    files = depset()
+    files = []
+
     if java_common.provider in f:
         java_provider = f[java_common.provider]
-        files = depset(transitive = [files, java_provider.transitive_runtime_jars])
-    if hasattr(f, "files"):  # a jar file
-        files = depset(transitive = [files, f.files])
-    return files
+        files.append(java_provider.transitive_runtime_jars)
+
+    f_files = f[DefaultInfo].files
+    if f_files != None:
+        files.append(f_files)
+
+    return depset(transitive = files)
 
 def java_files_with_data(f):
+    """Filter out the list of java source and data files from the given list of runfiles.
+
+    Args:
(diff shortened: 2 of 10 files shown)


java/image.bzl
```diff
@@ -114,19 +114,20 @@ def java_files(f):
     Returns:
         Depset of java source files.
     """
-    files = depset()
+    files = []
+
     if java_common.provider in f:
         java_provider = f[java_common.provider]
-        files = depset(transitive = [files, java_provider.transitive_runtime_jars])
-    if hasattr(f, "files"):  # a jar file
```
xingao267: We probably should still keep the check for legacy providers since it hasn't been removed?

Also why changing it back to List? I think it's preferred to use depset. I think it's better for data accumulation like this. https://docs.bazel.build/versions/master/skylark/lib/depset.html

java/image.bzl
```diff
@@ -114,19 +114,20 @@ def java_files(f):
...
+        files.append(f_files)
+
+    return depset(transitive = files)
 
 def java_files_with_data(f):
-    files = java_files(f)
-    if hasattr(f, "data_runfiles"):
-        files = depset(transitive = [files, f.data_runfiles.files])
-    return files
+    return depset(transitive = [java_files(f), f[DefaultInfo].data_runfiles.files])
```
xingao267: why not checking for f[DefaultInfo].data_runfiles.files not None here?

contrib/test.bzl
```diff
@@ -22,7 +22,7 @@ load(
     "container_bundle",
 )
 
-def _impl(ctx):
+def _container_test_impl(ctx):
```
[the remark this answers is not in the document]
ash2k: I will revert this for consistency with other files. The idea was to follow the convention where the implementing function is called `_<rule_name>_impl`. It's documented here https://docs.bazel.build/versions/master/skylark/rules.html#implementation-function

java/image.bzl
ash2k: Every rule type (even those using the legacy `struct()` instead of providers) expose the `DefaultInfo` provider, so this should be a safe change. See https://docs.bazel.build/versions/master/skylark/lib/DefaultInfo.html

It is preferred to use the depset, I'm only changing how it is constructed. `files` is a list of depsets and then on the last line a new depset is constructed from the list of depsets.

java/image.bzl
ash2k: There are many other places where this needs to be checked. I thought I'll do it in a separate commit. I've also left a comment https://github.com/bazelbuild/bazel/issues/3015#issuecomment-471513538 to see if `None` can be removed so that we don't need to have these checks.

contrib/test.bzl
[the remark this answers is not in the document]
xingao267: Thanks for the pointer. I'll double check with bazel owners. I think conventionally it's usually named _impl and the example given in the same section uses it too. It's not well documented and caused confusion.

java/image.bzl
xingao267: Ack. Thanks for the explanation.

java/image.bzl
xingao267: I think let's still do a check of `f[DefaultInfo].data_runfiles` being not `None` in this PR to leaving this at a safe state.

ash2k: I can do that, np. But there are many other places in code where this check is not performed. Do you want me to fix them too?

xingao267: That would be very appreciated! But they don't have to be in this PR.

java/image.bzl
```diff
@@ -114,18 +114,31 @@ def java_files(f):
...
+    Args:
+       f: Runfiles for a java_image rule.
+
+    Returns:
+       Depset of java source and data files.
+    """
     files = java_files(f)
-    if hasattr(f, "data_runfiles"):
-        files = depset(transitive = [files, f.data_runfiles.files])
+    runfiles = f[DefaultInfo].data_runfiles.files
```
xingao267: I think you wanna check `f[DefaultInfo].data_runfiles` is not None, otherwise f[DefaultInfo].data_runfiles.files will complain None type does not have `files` attribute.

ash2k: oh, yeah. Checked the wrong thing.

Issue #690, opened 2019-02-12
Issue compiling go_image on macOS


obeattie:
I am trying to compile a `go_image` target on macOS, and this is failing with an error about missing imports:

```
$ bazel build //service.api.ping:image
(20:41:03) INFO: Invocation ID: f27684c3-962f-4aba-a5d7-57245f4da640
(20:41:03) INFO: Current date is 2019-02-12
(20:41:03) INFO: Analysed target //service.api.ping:image (0 packages loaded, 0 targets configured).
(20:41:03) INFO: Found 1 target...
(20:41:04) ERROR: /private/var/tmp/_bazel_oliver/40333f6e8ec32396acff18e2ad143eea/external/org_golang_google_grpc/internal/syscall/BUILD.bazel:3:1: GoCompile external/org_golang_google_grpc/internal/syscall/linux_amd64_pure_stripped/go_default_library%/google.golang.org/grpc/internal/syscall.a failed (Exit 1) compile failed: error executing command bazel-out/host/bin/external/io_bazel_rules_go/go/tools/builders/darwin_amd64_stripped/compile -sdk external/go_sdk -installsuffix linux_amd64 -src ... (remaining 14 argument(s) skipped)

Use --sandbox_debug to see verbose messages from the sandbox
GoCompile: missing strict dependencies:
	/private/var/tmp/_bazel_oliver/40333f6e8ec32396acff18e2ad143eea/sandbox/darwin-sandbox/1/execroot/wearedev/external/org_golang_google_grpc/internal/syscall/syscall_linux.go: import of "golang.org/x/sys/unix"
Known dependencies are:
	google.golang.org/grpc/grpclog
Check that imports in Go sources match importpath attributes in deps.
Target //service.api.ping:image failed to build
Use --verbose_failures to see the command lines of failed build steps.
(20:41:04) INFO: Elapsed time: 0.600s, Critical Path: 0.20s
(20:41:04) INFO: 2 processes: 2 darwin-sandbox.
(20:41:04) FAILED: Build did NOT complete successfully
```

The BUILD file that I am building this from specifies `goarch` and `goos`, which I think is as you recommend it:

```py
load("@io_bazel_rules_docker//go:image.bzl", "go_image")
load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")

go_library(
    deps = [
        "//bedrock:go_default_library",
        "//bedrock/filter:go_default_library",
        "//service.api.ping/handler:go_default_library",
    ],
[...]

PR #730, opened 2019-03-13
Go image doc update

alex1545:
To address confusion in #690.
Prefer using the `--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64` flag over `goos`, `goarch` and `pure` attrs for `go_image` targets as recommended by go_binary docs [here](https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#attributes-2) and [here](https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#note-on-goos-and-goarch-attributes).

### DIFF
--- a/README.md
+++ b/README.md
@@ -577,14 +577,25 @@ go_image(
     name = "go_image",
     srcs = ["main.go"],
     importpath = "github.com/your/path/here",
+)
+```
+Notice that it is important to explicitly build this target with the
+`--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 ` flag
+as the binary should be built for Linux since it will run in a Linux container.
+Only in situations when the `--platforms` flag does not work, try specifying
+the `goarch`, `goos` and `pure` flags as follows:
+```python
+load("@io_bazel_rules_docker//go:image.bzl", "go_image")
+
+go_image(
+    name = "go_image",
+    srcs = ["main.go"],
+    importpath = "github.com/your/path/here",
     goarch = "amd64",
     goos = "linux",
     pure = "on",
 )
 ```
-
-Notice that it is important to explicitly specify `goarch`, `goos`, and `pure`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -573,6 +573,20 @@ following import:
...
+go_image(
+    name = "go_image",
+    srcs = ["main.go"],
+    importpath = "github.com/your/path/here",
+)
+```
+Notice that it is important to explicitly build this target with the
+`--platforms=@io_bazel_rules_go//go/toolchain:linux_amd64` flag
+as the binary should be built for Linux since it will run in a Linux container.
+Only in situations when the `--platforms` flag does not work, try specifying
```
nlopezgi: do we know now in which situations this does not work? If so lets document them briefly here.

alex1545: I'm actually unaware of any specific cases when the `--platforms` flag will not work.

@jayconrod, from the last part of [your comment](https://github.com/bazelbuild/rules_docker/issues/690#issuecomment-462945896), did you have any specific cases in mind as of when the `--platforms` flag can't be used to cross-compile with rules_go and using the `goos` and `goarch` attrs is inevitable?

From this [doc](https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#attributes-2) about the `goos` and `goarch` attributes it says that:
> This attribute has several limitations and should only be used in situations where the --platforms flag does not work. 

But doesn't mention any cases when the `--platforms` flag may not work.

jayconrod: The main situation where `--platforms` doesn't work is where a rule needs to target a specific platform that is different from Bazel's target platform. That mainly comes up in cases where `go_binary` is used in another rule, especially when binaries for multiple platforms are needed.

This use case should be handled by Starlark transitions, but I don't know if they're completely supported yet in Bazel, and rules_go will certainly need an update to take advantage of them.

alex1545: Thank you very much for the explanation.

PR #731, opened 2019-03-13
Add error message when docker is not available in incremental loader

clintharrison:
Right now, when `docker` is not found through the default-configured toolchain, the error message ends up being quite opaque:

```my/target: line 202: : command not found```


smukherj1:
@nlopezgi Can you take a look? Overall this makes sense to me because we don't emit any error messages when building containers if docker isn't found. However, using bazel run on the built container will result in an obscure error message as @clintharrison pointed out. It makes sense to give a more visible error message if the docker path in the incremental loader shell script is blank.

nlopezgi:
thanks, I'll take a look early next week

PR #733, opened 2019-03-14
Use a single url in the "urls" field for bazel-toolchains

smukherj1:
The intention is to get renovate to start picking this up

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -289,10 +289,7 @@ http_archive(
     name = "bazel_toolchains",
     sha256 = "4b1468b254a572dbe134cc1fd7c6eab1618a72acd339749ea343bd8f55c3b7eb",
     strip_prefix = "bazel-toolchains-d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4",
-    urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/bazel-toolchains/archive/d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4.tar.gz",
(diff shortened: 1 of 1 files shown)


WORKSPACE
```diff
@@ -289,10 +289,7 @@ http_archive(
     name = "bazel_toolchains",
     sha256 = "4b1468b254a572dbe134cc1fd7c6eab1618a72acd339749ea343bd8f55c3b7eb",
     strip_prefix = "bazel-toolchains-d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4",
-    urls = [
-        "https://mirror.bazel.build/github.com/bazelbuild/bazel-toolchains/archive/d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4.tar.gz",
-        "https://github.com/bazelbuild/bazel-toolchains/archive/d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4.tar.gz",
-    ],
+    urls = ["https://github.com/bazelbuild/bazel-toolchains/archive/d665ccfa3e9c90fa789671bf4ef5f7c19c5715c4.tar.gz"],
```
alex1545: I thought renovate won't update URLs when you specify it as a list (even when the list contains one entry). But I believe it works when you use the `url` attribute instead.

smukherj1: Not really, here is a case using urls (https://github.com/bazelbuild/rules_k8s/pull/278/files) that worked fine. So I'm taking this pattern that worked and trying to replicate it. Basically the renovate detection logic seems flaky

Issue #714, opened 2019-03-07
Wrong image digest pulled for google-containers/debian-iptables


drigz:
With the latest HEAD, I get strange behavior when pulling debian-iptables:

WORKSPACE:
```
load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")

git_repository(
    name = "io_bazel_rules_docker",
    commit = "b167423d6bcc0ee69e021432d8fc4ab7305b6649",
    remote = "https://github.com/bazelbuild/rules_docker",
)

load(
    "@io_bazel_rules_docker//repositories:repositories.bzl",
    container_repositories = "repositories",
)

container_repositories()

load(
    "@io_bazel_rules_docker//container:container.bzl",
    "container_pull",
)

container_pull(
    name = "iptables_base",
    digest = "sha256:cd81b1a8f40149b5061735927d2a2cf4b90fc27a52fc4cc66889b373368b6ef6",
    registry = "gcr.io",
    repository = "google-containers/debian-iptables",
)
```

Note the SHA256 for iptables_base: cd81b1...

```
> bazel build @iptables_base//image:all
DEBUG: Rule 'io_bazel_rules_docker' modified arguments {"shallow_since": "1551974828 -0500"}
DEBUG: Rule 'iptables_base' modified arguments {"digest": "sha256:d4ff8136b9037694a3165a7fff6a91e7fc828741b8ea1eda226d4d9ea5d23abb"}
INFO: Analysed target @iptables_base//image:image (26 packages loaded, 247 targets configured).
INFO: Found 1 target...
INFO: Writing explanation of rebuilds to '/tmp/explain.log'
Target @iptables_base//image:image up-to-date:
  bazel-bin/external/iptables_base/image/image.tar
INFO: Elapsed time: 6.570s, Critical Path: 1.64s
INFO: 9 processes: 9 linux-sandbox.
> cat bazel-*/../../external/iptables_base/image/image.digest       
sha256:d4ff8136b9037694a3165a7fff6a91e7fc828741b8ea1eda226d4d9ea5d23abb
```

For some reason, it has pulled a different image than desired, as seen in `DEBUG: Rule 'iptables_base' modified arguments {"digest": "sha256:d4ff81...` and the content of `image.digest`.

nlopezgi:
does this repro with a file_test similar to the one we added in https://github.com/bazelbuild/rules_docker/commit/b167423d6bcc0ee69e021432d8fc4ab7305b6649.
[...]

PR #734, opened 2019-03-14
Fail when pulled image with unexpected sha256

alex1545:
To address confusion in #714 when pinning to a manifest list image type.
This forces the user to only pin to a manifest type when using digest.

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -83,14 +83,6 @@ container_pull(
 )
 
 container_pull(
-    name = "k8s_pause_amd64",
-    # this is a manifest list, so the resolved digest should not match this digest
-    digest = "sha256:f78411e19d84a252e53bff71a4407a5686c46983a2c2eeed83929b888179acea",
-    registry = "k8s.gcr.io",
-    repository = "pause",
-)
-
-container_pull(
     name = "k8s_pause_arm64",
     architecture = "arm64",
     registry = "k8s.gcr.io",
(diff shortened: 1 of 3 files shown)


container/pull.bzl
```diff
@@ -191,6 +191,14 @@ exports_files(["image.digest", "digest"])
         fail("Failed to read digest: %s" % digest_result.stderr)
     updated_attrs["digest"] = digest_result.stdout
 
+    if repository_ctx.attr.digest and repository_ctx.attr.digest != updated_attrs["digest"]:
+        fail(("SHA256 of the image specified does not match SHA256 of the pulled image. " +
+              "Expected {}, but pulled image with {}. " +
+              "It is possible that you pin to a manifest list which points to another image").format(
```
nlopezgi: It is possible that you have a pin to a ...
which points to another image, if so, ... <provide advice as to how to solve>

WORKSPACE
```diff
@@ -82,14 +82,6 @@ container_pull(
     tag = "3.8",
 )
 
-container_pull(
-    name = "k8s_pause_amd64",
-    # this is a manifest list, so the resolved digest should not match this digest
-    digest = "sha256:f78411e19d84a252e53bff71a4407a5686c46983a2c2eeed83929b888179acea",
```
nlopezgi: lets try to leave this test but remove the digest attr?

alex1545: Removing the digest attr here will then use the default "latest" tag, which will be hard to test with `file_test`.
And we already have a test that relies on a tag (one target below called `k8s_pause_arm64`).
Let me know if you still want to keep this `container_pull` target without the digest attr and leave it untested.

nlopezgi: sgtm

PR #737, opened 2019-03-15
Repo rule to build an image from Dockerfile and save as tar

alex1545:
Repo rule that uses the docker binary installed in the system to build a docker image from a given Dockerfile. It then saves that image as a tar file.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -83,6 +83,7 @@ platforms:
     - "-//testdata:pkg_control_xz.deb"
     - "-//testdata:pkg_control_xz"
     # Disabled as docker is not available
+    # TODO(alex1545): enable once docker is supported on buildkite
     - "-//tests/docker:basic_dockerfile_image"
     build_flags:
     - "--extra_toolchains=@bazel_toolchains//configs/ubuntu16_04_clang/latest:toolchain_docker"
@@ -115,6 +116,7 @@ platforms:
     - "-//testdata:pkg_control_xz.deb"
     - "-//testdata:pkg_control_xz"
     # Disabled as docker is not available
+    # TODO(alex1545): enable once docker is supported on buildkite
     - "-//tests/docker:basic_dockerfile_image"
     test_flags:
     - "--extra_toolchains=@bazel_toolchains//configs/ubuntu16_04_clang/latest:toolchain_docker"
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -82,6 +82,8 @@ platforms:
     - "-//tests/docker:deb_image_with_dpkgs"
     - "-//testdata:pkg_control_xz.deb"
     - "-//testdata:pkg_control_xz"
+    # Disabled as docker is not available
+    - "-//tests/docker:basic_dockerfile_image"
     build_flags:
     - "--extra_toolchains=@bazel_toolchains//configs/ubuntu16_04_clang/latest:toolchain_docker"
     - "--extra_execution_platforms=@bazel_toolchains//configs/ubuntu16_04_clang/latest:platform_docker"
@@ -112,6 +114,8 @@ platforms:
     - "-//tests/docker:deb_image_with_dpkgs"
     - "-//testdata:pkg_control_xz.deb"
     - "-//testdata:pkg_control_xz"
+    # Disabled as docker is not available
+    - "-//tests/docker:basic_dockerfile_image"
     test_flags:
     - "--extra_toolchains=@bazel_toolchains//configs/ubuntu16_04_clang/latest:toolchain_docker"
     - "--extra_execution_platforms=@bazel_toolchains//configs/ubuntu16_04_clang/latest:platform_docker"
--- a/contrib/dockerfile_build.bzl
+++ b/contrib/dockerfile_build.bzl
@@ -21,6 +21,11 @@ use (e.g. container_image, container_load, container_test)
 
 The built docker image is always called after the dockerfile_image target used
 to build the image and tagged with `dockerfile_image` (i.e. <target_name>:dockerfile_image).
+This means that if an image with name <target_name>:dockerfile_image already
+exists and a dockerfile_image target called <target_name> is built, then the
+existing image will change its name to <none>:<none> and the newly built image
+will get the <target_name>:dockerfile_image name.
+
 The produced tar file is available at @<repo_name>//image:dockerfile_image.tar
 """
 
@@ -50,24 +55,17 @@ def _impl(repository_ctx):
     # file with repository_ctx.read() and repository_ctx.file() instead of
     # executing the `cp` command
     # dockerfile_content = repository_ctx.read(repository_ctx.attr.dockerfile)
-    # repository_ctx.file("Dockerfile", dockerfile_content)
-
-    dockerfile_src = repository_ctx.path(repository_ctx.attr.dockerfile)
-    dockerfile_dest = repository_ctx.path("Dockerfile")
 
     # Copy the provided Dockerfile into the root of this workspace since using
     # the -f flag with the docker build command is problematic.
-    cp_result = repository_ctx.execute([
-        "cp",
-        dockerfile_src,
-        dockerfile_dest,
+    cat_result = repository_ctx.execute([
+        "cat",
+        repository_ctx.path(repository_ctx.attr.dockerfile),
     ])
-    if cp_result.return_code:
-        fail("Failed to copy Dockerfile from {} to {}: {}".format(
-            dockerfile_src,
-            dockerfile_dest,
-            cp_result.stderr,
-        ))
+    if cat_result.return_code:
+        fail("Failed to read Dockerfile content: {}".format(cat_result.stderr))
+    dockerfile_content = cat_result.stdout
+    repository_ctx.file("Dockerfile", dockerfile_content)
 
     img_name = repository_ctx.name + ":dockerfile_image"
     build_args = [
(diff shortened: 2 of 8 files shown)


docker/dockerfile_build.bzl
```diff
@@ -0,0 +1,113 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
nlopezgi: Lets move this rule to contrib,
to indicate its still experimental

docker/dockerfile_build.bzl
```diff
@@ -0,0 +1,113 @@
...
+   into a tar file.
+
+This rule uses the docker tool installed in the system to build the image and
+then save it as a tar file. The saved tar file is available for other rules to
+use (e.g. container_image, container_load, container_test)
+
+The produced tar file is available at @<repo_name>//image:dockerfile_image.tar
+"""
+
+def docker(repository_ctx):
```
nlopezgi: def _docker

docker/dockerfile_build.bzl
```diff
@@ -0,0 +1,113 @@
...
+    """Core implementation of dockerfile_image."""
+
+    docker_path = docker(repository_ctx)
+
+    dockerfile_src = repository_ctx.path(repository_ctx.attr.dockerfile)
+    dockerfile_dest = repository_ctx.path("Dockerfile")
+
+    # Copy the provided Dockerfile into the root of this workspace since using
+    # the -f flag with the docker build command is problematic.
+    cp_result = repository_ctx.execute([
```
nlopezgi: use repository_ctx.read to read repository_ctx.attr.dockerfile
and use repository_ctx.file to create the Dockerfile. https://docs.bazel.build/versions/master/skylark/lib/repository_ctx.html#file

docker/dockerfile_build.bzl
```diff
@@ -0,0 +1,113 @@
...
+        dockerfile_dest,
+    ])
+    if cp_result.return_code:
+        fail("Failed to copy Dockerfile from {} to {}: {}".format(
+            dockerfile_src,
+            dockerfile_dest,
+            cp_result.stderr,
+        ))
+
+    img_name = repository_ctx.name + ":dockerfile_image"
```
nlopezgi: document somewhere that this will be the produced image name?

docker/dockerfile_build.bzl
```diff
@@ -0,0 +1,113 @@
...
+    build_args = [
+        docker_path,
```
nlopezgi: We need to add arguments to the docker call so that it DOES not use the docker cache (i.e., it always builds the full image using the DOCKERFILE from scratch) (Pointer I found: https://stackoverflow.com/questions/35594987/how-to-force-docker-for-a-clean-build-of-an-image)

docker/dockerfile_build.bzl
```diff
@@ -0,0 +1,113 @@
...
+        "build",
+        "-t",
+        img_name,
```
nlopezgi: what would happen if an image had already been created with this name previously?

docker/dockerfile_build.bzl
```diff
@@ -0,0 +1,113 @@
...
+        repository_ctx.path(""),
+    ]
+    build_result = repository_ctx.execute(build_args)
+    if build_result.return_code:
+        fail("docker build command failed: {} ({})".format(
+            build_result.stderr,
+            " ".join(build_args),
+        ))
+
+    image_tar = "dockerfile_image.tar"
```
nlopezgi: document somewhere the image tar will be produced with this name

docker/dockerfile_build.bzl
alex1545: Done

docker/dockerfile_build.bzl
alex1545: Done

docker/dockerfile_build.bzl
alex1545: `repository_ctx.read` doesn't seem to be available in Bazel 0.23.2 (maybe it's only available at head?). I added a TODO to make this change once the `read()` function is available.

docker/dockerfile_build.bzl
alex1545: Added this to the top level comment.

docker/dockerfile_build.bzl
alex1545: Added the `--no-cache` flag.

docker/dockerfile_build.bzl
alex1545: This was documented in the top level comment.

docker/dockerfile_build.bzl
alex1545: I thought about this problem too because we don't want to overwrite any images that could be in use. But to answer the question, with the `--no-cache` flag, the old image called `name:tag` will become `<none>:<none>` and the newly built image will become `name:tag`. To decrease the chances of hitting an unrelated image (and changing its name to `<none>:<none>`) I have the `dockerfile_image` tag. But we can probably do better by putting a time stamp in the tag instead. This will lower the chances, but will still not guarantee that we won't hit an unwanted image.

contrib/dockerfile_build.bzl
```diff
@@ -0,0 +1,122 @@
...
+        return repository_ctx.which("docker")
+
+    fail("Path to the docker tool was not provided and it could not be resolved automatically.")
+
+def _impl(repository_ctx):
+    """Core implementation of dockerfile_image."""
+
+    docker_path = _docker(repository_ctx)
+
+    # TODO(alex1545): once the read() function is available, enable copying
```
nlopezgi: use cat then to read the contents, then use repository_ctx.file() to write them. Example cat use: https://github.com/bazelbuild/rules_docker/blob/master/container/pull.bzl#L189

contrib/dockerfile_build.bzl
```diff
@@ -0,0 +1,122 @@
...
+   into a tar file.
+
+This rule uses the docker tool installed in the system to build the image and
+then save it as a tar file. The saved tar file is available for other rules to
+use (e.g. container_image, container_load, container_test)
+
+The built docker image is always called after the dockerfile_image target used
+to build the image and tagged with `dockerfile_image` (i.e. <target_name>:dockerfile_image).
+The produced tar file is available at @<repo_name>//image:dockerfile_image.tar
+"""
```
nlopezgi: Add note about expected behavior if an image with name <target_name>:dockerfile_image already exists.

alex1545: Done

docker/dockerfile_build.bzl
alex1545: Using `cat` command instead of `repository_ctx.read()` for now. Left the TODO here for now.

contrib/dockerfile_build.bzl
alex1545: Done

docker/dockerfile_build.bzl
alex1545: Added comment with what to expect in this case.

alex1545:
If we don't mind adding dockerfile_image targets into our main WORKSPACE file, then it's easy to add a test in this PR. If we want to test this rule in testing/e2e.sh then it's easier to add a test in #740.

nlopezgi:
lets add a dockerfile_image target in our main WORKSPACE, just for testing purposes so that this runs with buildkite CI not with e2e tests

.bazelci/presubmit.yml
```diff
@@ -82,6 +82,8 @@ platforms:
     - "-//tests/docker:deb_image_with_dpkgs"
     - "-//testdata:pkg_control_xz.deb"
     - "-//testdata:pkg_control_xz"
+    # Disabled as docker is not available
```
nlopezgi: Add TODO to enable once docker is supported on buildkite

testing/e2e.sh
```diff
@@ -522,6 +522,12 @@ function test_container_pull_cache() {
   rm -rf $scratch_dir
 }
 
+function test_dockerfile_image_basic() {
```
nlopezgi: Add todo to remove from here and move to buildkite once docker is supported

.bazelci/presubmit.yml
alex1545: Done

testing/e2e.sh
alex1545: Done

nlopezgi:
they are just examples for now. And yes, they are not hermetic, implemented as repo rules to reduce the chance of issues, and using the --no_cache docker flag to further try to avoid issues.

docker/dockerfile_build.bzl
f355: Sorry for bumping an old thread, but I want to understand the reasoning behind not using the docker cache. I was under impression that docker builds are hermetic, since they are cached based on the file content hashes. Is that not true?

alex1545: One big reason is that a Dockerfile allows specifying arbitrary commands inside the RUN instruction. This implies that the command under the RUN instruction may produce different results when running multiple times, but using the docker cache will prevent a new image to be built. For example, you could be installing some apt package foo with
`apt-get install foo`
and that package gets updated in 5 mins. You rebuild your image, but won't get the updated package since your layers will be cached.

f355: Thanks, that makes sense. I wouldn't mind having some way to control this behavior though, it's a pain in the neck when developing stuff locally.

nlopezgi: @f355, thanks for your comments. The big argument against using the Docker cache is that Bazel has its own caching mechanism. Getting the two to play along nicely is not possible afaik. The most likely result is very frustrating situations where even running `bazel clean --expunge_async` yields non-hermetic / non-deterministic behavior.
If by 'pain in the neck' you mean that stuff runs a bit slowly, I'd take that every day vs having users complain about hermeticity problems (i.e., why does the test run in my machine but fails in yours?). We've seen this pain with docker images and wish it upon noone. If you feel building with a Dockerfile is too slow, our advice is to start migrating it to rules_docker to get better performance.

PR #747, opened 2019-03-20
Create empty workspace directory to fix using python deps as layers

### LINKED ISSUE #161: py_image deps/layers discrepancy.
If [this line](https://github.com/bazelbuild/rules_k8s/blob/93bf3ead1904a3b0e5c31312718f2aa98abbaac2/examples/hellohttp/py/BUILD#L25) is changed to `layers=[]` instead of `deps=[]` then the image no longer works properly.

smukherj1:
This fixes #161


testdata/test/py_image_using_layers.py
```diff
@@ -0,0 +1,10 @@
+"""Ensure the "six" module can be imported.
+
+Used to end to end test python dependencies to py_image work when specified
+as layers. https://github.com/bazelbuild/rules_docker/issues/161
+"""
+
+import six
+
+if __name__ == "__main__":
+    print("Successfully imported {} {}.".format(six.__name__, six.__version__))
```
xingao267: empty line at the end

smukherj1: Done

testing/e2e.sh
```diff
@@ -131,7 +131,10 @@ EOF
 
 
 function clear_docker() {
-  docker rmi -f $(docker images -aq) || true
+  # Get the IDs of images except the local registry image "registry:2". We only
```
xingao267: where is this registry:2 coming from?

smukherj1: Used in lines 396 and 406 to run a local docker registry to test container push/pull with

xingao267: But here you are just excluding registry:2, and you will still delete images that are not created in the tests iiuc?

smukherj1: This is just a minor upgrade from what this code was doing before. Previously it would delete *everything*. With this, I save the registry image which is used in more than one test to avoid having to pull it more than once. It's not perfect but slightly better than before.

xingao267: That sgtm, but we should just be more clear in the comments then.

smukherj1: Fixed

PR #749, opened 2019-03-20
Extended dockerfile example

alex1545:
Another example of a Dockerfile and its equivalent container_image target.
Includes tests to verify equivalence of images built in different ways.


testing/examples/extended/BUILD
```diff
@@ -0,0 +1,52 @@
...
+        "envVar": "My environment variable",
+        "xyz": "321",
+    },
+    files = [
+        "//extended/image_data:Data_file.txt",
+        "//extended/image_data:More_data.txt",
+        "//extended/image_data:file_to_copy.txt",
+    ],
+    labels = {
+        "version": "7.7",
```
nlopezgi: run buildifier to put these in order

alex1545: Done

PR #750, opened 2019-03-20
Add dockerfile example that uses the RUN instruction

alex1545:


testing/examples/run_instruction/BUILD
```diff
@@ -0,0 +1,40 @@
+load("@base_images_docker//util:run.bzl", "container_run_and_commit")
+load("@io_bazel_rules_docker//container:container.bzl", "container_image")
+load("@io_bazel_rules_docker//contrib:test.bzl", "container_test")
+
+container_run_and_commit(
```
nlopezgi: this example should use  download and install packages, not run and commit
https://github.com/GoogleContainerTools/base-images-docker/blob/master/package_managers/download_pkgs.bzl

https://github.com/GoogleContainerTools/base-images-docker/blob/master/package_managers/install_pkgs.bzl

if you want to do run and commit here and leave download/install for another example, then dont do apt get (suggestion - maybe do pip install?)

xingao267: you can also use toolchain_container rule to achieve that which has already wrapped download_pkgs and install_pkgs.

alex1545: Changed this to use download_pkgs and install_pkgs instead.
@nlopezgi, what do you think, would a `toolchain_container` target serve a better example here?

nlopezgi: as discussed offline, lets also add a test with toolchain_container that does something more complex (e.g., requiring a gpg key to be installed)

PR #755, opened 2019-03-22
Return providers instead of struct from rule impl functions

alex1545:
This addresses the incompatible change via the ----incompatible_disallow_struct_provider_syntax flag described [here](https://github.com/bazelbuild/bazel/issues/7347).

### DIFF
--- a/container/bundle.bzl
+++ b/container/bundle.bzl
@@ -14,6 +14,7 @@
 """Rule for bundling Container images into a tarball."""
 
 load("@bazel_skylib//lib:dicts.bzl", "dicts")
+load("@io_bazel_rules_docker//container:providers.bzl", "BundleInfo")
 load(
     "//container:layer_tools.bzl",
     _assemble_image = "assemble",
@@ -21,7 +22,6 @@ load(
     _incr_load = "incremental_load",
     _layer_tools = "tools",
 )
-load("//container:providers.bzl", "BundleInfo")
 load(
     "//skylib:label.bzl",
     _string_to_label = "string_to_label",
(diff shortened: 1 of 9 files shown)


alex1545:
This is blocked by `file_test` [returning a struct](https://github.com/bazelbuild/bazel/blob/53edc2f2db90929a6610fbd09853409d398854ec/tools/build_rules/test_rules.bzl#L304) and `scala_import` [returning a struct](https://github.com/bazelbuild/rules_scala/blob/379c65e25c71ac8329f004fdd5fa7d6595558cf1/scala/scala_import.bzl#L27).

alex1545:
As discussed offline with @nlopezgi, we will merge this now, but not test with the `--incompatible_disallow_struct_provider_syntax` flag because of external dependencies that still haven't addressed this issue.

alex1545:
**IMPORTANT**
This PR changes the return value type from some rule implementation functions from `struct` to a list of providers.

No action is required if you are already using the providers returned by the below mentioned rules as described [here](https://docs.bazel.build/versions/master/skylark/rules.html#providers) (e.g. `target[ProviderName].fieldName`).

Your action may be required if you are using either the implementation function or any targets of the following rules (in your own rule implementation functions or macros):

- `container_bundle` (in container/bundle.bzl)
- `container_image` (in container/imagebzl)
- `container_import` (in container/import.bzl)
- `filter_layer` (in lang/image.bzl)

Actions Required by Rule
- `container_bundle`
If you use the `container_images` and/or the `stamp` field of the returned struct, then get these fields from the returned `BundleInfo` provider as shown [here](https://github.com/bazelbuild/rules_docker/pull/755/files#diff-5d2856be5c0de04e58eafe2848362802).
- `container_image`
If you use the `container_parts` field of the returned struct, then get this field from the returned `ImageInfo` provider as shown [here](https://github.com/bazelbuild/rules_docker/pull/755/files#diff-5e60f2badba55c27a6df9c3735b1c12c)
- `container_import`
If you use the `container_parts` field of the returned struct, then get this field from the returned `ImportInfo` provider (e.g. `target[ImportInfo].container_parts` instead of `target.container_parts`)
- `filter_layer`
If you use the `py` field of the returned struct as `target.py`, then get this field using the returned `PyInfo` provider as `target[PyInfo]`.

_Note_
If you access the returned providers with the following syntax:
`target.providers[<index>].<field>`
then you most likely need to change this to:
`target[<index>].<field>`

PR #757, opened 2019-03-25
Allow passwd entry with non-existing home path

chanseokoh:
This will be useful to create an entry like `nobody:x:65534:65534:nobody:/nonexistent:/usr/sbin/nologin`.

I'm new to Starlark, so I need some help actually.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -43,3 +43,7 @@ bazel-bin
 bazel-genfiles
 bazel-out
 bazel-testlogs
+
+# Eclipse and PyDev
+.project
+.pydevproject
--- a/container/image_test.py
+++ b/container/image_test.py
@@ -393,7 +393,7 @@ class ImageTest(unittest.TestCase):
 
   def test_with_passwd(self):
     with TestImage('with_passwd') as img:
-      self.assertDigest(img, '27149ceaab154631346209b42c9494708210901fbb6e9f88cb9370fb51f30999')
+      self.assertDigest(img, 'b5ddebd09ebfc17bee33929d65a925416739acaeefefe82d87197f112cabbb7f')
       self.assertEqual(1, len(img.fs_layers()))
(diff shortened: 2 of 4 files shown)


xingao267:
Hi @chanseokoh thanks for the PR, could you add passwd_entry, passwd_file, passwd_tar targets for this use case and tests in testdata/BUILD? Like these ones https://github.com/bazelbuild/rules_docker/blob/master/testdata/BUILD#L517

chanseokoh:
Thanks @xingao267, will do (some time later).

chanseokoh:
Added tests. Hopefully the tests pass on Travis. I couldn't run tests locally. Don't know what's wrong.

```
$ bazel test //container:image_test
INFO: Invocation ID: caf22c72-2cba-457c-baa8-f45f82348e1d
ERROR: .../rules_docker/lang/image.bzl:320:44: name 'PyInfo' is not defined (did you mean 'ApkInfo'?)
ERROR: error loading package '': in .../rules_docker/python/image.bzl: Extension 'lang/image.bzl' has errors
ERROR: error loading package '': in .../rules_docker/python/image.bzl: Extension 'lang/image.bzl' has errors
INFO: Elapsed time: 0.122s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
FAILED: Build did NOT complete successfully (0 packages loaded)
```

chanseokoh:
Never mind. Using the latest Bazel works. There seem test failures, so I'll fix them.

chanseokoh:
Tests passing. I believe the `self.assertTopLayerContains(img, ['.', './etc', './etc/password', './root', './myhomedir'])` confirms that `/nonexistent` doesn't exist in the tar??

chanseokoh:
Indeed, I verified an actual image created with `create_home = False`, and it doesn't create the directory.

PR #773, opened 2019-04-01
Add more doc about `legacy_run_behavior`.

xingao267:


container/image.bzl
```diff
@@ -473,8 +473,9 @@ _attrs = dicts.add(_layer.attrs, {
     "legacy_run_behavior": attr.bool(
         # TODO(mattmoor): Default this to False.
         default = True,
-        doc = ("If set to False, `bazel run` will directly invoke `docker run`" +
-               " with flags specified in the `docker_run_flags` attribute."),
+        doc = ("If set to False, `bazel run` will directly invoke `docker run` " +
+               "with flags specified in the `docker_run_flags` attribute. " +
+               "Note that it is default to False when using <lang>_image rules."),
```
nlopezgi: Note that it defaults to False 
or
Note that the default is set to False

PR #787, opened 2019-04-05
use python_version = PY2 for par_binary and py_binary targets

nlopezgi:
Addresses https://github.com/bazelbuild/rules_k8s/issues/305
Sending PR to kick off presubmits and also check if rules_k8s will be fixed by changing pin to this commit + adding some more  `python_version = PY2` there.

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,6 +1,23 @@
-build --python_top=//python:default_py_runtime
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


nlopezgi:
This PR still needs a bit more documentation and organizing stuff. Please hold off reviewing.

nlopezgi:
I also need to change how the default_py_runtime is generated, right now it has a hardcoded path that will not work with travisCI setup. Will need to create a repo rule that uses which to find location of py2 and py3 binaries.

nlopezgi:
closing in favor of https://github.com/bazelbuild/rules_docker/pull/826

PR #793, opened 2019-04-10
Add cleanup before docker run in incremental loader

clintharrison:
When #667 was merged, the assumption was that the cleanup handler does not do anything -- but we've since seen a lot of files left around in /tmp on our CI machines from this change.

Since we've already loaded the layers into the docker daemon before `docker run`ing, we can run the cleanup before we `exec docker`.

As an aside...
I'm slightly hesitant that the cleanup step could take a while, which means we're making the race condition from `docker tag` and a much-delayed `docker run` worse.

Maybe the `tag_reference` in the built `run_statements` should use `tag@sha256:digest` format, since we _do_ have the digest available?

### DIFF
--- a/container/incremental_load.sh.tpl
+++ b/container/incremental_load.sh.tpl
@@ -241,6 +241,10 @@ function read_variables() {
 # An optional "docker run" statement for invoking a loaded container.
 # This is not executed if the single argument --norun is passed.
 if [ "a$*" != "a--norun" ]; then
+  # Once we've loaded the images for all layers, we no longer need the temporary files on disk.
+  # We can clean up before we exec docker, since the exit handler will no longer run.
+  cleanup
+
   # This generated and injected by docker_*.
   exec %{run_statements}
 fi


smukherj1:
Hi @clintharrison,

Thanks for your PR. The change LGTM. Could you explain what you mean by

> I'm slightly hesitant that the cleanup step could take a while, which means we're making the race condition from docker tag and a much-delayed docker run worse.

Are you referring to the tag statements [here](https://github.com/bazelbuild/rules_docker/blob/d69648ea9fc10af37424ecd159578ba5614b5465/container/incremental_load.sh.tpl#L239)? I thought these statements are executed synchronously. Why would they introduce a race condition?

clintharrison:
@smukherj1 yes, exactly those: imagine the scenario

| process 1 (bazel run) | process 2 (maybe another bazel run) |
| --- | --- |
| docker tag sha256:foo bazel/some:target| |
| cleanup | docker tag sha256:bar bazel/some:target |
| (cleanup still deleting files) | ... |
| docker run bazel/some:target | ... |

At the end of this, we'd end up running a different image than we originally tagged.

Since we have the digest, we could instead do `docker run bazel/some@sha256:foo` so this isn't an issue.
Happy to send out another PR for this and have the conversation there :)

smukherj1:
@clintharrison Ok I see. I'm not sure if we actually support building the same target across different bazel invocations in the same WORKSPACE. @nlopezgi any idea what bazel does in this scenario?

Regardless, this change LGTM until we figure out if we need to fix the race condition you described

phb:
This change is causing the cleanup to run twice in the cases where there are no run_statements defined.  (second time will fail with an error since the files are already removed)

clintharrison:
Ahh, I totally see how that happens :( Let me send a follow-up PR.

PR #797, opened 2019-04-16
Dockerfile vs java_image example

alex1545:


testing/examples/java_app/image_data/Greeting.java
```diff
@@ -0,0 +1,7 @@
+// package com.example;
```
nlopezgi: add  license header

testing/examples/java_app/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM openjdk
```
nlopezgi: add brief doc

alex1545: Done

testing/examples/java_app/image_data/Greeting.java
alex1545: Done

PR #826, opened 2019-05-07
test use of incompatible_use_python_toolchains flag with Bazel 0.25.0

nlopezgi:
Changes required for use upcoming python toolchains features:
* create a `py_runtime_pair` and toolchain that is compatible with the base containers used in `py_image` and `py3_image`
* register toolchains in `deps` targets for `py_image` and `py3_image`
* Set explicitly `python_version` in `native.py_binary` calls in image.bzl for py and py3
* Adds flags to `.bazelrc` so all our CI runs with testing these fatures
  Note: flags that will only work with 0.25.0. But I tested that changes are still OK on CI if flags in bazelrc are not turned on.
* Update readme to document use of python toolchains
* extra: Use .bazelignore to avoid building anything under testing/examples/

### DIFF
--- a/README.md
+++ b/README.md
@@ -1119,13 +1119,13 @@ This can then be referenced in `BUILD` files as `@gitlab//image`.
 
 ## Python tools
 
-Starting with Bazel 0.25.0 its possible to configure python toolchains
+Starting with Bazel 0.25.0 it's possible to configure python toolchains
 for `rules_docker`.
 
 To use these features you need to enable the flags in the `.bazelrc`
 file at the root of this project.
 
-Use of these features requires a python toolchain to be registered.
+Use of these features require a python toolchain to be registered.
 `//py_images/image.bzl:deps` and `//py3_images/image.bzl:deps` register a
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -1109,6 +1117,44 @@ This can then be referenced in `BUILD` files as `@gitlab//image`.
 
 **NOTE:** This will only work on systems with Python >2.7.6
 
+## Python tools
+
+Starting with Bazel 0.25.0 its possible to configure python toolchains
```
smukherj1: its -> it's

README.md
```diff
@@ -1109,6 +1117,44 @@ This can then be referenced in `BUILD` files as `@gitlab//image`.
...
+for `rules_docker`.
+
+To use these features you need to enable the flags in the `.bazelrc`
+file at the root of this project.
+
+Use of these features requires a python toolchain to be registered.
```
smukherj1: requires -> require

README.md
nlopezgi: done

README.md
nlopezgi: done

PR #830, opened 2019-05-10
Add initial skeleton for go implementation of image_config.py

smukherj1:
The objective is to begin setting up a code structure & copybara flow to
have Go code that imports go-containerregistry in rules_docker.


WORKSPACE
```diff
@@ -216,6 +216,16 @@ go_rules_dependencies()
 
 go_register_toolchains()
 
+http_archive(
```
nlopezgi: these probably need to go to https://github.com/bazelbuild/rules_docker/blob/master/repositories/repositories.bzl so that if you pull this repo as an external repo you can get these deps too

smukherj1: As discussed offline, this is not necessary. Gazelle is only used to generate BUILD files when Go code in rules docker changes. As such the gazelle target in rules docker is of no use to repos that import rules docker.

PR #843, opened 2019-05-17
Support Python 2 and Python 3

drigz:
This should avoid users needing to set --host_force_python.

Tested with bazel 0.25.2:

bazel test --define ENV_KEY=my_key --define ENV_VALUE=my_value \
  tests/docker:all --host_force_python=PY2
bazel test --define ENV_KEY=my_key --define ENV_VALUE=my_value \
  tests/docker:all --host_force_python=PY3
bazel test --define ENV_KEY=my_key --define ENV_VALUE=my_value \
  tests/docker:all --noincompatible_use_python_toolchains

### DIFF
--- a/repositories/repositories.bzl
+++ b/repositories/repositories.bzl
@@ -79,6 +79,11 @@ def repositories():
             name = "containerregistry",
             sha256 = "a8cdf2452323e0fefa4edb01c08b2ec438c9fa3192bc9f408b89287598c12abc",
             strip_prefix = "containerregistry-" + CONTAINERREGISTRY_RELEASE[1:],
+            # containerregistry assumes a repository named @httplib2, but
+            # https://github.com/bazelbuild/bazel/issues/3998 means that this
+            # causes problems with the import path. Instead, we use
+            # repo_mapping and a fully-qualified repository name.
+            repo_mapping = {"@httplib2": "@com_github_httplib2_httplib2"},
             urls = [("https://github.com/google/containerregistry/archive/" +
                      CONTAINERREGISTRY_RELEASE + ".tar.gz")],
         )
@@ -86,19 +91,29 @@ def repositories():
     # TODO(mattmoor): Remove all of this (copied from google/containerregistry)
     # once transitive workspace instantiation lands.
 
-    if "httplib2" not in excludes:
+    if "com_github_httplib2_httplib2" not in excludes:
         # TODO(mattmoor): Is there a clean way to override?
         http_archive(
-            name = "httplib2",
+            name = "com_github_httplib2_httplib2",
             build_file_content = """
 py_library(
    name = "httplib2",
-   srcs = glob(["**/*.py"]),
-   data = ["cacerts.txt"],
+   srcs = select({
+      "@bazel_tools//tools/python:PY2": glob(["python2/httplib2/**/*.py"]),
+      "@bazel_tools//tools/python:PY3": glob(["python3/httplib2/**/*.py"]),
+   }),
+   data = select({
+      "@bazel_tools//tools/python:PY2": ["python2/httplib2/cacerts.txt"],
+      "@bazel_tools//tools/python:PY3": ["python3/httplib2/cacerts.txt"],
+   }),
+   imports = select({
+      "@bazel_tools//tools/python:PY2": ["python2"],
+      "@bazel_tools//tools/python:PY3": ["python3"],
+   }),
    visibility = ["//visibility:public"]
 )""",
             sha256 = "2dcbd4f20e826d6405593df8c3d6b6e4e369d57586db3ec9bbba0f0e0cdc0916",
-            strip_prefix = "httplib2-0.12.1/python2/httplib2/",
+            strip_prefix = "httplib2-0.12.1",
             type = "tar.gz",
             urls = ["https://codeload.github.com/httplib2/httplib2/tar.gz/v0.12.1"],
         )
@@ -138,7 +153,7 @@ py_library(
    srcs = glob(["**/*.py"]),
    visibility = ["//visibility:public"],
    deps = [
-     "@httplib2//:httplib2",
+     "@com_github_httplib2_httplib2//:httplib2",
      "@six//:six",
    ]
 )""",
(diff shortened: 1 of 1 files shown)


repositories/repositories.bzl
```diff
@@ -79,26 +79,41 @@ def repositories():
             name = "containerregistry",
             sha256 = "a8cdf2452323e0fefa4edb01c08b2ec438c9fa3192bc9f408b89287598c12abc",
             strip_prefix = "containerregistry-" + CONTAINERREGISTRY_RELEASE[1:],
+            # containerregistry assumes a repository named @httplib2, but
+            # https://github.com/bazelbuild/bazel/issues/3998 means that this
+            # causes problems with the import path. Instead, we use
+            # repo_mapping and a fully-qualified repository name.
```
brandjon: I'm not sure that using a repo mapping will avoid name conflicts. Last I looked, repo mapping affected how bazel processed labels, but not what name it used for the repo underneath a target's runfiles root. And the runfiles root is one of the things put on `PYTHONPATH`.

repositories/repositories.bzl
```diff
@@ -79,26 +79,41 @@ def repositories():
...
         # TODO(mattmoor): Is there a clean way to override?
         http_archive(
-            name = "httplib2",
+            name = "com_github_httplib2_httplib2",
             build_file_content = """
 py_library(
    name = "httplib2",
-   srcs = glob(["**/*.py"]),
-   data = ["cacerts.txt"],
+   srcs = select({
```
brandjon: If your issue is related to building this library in the host configuration, then these selects won't do much since they'll always match whichever value you pass to `--host_force_python`. Select reads the version from the configuration, but the whole problem with the host config is that it doesn't allow you to use any version besides whatever is passed into that flag.

repositories/repositories.bzl
drigz: Renaming the repository from "httplib2" to "com_github_httplib2_httplib2" resolves the name conflict, but breaks because containerregistry expects the name "httplib2". Would it make it clearer if I move the comment to the definition of @httplib2, and change it to:

```
# Other repositories, including @containerregistry, expect this be called @httplib2. However,
# due to #3998, this causes naming conflicts, so we use the fully qualified name here.
```

repositories/repositories.bzl
drigz: > the whole problem with the host config is that it doesn't allow you to use any version besides whatever is passed into that flag.

Hmm, I'm not sure I understand. I want the build to use the source code that matches how it will be executed, ie with --host_force_python=PY<x>, it should use the python<x> source.

My understanding: currently, genrules execute under PY2 regardless of --host_force_python (at least, --host_force_python=PY3 doesn't seem to break rules_docker). With --incompatible_use_python_toolchains, genrules execute under PY3 by default, which breaks rules_docker unless --host_force_python=PY2 is used. By adding select()s, httplib2 uses the source that matches the value of --host_force_python, so genrules work with either Python version. Does that make sense?

What I'm (even) less sure about is how this select() affects the version of Python used when building in the non-host configuration: would this change make it impossible to have a PY2 py_binary that depends on httplib2 (assuming --host_force_python=PY3)? Is it possible to have a py_library that uses different source depending on whether it is depended upon by a PY2 or PY3 py_binary?

drigz:
The CI failure is due to invocation of `bazel run @io_bazel_rules_docker//tests/docker:push_test` in a test workspace. This fails because with --noincompatible_use_python_toolchains, the python3 source is selected and executed with python2. I presume this is because of https://github.com/bazelbuild/bazel/issues/4815, but I don't know if it can be resolved.

repositories/repositories.bzl
brandjon: I see, your rename works not because of the repo_mapping per se but because the http_archive.name attribute was also changed. Makes sense.

repositories/repositories.bzl
brandjon: Genrule tools (and other things in host configuration) will be configured with the python version set to whatever `--host_force_python` is, which defaults to `PY3` (at least when `--incompatible_py3_is_default` is enabled, which is flipped to true in 0.24). The `python_version` attribute is totally ignored for host-configured things.

However, if Python toolchains are not enabled (you haven't set `--incompatible_use_python_toolchains`), it's possible that these tools are running with a Python 2 interpreter at execution time even though they're configured to use Python 3 due to the host configuration (see bazelbuild/bazel#4815). So the limitation of what interpreter you can request in the host config, is canceled out by a bug that doesn't give you the interpreter you requested. This may be why setting `--host_force_python=PY3` didn't break rules_docker. Enabling toolchains fixes the latter bug and reveals the need to set `--host_force_python` for the build to work.

bazelbuild/bazel#4815 never affected how selects resolve. In the host config, they'll always think the version is whatever `--host_force_python` is, regardless of whether or not you're using toolchains -- and that's correct in the sense that that *is* what the version is in the host config.

Outside of the host config, you can use these selects in `py_library` and elsewhere to vary the behavior depending on whether it's underneath a PY2 or PY3 `py_binary`.

brandjon:
> This fails because with --noincompatible_use_python_toolchains, the python3 source is selected and executed with python2. I presume this is because of bazelbuild/bazel#4815, but I don't know if it can be resolved.

Yes, it's #4815. You can circumvent it with [this workaround](https://github.com/bazelbuild/bazel/issues/4815#issuecomment-460777113), which passes a `py_runtime` that distinguishes between the two versions as `--python_top`.

drigz:
This is still quite confusing to me, but I think I can conclude that select() is not very useful without --incompatible_use_python_toolchains. I will close this PR for now, and maybe some time after 0.27 when people are interested in migrating to PY3, the approach can be resurrected if it's still relevant.

brandjon:
Basically, Bazel has a notion of whether each target is "PY2" or "PY3", and it uses this notion to resolve select()s and to validate srcs_version. When toolchains are enabled, it will also use this notion to choose the right interpreter to use.

I think a best practice would be to do each of the following:

1. use analysis-time version features like select()s / `srcs_version` as normal, even in the host config
2. pass `--host_force_python` to whichever version your build requires
3. either enable toolchains (default in 0.27+) or use the workaround I mentioned to avoid problems due to bazelbuild/bazel#4815.

Granted, this requires passing one or two flags to the build for downstream users, but it gives you consistent/correct version behavior throughout the build, you just can't use two different versions in the host config.

PR #859, opened 2019-05-27
Allow dockerfile build to be cached.

Globegitter:
Some teams prefer to trade build speed over hermeticity, so allow docker caching to be turned on for the build.

### DIFF
--- a/contrib/dockerfile_build.bzl
+++ b/contrib/dockerfile_build.bzl
@@ -59,13 +59,18 @@ def _impl(repository_ctx):
         for pair in repository_ctx.attr.build_args.items():
             build_args.extend(["--build-arg", "%s=%s" % (pair[0], pair[1])])
 
+    cache_args = []
+    if not repository_ctx.attr.cache:
+        cache_args.append("--no-cache")
+
+
     # The docker bulid command needs to run using the supplied Dockerfile
     # because it may refer to relative paths in its ADD, COPY and WORKDIR
     # instructions.
     command = [docker_path, "build"]
     command.extend(build_args)
+    command.extend(cache_args)
     command.extend([
-        "--no-cache",
         "-f",
         dockerfile_path,
         "-t",
@@ -115,6 +120,10 @@ dockerfile_image = repository_rule(
             allow_single_file = True,
             mandatory = True,
             doc = "The label for the Dockerfile to build the image from.",
+        ),
+        "cache": attr.bool(
+            default = False,
+            doc = "Turn on docker caching for the build.",
         ),
     },
     implementation = _impl,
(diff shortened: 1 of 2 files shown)


contrib/dockerfile_build.bzl
```diff
@@ -106,6 +110,10 @@ dockerfile_image = repository_rule(
             doc = "A map of args to pass to the --build-arg option in the " +
                   "`docker build` command.",
         ),
+        "cache": attr.bool(
+            default = False,
+            doc = "Turn on docker caching for the build.",
```
nlopezgi: hmm... Lets add a big warning that this will mess up with Bazel caching mechanisms and in case of failures / issues you might need to do a bazel clean --expunge_async + do a docker cache purge.

testing/examples/WORKSPACE
```diff
@@ -63,6 +63,7 @@ container_pull(
 
 dockerfile_image(
     name = "basic_alpine_dockerfile",
+    cache = True,
```
nlopezgi: This cant really be tested properly without doing an e2e test. 
The test might do the following:
setup  - have a rule w cache = True in this `WORSKPACE` file (a different target, though)
1- Create an image from dockerfile w/o using bazel, check its  sha
2- Create an image using Bazel. Check its sha256 is the same as the one produced in step 1
As a base case when not enabled the test would need to:
setup  - have the exact same  w/o cache = True (reuse this `basic_alpine_dockerfile` target) 
3- build the image using bazel
4- check its sha is different from the one produced in step 1 and 2.

One key question I have is if we have all the outputs in this rule already to expose the sha so you can do the check you need to do.

cjwagner:
Please keep this PR on hold until it is actually mergeable and then unhold with `/hold cancel`.
ref #862 
/hold

smukherj1:
@nlopezgi @cjwagner Tide is unable to merge this because we have the branch protection setting that requires the pull request source branch to be up to date for it to be mergeable. This is a known issue but this pull request is an example where we accidentally got a pull request in a state where this issue happens but we didn't realize it because we got no indication from prow/tide that it is stuck in this loop.

We are working on a pull request automation utility and I think we can add a feature to that utility that looks for pull requests that are approved but have an out of date source branch. If so, it can automatically add "hold" to those pull requests so that tide doesn't get into the infinite loop. Once the pull request is updated by merging the latest changes from master, this utility can come in and remove the "hold" label.

cjwagner:
See number 5 here: https://github.com/kubernetes/test-infra/blob/master/prow/cmd/tide/maintainers.md#best-practices

Repos that are managed by Tide should not require that pull request branches are up to date. Tide provides the exact same testing guarantees in a more intelligent way so that PR authors are not required to update their PR branches when the base branch is updated. This is achieved by tracking the base branch commit SHA that presubmit tests are run against and rerunning tests if the base branch commit SHA changes.

nlopezgi:
Hi @Globegitter, 
do you think you'll get a chance to continue this PR or should we close?

Globegitter:
Let's close this for now, if I get around to this again I will reopen.

PR #867, opened 2019-06-04
fast puller argument parser

xwinxu:


container/go/fast_puller.go
```diff
@@ -0,0 +1,47 @@
+package main
```
smukherj1: Add a top level doc for what this binary does. See [here](https://github.com/bazelbuild/rules_docker/blob/aafaf340e545322918aac22729ff11638b8c2452/container/go/cmd/extract_config/main.go#L1) for an example.

container/go/fast_puller.go
```diff
@@ -0,0 +1,47 @@
...
+
+import (
+	"flag"
+	"log"
+)
+
+var (
+	name = flag.String("name",
```
smukherj1: Get rid of line wrapping here and for every other variable below. The convention in Go is to let lines run as long as necessary and let your editor soft wrap lines for you. See [here](https://github.com/bazelbuild/rules_docker/blob/aafaf340e545322918aac22729ff11638b8c2452/container/go/cmd/extract_config/main.go#L13)

container/go/fast_puller.go
```diff
@@ -0,0 +1,47 @@
...
+
+	cache = flag.String("cache",
+		"",
+		"Image's files cache directory.",
+	)
+)
+
+func main() {
+	flag.Parse()
+	log.Println("Running the Image Config & Manifest Extractor.")
```
smukherj1: Well not really :P
I assume you want to say `Running the Image Puller.` here

container/go/fast_puller.go
```diff
@@ -0,0 +1,47 @@
...
+	log.Println("Command line arguments:")
+	log.Printf("-imageTar: %q", *name)
```
smukherj1: Change -imageTar to -name

container/go/fast_puller.go
```diff
@@ -0,0 +1,47 @@
...
+	log.Printf("-outConfig: %q", *directory)
```
smukherj1: -directory. Ditto for every other argument below

container/go/fast_puller.go
```diff
@@ -0,0 +1,47 @@
...
+	log.Printf("-outManifest: %q", *configFlag)
+	log.Printf("-outManifest: %q", *cache)
+
+	// mandatory arguments
```
smukherj1: This comment is not strictly necessary as it's quite evident from the log.Fatalln calls below that the arguments being checked below are mandatory.

Also, comments in Go should begin with capital letters and end with a period or the appropriate punctuation. However, for this specific case, I suggest removing the comment.

container/go/cmd/puller/BUILD
```diff
@@ -0,0 +1,14 @@
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
```
xingao267: license header in all files.

xingao267: You can just copy this one https://github.com/bazelbuild/rules_docker/blob/master/container/build_tar.py

container/go/fast_puller.go
xingao267: I guess you probably want to delete this file as container/go/cmd/puller/puller.go is the new one?

container/go/cmd/puller/BUILD
xingao267: Also add license header in the BUILD file as well this line licenses(["notice"])  # Apache 2.0

similar to https://github.com/bazelbuild/rules_docker/blob/master/BUILD

container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
...
+	"os"
+)
+
+var (
+	imgName         = flag.String("name", "", "The name location including repo and digest/tag of the docker image to pull and save. Supports fully-qualified tag or digest references.")
+	directory       = flag.String("directory", "", "Where to save the images files.")
+	clientConfigDir = flag.String("client-config-dir", "", "Specifies where the custom docker config.json is located. Overiddes the value from DOCKER_CONFIG.")
+	cachePath       = flag.String("cache", "", "Image's files cache directory.")
+	arch            = flag.String("architecture", "", "Image platform's CPU architecture.")
+	os1             = flag.String("os", "linux", "The image's OS, if referring to a multi-platform manifest list. Default linux.")
```
xingao267: os?

xingao267: why set the default here?

container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
...
+	osVers          = flag.String("os-version", "", "The image's os version to pull if referring to a multi-platform manifest list.")
```
xingao267: I would put the whole name here osVersion, osFeature

container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
...
+	osFeat          = flag.String("os-features", "", "The image's os features when pulling a multi-platform manifest list.")
+	variant         = flag.String("variant", "", "The desired CPU variant when image refers to a multi-platform manifest list.")
+	features        = flag.String("features", "", "The desired platform features when image refers to a multi-platform manifest list.")
```
xingao267: Be consistent with the docs. You have `when image refers to`, `when pulling ...` and `if referring to ...`

container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
...
+import (
+	"flag"
+	"log"
+	"os"
+)
+
+var (
+	imgName         = flag.String("name", "", "The name location including repo and digest/tag of the docker image to pull and save. Supports fully-qualified tag or digest references.")
+	directory       = flag.String("directory", "", "Where to save the images files.")
+	clientConfigDir = flag.String("client-config-dir", "", "Specifies where the custom docker config.json is located. Overiddes the value from DOCKER_CONFIG.")
```
xingao267: "The path to the directory where the client configuration files are located. Overrides ..."

container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
...
+	cachePath       = flag.String("cache", "", "Image's files cache directory.")
+	arch            = flag.String("architecture", "", "Image platform's CPU architecture.")
```
xingao267: You can refer to https://github.com/google/containerregistry/blob/master/tools/platform_args_.py for the platform docs. And be consistent with one stype for all your flags.

container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
...
+		log.Fatalln("Required option -name was not specified.")
+	}
+	if *directory == "" {
+		log.Fatalln("Required option -directory was not specified.")
+	}
+
+	// If the user provided a client config directory, instruct the keychain resolver
+	// to use it to look for the docker client config
+	if *clientConfigDir != "" {
+		os.Setenv("DOCKER_CONFIG", *clientConfigDir)
```
xingao267: We might not need this. As we can set the env var on the flag at the starkark rule level, and this binary doesn't even need to take the clientConfigDir arg. But we can keep this here for now.

container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
...
+//////////////////////////////////////////////////////////////////////
+// This binary pulls images from a Docker Registry.
+// Unlike regular docker pull, the format this package uses is proprietary.
+
+package main
+
+import (
+	"flag"
+	"log"
+	ospkg "os"
```
smukherj1: I would not rename the standard package as it hinders readability. I would rename the variable that captures the "os" flag something else like "osVar". imo, readers of a language get quite used to the standard library and their API and having one renamed hinders readability more than a command line variable renamed to avoid a name conflict.

However, this is just my opinion and you may choose to disagree :)

smukherj1: Just spoke with @xingao267. Ignore this comment

container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
...
+
+var (
+	imgName         = flag.String("name", "", "The name location including repo and digest/tag of the docker image to pull and save. Supports fully-qualified tag or digest references.")
+	directory       = flag.String("directory", "", "Where to save the images files.")
+	clientConfigDir = flag.String("client-config-dir", "", "The path to the directory where the client configuration files are located. Overiddes the value from DOCKER_CONFIG.")
+	cachePath       = flag.String("cache", "", "Image's files cache directory.")
+	arch            = flag.String("architecture", "", "Image platform's CPU architecture.")
+	os              = flag.String("os", "", "Image's operating system, if referring to a multi-platform manifest list. Default linux.")
+	osVersion       = flag.String("os-version", "", "Image's operating system version, if referring to a multi-platform manifest list.")
+	osFeature       = flag.String("os-features", "", "Image's operating system features, if referring to a multi-platform manifest list.")
```
xingao267: osFeatures

PR #868, opened 2019-06-05
Remove symlinks from lang image runfiles

scele:
Previously, we would place runfiles in an image-agnostic path, such as:
  ```
  /app/pypi__six_1_11_0/six.py
  ```
and then (in a different layer), symlink it to a suitable path under runfiles:
  ```
  /app/testdata/py_image_complex.binary.runfiles/pypi__six_1_11_0/six.py -> /app/pypi__six_1_11_0/six.py
  ```
Change the image structure such that the runfiles tree is constructed in an
image-agnostic path at `/app.runfiles`, which makes it unnecessary to create
symlinks, while still keeping the file paths image-agnostic to support sharing
layers between images.  In the new layout, the runfiles are placed at:
  ```
  /app.runfiles/pypi__six_1_11_0/six.py
  ```
Runfiles are discovered (unless `RUNFILES_DIR` is already set) by looking at
`$0.runfiles`, so we also create a symlink from /app to the py_binary launcher
script, and set image `ENTRYPOINT` to `["python", "/app"]`.

### DIFF
--- a/cc/image.bzl
+++ b/cc/image.bzl
@@ -87,7 +87,6 @@ def cc_image(name, base = None, deps = [], layers = [], binary = None, **kwargs)
     base = base or DEFAULT_BASE
     for index, dep in enumerate(layers):
         base = app_layer(name = "%s.%d" % (name, index), base = base, dep = dep)
-        base = app_layer(name = "%s.%d-symlinks" % (name, index), base = base, dep = dep, binary = binary)
 
     visibility = kwargs.get("visibility", None)
     tags = kwargs.get("tags", None)
--- a/container/image_test.py
+++ b/container/image_test.py
@@ -464,44 +464,28 @@ class ImageTest(unittest.TestCase):
       # Check the application layer, which is on top.
       self.assertTopLayerContains(img, [
         '.',
-        './app',
-        './app/testdata',
-        './app/testdata/py_image.binary.runfiles',
-        './app/testdata/py_image.binary.runfiles/io_bazel_rules_docker',
-        './app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/testdata',
-        './app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/testdata/py_image.py',
-        './app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/testdata/py_image.binary',
-        './app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/testdata/BUILD',
-        './app/testdata/py_image.binary.runfiles/io_bazel_rules_docker/testdata/__init__.py',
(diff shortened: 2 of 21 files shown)


nlopezgi:
Hi @scele 
Thanks for sending this change.
Could you please add some motivation as to why this change is desirable/necessary and what benefits we expect to get from it?
Also, could you comment what is the expected user visible impact of this change, if any?

Upfront warning about this very large change: Getting this change in will be very challenging and time consuming for us, as we will need to be careful that this does not affect/break internal users of rules_docker. If the motivation behind it is not compelling enough we will most likely push back. Even in the case this is compelling enough we will realistically not have the time to perform the necessary careful revision this Q. So please bear with us, as we do highly appreciate your contributions and will do our best to work with you in this PR.

scele:
Hi @nlopezgi, thanks for the heads up. My motivation was twofold:

- Reduce the number of layers in the generated `py_image`s. Currently each item in the `layers = [...]` attribute generates two layers into the image. There is a hard limit on the number of docker image layers, and the incremental loader script becomes slower as the number of layers grows. I was working on breaking our images into finer layers, and realized that there probably isn't any reason for not putting all files to target-agnostic paths, and removing the symlink layers.
- Reduce the complexity of the rule. I was also looking into rewriting the lang image rules on top of `container_layer`, and wanted to start by incrementally reducing the complexity.

However, I now ended up writing a customized `py_image` macro in our downstream code, that drops the symlink layers. So if you think removing the symlinks is too disruptive (it can be indeed, since the root of the runfiles dir within the image would change), then that's fair. **Feel free to close this PR.** We can always revisit this later if needed.

---

Btw, you might be interested in the specific problem I was working on and the solution that I came up with. The problem was that I had 10 `py_image` targets, each depending on large `py_library` targets, say `requirement("tensorflow")`. Even if I use `py_image(..., layers = [requirement("tensorflow")])`, it still ends up generating 10 different layer tarballs that contain tensorflow files, because each invocation of `py_image` macro instantiates separate targets for the layers. But the 10 tarballs are identical, such a waste of disk space and build time.

So, I'm using aspects to generate `container_layer` targets for any `py_library` in the transitive deps of the `py_image` that has `tags = ["container-layer"]` attribute. Using aspects means that these `container_layer` rules are instantiated at the same path as the tagged `py_library`, something like `@pip_deps__tensorflow//:layer.tar`. This way the same tarball can be shared between all 10 `py_image` rules, and layering can be controlled with tags in addition to `layers=[]` attribute.

smukherj1:
beep boop I'm a bot

nlopezgi:
Hey @scele, is this PR still something you will be working on or can we close?

scele:
I think I'll close this. I can reopen if I have a good case for this.

PR #870, opened 2019-06-07
Add puller function in go

xiaohegong:
Added function `pull` that pulls an image in go.

The function reuses the implementation of `crane pull` [here](https://github.com/google/go-containerregistry/blob/15879f261e2082fe64b9f92d6738f4d959615dfc/pkg/crane/pull.go#L51) with additional platform specification feature.

### DIFF
--- a/container/go/cmd/puller/puller.go
+++ b/container/go/cmd/puller/puller.go
@@ -1,0 +1,57 @@
+package puller
+
+import (
+	"fmt"
+	"log"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/cache"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+)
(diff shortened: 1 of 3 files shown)


container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
...
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+
+	"github.com/google/go-containerregistry/pkg/authn"
+	"github.com/google/go-containerregistry/pkg/name"
+	"github.com/google/go-containerregistry/pkg/v1/cache"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+)
+
+const iWasADigestTag = "i-was-a-digest"
```
xingao267: Also copy the comment above it.

container/go/cmd/puller/puller.go
```diff
@@ -0,0 +1,57 @@
+package puller
```
xingao267: Add license header.

container/go/cmd/puller/puller.go
xingao267: Also make a note that this file is mostly copied from https://github.com/google/go-containerregistry/blob/master/pkg/crane/pull.go

PR #875, opened 2019-06-11
Fix incremental loader cleanup when no run statements are defined

clintharrison:
This fixes the bug mentioned in #793. When we don't generate any run statements (that is, `run` in `incremental_load()` is false), we don't `exec docker`, and the cleanup runs twice.

### DIFF
--- a/container/incremental_load.sh.tpl
+++ b/container/incremental_load.sh.tpl
@@ -239,8 +239,9 @@ function read_variables() {
 %{tag_statements}
 
 # An optional "docker run" statement for invoking a loaded container.
-# This is not executed if the single argument --norun is passed.
-if [ "a$*" != "a--norun" ]; then
+# This is not executed if the single argument --norun is passed or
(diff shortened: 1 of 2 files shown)


container/incremental_load.sh.tpl
```diff
@@ -239,8 +239,9 @@ function read_variables() {
 %{tag_statements}
 
 # An optional "docker run" statement for invoking a loaded container.
-# This is not executed if the single argument --norun is passed.
-if [ "a$*" != "a--norun" ]; then
+# This is not executed if the single argument --norun is passed or
+# no run_statements are generated (in which case, 'run' is 'False').
+if [[ "a$*" != "a--norun" && "%{run}" == "True" ]]; then
```
xingao267: You can probably reuse run_statements and check if that's not an empty string here.

clintharrison: I was hesitant to do that -- `run_statements` can be modified by the consumer's `docker_run_flags` attribute, and this might contain double quotes.

PR #876, opened 2019-06-12
Expose attributes from lang/image.bzl as struct

scele:
Besides exposing the rule itself, it is a common pattern to expose
rule attributes, outputs, toolchains and the implementation function
as a struct.  This makes it easier to wrap the rule in another rule,
without duplicating the attribute definitions.  Apply this pattern
to lang/image.bzl, and remove duplicated definitions from the wrapping
rules in java/image.bzl and nodejs/image.bzl.

### DIFF
--- a/java/image.bzl
+++ b/java/image.bzl
@@ -27,8 +27,8 @@ load(
 )
 load(
     "//lang:image.bzl",
-    "app_layer_impl",
     "layer_file_path",
+    lang_image = "image",
 )
 load(
     "//repositories:repositories.bzl",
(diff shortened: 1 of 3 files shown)


java/image.bzl
```diff
@@ -155,30 +155,13 @@ def _jar_dep_layer_impl(ctx):
...
-    attrs = dicts.add(_container.image.attrs, {
-        # The base image on which to overlay the dependency layers.
-        "base": attr.label(mandatory = True),
-
-        # The binary target for which we are synthesizing an image.
-        "binary": attr.label(mandatory = False),
-        # Set this to true to create an empty workspace directory under the
-        # app directory specified as the 'directory' attribute.
-        "create_empty_workspace_dir": attr.bool(default = False),
-        # https://github.com/bazelbuild/bazel/issues/2176
```
xingao267: This comment and link to the issue is still valid. Let's move it to the comment in the lang/image.bzl.

scele: Done!

PR #880, opened 2019-06-13
Remove reference to cache which doesn't exist in internal codebase

smukherj1:


Globegitter:
@smukherj1 Does that mean caching will not be supported once things are migrated to go?

xingao267:
@Globegitter this change is temporary. We've fixed the internal codebase and will add this back soon.

PR #882, opened 2019-06-13
Read image from OCI image layout 

xiaohegong:
- Added a `Read` function that parses an [OCI image index](https://github.com/opencontainers/image-spec/blob/master/image-layout.md) and returns the image inside of the given path. Image returned by `Read` will be used by `pusher` later.  `Read` only works with single image manifests for now. 

- Added a testing OCI image index `/test_index` that works with this function.


container/go/pkg/oci/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/layout"
+)
+
+// Read gets and returns the image in the given path <src> that follows the OCI Image Layout.
+// (https://github.com/opencontainers/image-spec/blob/master/image-layout.md#oci-image-layout-specification)
+// Specifically, <src> must contains "index.json" that servers as an entrypoint for the contained image.
+// The image content and configs must be non-empty and stored in <src>/blobs/<SHAxxx>/.
+// NOTE: this only reads index with a single image.
+func Read(src string) v1.Image {
+	// Open the layout at /src as an Image Index.
```
smukherj1: This comment almost literally translates the next line of code. Suggest removing it or replacing it with something like `Reading the docker image from the given directory as an OCI image`

container/go/pkg/oci/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	idx, err := layout.ImageIndexFromPath(src)
+	if err != nil {
+		log.Fatal(err)
```
smukherj1: Add more context to error. Also avoid fatal errors inside functions. Convention in Go is to propagate the error upwards to the main function where a call to `log.Fatal` can be made.

Suggest changing the return value of `func Read` to be `(v1.Image, error)` and returning:
`return nil errors.Wrapf(err, "unable to read image from %s", src)`

container/go/pkg/oci/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	// Read the contents of the layout -- we expect to find a single image.
+	// TODO (xiaohegong): handle case with multiple manifests.
+	manifest, err := idx.IndexManifest()
+	if err != nil {
+		log.Fatal(err)
```
smukherj1: return error with context:

`return nil, errors.Wrapf(err, "unable to get the manifest for Image loaded from %s", src)`

container/go/pkg/oci/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	if len(manifest.Manifests) > 1 {
+		log.Fatalf("found %d manifests, expected 1", len(manifest.Manifests))
```
smukherj1: Return error instead. Convention in Go for value mismatch is got/want. So something like:
`return nil, fmt.Errorf("got %d manifests, want 1", len(manifest.Manifests)))`

container/go/pkg/oci/reader.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+
+	// Read that single image as a v1.Image and return it.
+	digest := manifest.Manifests[0].Digest
+
+	img, err := idx.Image(digest)
+	if err != nil {
+		log.Fatal(err)
```
smukherj1: return error with context

container/go/pkg/oci/test_index/blobs/sha256/a0cfcd4cc98a67def7ce9a0c7644d1c415d56d6d44c4a079a447f7eafb253048
```diff
@@ -0,0 +1 @@
+{"architecture": "amd64", "author": "Bazel", "config": {"Env": ["PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin", "SSL_CERT_FILE=/etc/ssl/certs/ca-certificates.crt"]}, "created": "1970-01-01T00:00:00Z", "history": [{"author": "Bazel", "created": "1970-01-01T00:00:00Z", "created_by": "bazel build ..."}, {"author": "Bazel", "created": "1970-01-01T00:00:00Z", "created_by": "bazel build ..."}], "os": "linux", "rootfs": {"diff_ids": ["sha256:01092e5921c5543a918d54d9df752ee09a84c912a1d914b7eb37e7152f20b951", "sha256:87c747af6dc3478b57493f1f3a2e0821696f8d56b86b1fef1128d1d74881cf7c"], "type": "layers"}}
```
smukherj1: Are these image related files necessary? I don't see them used anywhere

xiaohegong: I added `test_index` as an example OCI index that `Read` can take in for testing purpose.

container/go/pkg/oci/reader.go
xiaohegong: I changed it to return the err message from `err`, since there can be various context for error from this function.

container/go/pkg/oci/reader.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+// Read returns a docker image from the given path. The docker image should have been written by "Write".
+// The image in the given path <src> follows the OCI Image Layout.
+// (https://github.com/opencontainers/image-spec/blob/master/image-layout.md#oci-image-layout-specification)
+// Specifically, <src> must contains "index.json" that servers as an entrypoint for the contained image.
+// The image content and configs must be non-empty and stored in <src>/blobs/<SHAxxx>/.
+// NOTE: this only reads index with a single image.
+func Read(src string) (v1.Image, error) {
+	// Reading the image layout.
```
smukherj1: Actually I would just remove this comment because I don't see it adding much value. The error handling right after gives me info to understand what's going on here.

container/go/pkg/oci/reader.go
```diff
@@ -0,0 +1,60 @@
...
+	if len(manifest.Manifests) > 1 {
+		return nil, fmt.Errorf("got %d manifests, want 1", len(manifest.Manifests))
+	}
+
+	// Read that single image as a v1.Image and return it.
+	digest := manifest.Manifests[0].Digest
+
+	img, err := idx.Image(digest)
+	if err != nil {
+		return nil, fmt.Errorf(err.Error())
```
smukherj1: You can just do `return nil, err` here.

However, instead I would suggest:
`return nil, errors.Wrapf(err, "unable to load image with digest %s obtained from the manifest", digest)`

container/go/pkg/oci/test_index/blobs/sha256/a0cfcd4cc98a67def7ce9a0c7644d1c415d56d6d44c4a079a447f7eafb253048
smukherj1: I would add these files when you actually add the test instead of now.

xiaohegong: Ok

PR #886, opened 2019-06-13
Implement OCI Image format puller

xwinxu:
Extended rules_docker API to now contain the `Write` function which outputs the images from the puller in OCI standard format. BUILD files updated in the process.

### DIFF
--- a/container/go/cmd/puller/puller.go
+++ b/container/go/cmd/puller/puller.go
@@ -21,6 +21,7 @@ import (
 	"flag"
 	"log"
 	ospkg "os"
+	"strings"
 
 	"github.com/bazelbuild/rules_docker/container/go/pkg/oci"
 	v1 "github.com/google/go-containerregistry/pkg/v1"
@@ -53,7 +54,7 @@ const iWasADigestTag = "i-was-a-digest"
 // with slight modification to take in a platform argument.
 // Pull the image with given <imgName> to destination <dstPath> with optional
 // cache files and required platform specifications.
-func pull(imgName, dstPath, cachePath string, platform *v1.Platform) {
+func pull(imgName, dstPath, cachePath string, platform v1.Platform) {
 	// Get a digest/tag based on the name.
 	ref, err := name.ParseReference(imgName)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


container/new_pull.bzl
```diff
@@ -0,0 +1,209 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
```
xingao267: I think you probably don't want the starlark part in this PR?

container/go/cmd/puller/puller.go
```diff
@@ -12,27 +12,30 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/cache"
 	"github.com/google/go-containerregistry/pkg/v1/remote"
-	"github.com/google/go-containerregistry/pkg/v1/tarball"
 )
 
 var (
 	imgName         = flag.String("name", "", "The name location including repo and digest/tag of the docker image to pull and save. Supports fully-qualified tag or digest references.")
 	directory       = flag.String("directory", "", "Where to save the images files.")
 	clientConfigDir = flag.String("client-config-dir", "", "The path to the directory where the client configuration files are located. Overiddes the value from DOCKER_CONFIG.")
+	cachePath       = flag.String("cache", "", "Image's files cache directory.")
```
xingao267: Suvanjan removed this due to importing issue, we probably shouldn't add this back now.

container/new_pull.bzl
xwinxu: git problem again. I switched branches but it somehow didn't push properly  :(

xwinxu: I'll delete the file and try again.

container/go/cmd/puller/puller.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 The Bazel Authors. All rights reserved.
+/// Copyright 2015 The Bazel Authors. All rights reserved.
```
xingao267: Try to avoid this kind of unnecessary change which is not really related to the purpose of this PR.

container/go/pkg/oci/BUILD
```diff
@@ -1,12 +1,31 @@
...
+# limitations under the License.
+# ###
+# Build for the new writer function to write OCI Format Images to disk.
 load("@io_bazel_rules_go//go:def.bzl", "go_library")
 
 go_library(
     name = "go_default_library",
-    srcs = ["reader.go"],
+    srcs = [
+        "reader.go",
```
xingao267: Are you sure it builds? reader.go seems do not exist.

xingao267: Sorry...my bad. I missed Heddy's PR. Ignore this comment.

PR #894, opened 2019-06-17
Implement OCI image puller binary and writer utility

xwinxu:
Extended rules_docker API to now contain the `Write` function which outputs the images from the puller in OCI standard format. BUILD files updated in the process.

### DIFF
--- a/container/go/cmd/puller/BUILD
+++ b/container/go/cmd/puller/BUILD
@@ -11,25 +11,31 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+# ###
+# Build file for new puller binary based on go-containerregistry backend.
 
+load("@bazel_gazelle//:def.bzl", "gazelle")
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
 
-go_library(
-    name = "go_default_library",
-    srcs = ["puller.go"],
-    importpath = "github.com/bazelbuild/rules_docker/container/go/cmd/puller",
-    visibility = ["//visibility:private"],
-    deps = [
-        "@com_github_google_go_containerregistry//pkg/authn:go_default_library",
-        "@com_github_google_go_containerregistry//pkg/name:go_default_library",
-        "@com_github_google_go_containerregistry//pkg/v1:go_default_library",
-        "@com_github_google_go_containerregistry//pkg/v1/remote:go_default_library",
-        "@com_github_google_go_containerregistry//pkg/v1/tarball:go_default_library",
-    ],
-)
+# gazelle:prefix github.com/bazelbuild/rules_docker
+gazelle(name = "gazelle")
 
 go_binary(
     name = "puller",
     embed = [":go_default_library"],
     visibility = ["//visibility:public"],
+)
+
+go_library(
+    name = "go_default_library",
+    srcs = ["puller.go"],
+    importpath = "github.com/bazelbuild/rules_docker",
+    visibility = ["//visibility:private"],
+    deps = [
+        "//container/go/pkg/oci:go_default_library",
+        "@com_github_google_go_containerregistry//pkg/authn:go_default_library",
+        "@com_github_google_go_containerregistry//pkg/name:go_default_library",
+        "@com_github_google_go_containerregistry//pkg/v1:go_default_library",
+        "@com_github_google_go_containerregistry//pkg/v1/remote:go_default_library",
+    ],
 )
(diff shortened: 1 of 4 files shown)


container/go/cmd/puller/puller.go
```diff
@@ -1,4 +1,4 @@
-// Copyright 2015 The Bazel Authors. All rights reserved.
+/// Copyright 2015 The Bazel Authors. All rights reserved.
```
xingao267: Double check for these kind of unnecessary changes.

container/go/cmd/puller/puller.go
```diff
@@ -41,67 +43,64 @@ var (
...
-			log.Fatal("ref wasn't a tag or digest")
-		}
-		s := fmt.Sprintf("%s:%s", d.Repository.Name(), iWasADigestTag)
-		tag, err = name.NewTag(s)
-		if err != nil {
-			log.Fatalf("parsing digest as tag (%s): %v", s, err)
-		}
+	// // Image file to write to disk.
+	if err := oci.Write(img, dstPath); err != nil {
+		// if err := oci.Write(img, path); err != nil {
```
xingao267: remove

container/go/cmd/puller/puller.go
```diff
@@ -41,67 +43,64 @@ var (
...
+	if *clientConfigDir != "" {
+		ospkg.Setenv("DOCKER_CONFIG", *clientConfigDir)
+	}
+
+	// Create a Platform struct with arguments
+	platform := v1.Platform{
+		Architecture: *arch,
+		OS:           *os,
+		OSVersion:    *osVersion,
+		OSFeatures:   strings.Fields(*osFeatures),
```
xingao267: strings.Fields assumes the input strings are separated by space. Specify this explicitly in the flag documentation for osFeatures and features.

container/go/pkg/oci/write.go
```diff
@@ -0,0 +1,49 @@
...
+func Write(img v1.Image, dstPath string) error {
+	// Path represents an OCI image layout ooted in a file system path
+	var path layout.Path
+	var err error
+
+	// Open the layout, if it already exists.
+	if path, err = layout.FromPath(dstPath); err != nil {
+		// Does not already exist, so initialize it with an empty index.
+		if path, err = layout.Write(dstPath, empty.Index); err != nil {
+			log.Fatalf("cannot initialize layout: %v", err)
```
xingao267: Return the error 

return fmt.Errorf(...)

container/go/pkg/oci/write.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/empty"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
+)
+
+// Write writes a v1.Image to the Path and updates the index.json to reference it.
+// This is just syntactic sugar wrapping Path.AppendImage from the go-container registry.
+func Write(img v1.Image, dstPath string) error {
+	// Path represents an OCI image layout ooted in a file system path
+	var path layout.Path
+	var err error
```
xingao267: You don't need to do this.

You can just initialize err inline in your first call by changing = to :=

container/go/pkg/oci/write.go
```diff
@@ -0,0 +1,49 @@
...
+	if path, err = layout.FromPath(dstPath); err != nil {
+		// Does not already exist, so initialize it with an empty index.
+		if path, err = layout.Write(dstPath, empty.Index); err != nil {
+			log.Fatalf("cannot initialize layout: %v", err)
+		}
+
+	}
+
+	if err := path.AppendImage(img); err != nil {
+		return fmt.Errorf("unable to write image to path")
```
xingao267: Include the original err

return fmt.Errorf("unable to write image to path: %v", err)

xingao267: Include the original err

return fmt.Errorf("unable to write image to path: %v", err)

Also use errors.Wrapf everywhere. I think that's more recommended.

container/go/pkg/oci/write.go
xwinxu: I tried that before. But it gives me a path undefined error because I reuse path again in line 43.

container/go/cmd/puller/BUILD
```diff
@@ -11,25 +11,31 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+# ###
+# Build file for new puller binary based on go-containerregistry backend.
```
xingao267: Usually we don't write this kind of comment. But it doesn't hurt to have it. No need to change.

PR #896, opened 2019-06-17
Add TestRead and simple testing index

xiaohegong:
- Added `TestRead` that checks the v1.Image outputted by `Read`
  - The checks are mostly covered by `validate.Image(img)`
- Added small test case index from [here](https://pantheon.corp.google.com/gcr/images/distroless/GLOBAL/base@sha256:edc3643ddf96d75032a55e240900b68b335186f1e5fea0a95af3b4cc96020b77/details?tab=info)


container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,98 @@
...
+import (
+	"testing"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/google/go-containerregistry/pkg/v1/validate"
+)
+
+// This test index is pulled by puller.go from gcr.io/distroless/base@sha256:edc3643ddf96d75032a55e240900b68b335186f1e5fea0a95af3b4cc96020b77
+var (
```
smukherj1: Suggest using table driven tests here. See https://github.com/golang/go/wiki/TableDrivenTests#example-of-a-table-driven-test. Create an anonymous struct that stores the variables you are storing here. This way we can add more test cases easily in the future that test more types of images.

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,98 @@
...
+	if got, want := cfgDigest, cfgName; got != want {
+		t.Errorf("ConfigName(); want: %v got: %v", want, got)
+	}
+
+	layers, err := img.Layers()
+	if err != nil {
+		t.Fatalf("img.Layers(): %v", err)
+	}
+
+	layer1Digest, err := layers[0].Digest()
```
smukherj1: I would suggest looping over each layer here instead of only checking the first layer. In the anonymous struct in your table driven test, add a new field that indicates the number of layers to expect in the image and add a code to verify gotLayers == wantLayers. Refactor the rest of this function in a separate "validateLayer" that validates a given layer.

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"testing"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/google/go-containerregistry/pkg/v1/validate"
+)
+
+var readertests = []struct {
+	manifestDigest v1.Hash
```
smukherj1: Add comments for each field.

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,108 @@
...
+	configDigest   v1.Hash
+	layerHashes    []v1.Hash
+	mediaType      types.MediaType
+	testPath       string
```
smukherj1: Add a `name` field to give each test case a short descriptive name & print that when you call t.Run

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,108 @@
...
+		},
+		types.DockerManifestSchema2,
+		"testdata/test_index1",
+	},
+}
+
+// TestRead checks the v1.Image outputted by <Read> by validating its manifest, layers and configs.
+func TestRead(t *testing.T) {
+	for _, rt := range readertests {
+		t.Run("testsrc:"+rt.testPath, func(t *testing.T) {
```
smukherj1: `t.Run(rt.name, ...)` here

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,108 @@
...
+			img, err := Read(rt.testPath)
+			if err != nil {
+				t.Fatalf("Read(%s): %v", rt.testPath, err)
+			}
+
+			if err := validate.Image(img); err != nil {
```
smukherj1: Short comment explaining what validate.Image does.

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,108 @@
...
+				t.Errorf("validate.Image(): %v", err)
+			}
+
+			mt, err := img.MediaType()
+			if err != nil {
+				t.Errorf("img.MediaType(): %v", err)
+			} else if got, want := mt, rt.mediaType; got != want {
+				t.Errorf("img.MediaType(); want: %v got: %v", want, got)
```
smukherj1: got before want in error messages.

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,108 @@
...
+			if err != nil {
+				t.Fatalf("Read(%s): %v", rt.testPath, err)
+			}
+
+			if err := validate.Image(img); err != nil {
+				t.Errorf("validate.Image(): %v", err)
+			}
+
+			mt, err := img.MediaType()
+			if err != nil {
```
smukherj1: `mt, err := img.MediaType(); err != nil {`

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,108 @@
...
+			if got, want := cfgDigest, cfgName; got != want {
+				t.Errorf("ConfigName(); want: %v got: %v", want, got)
+			}
+
+			layers, err := img.Layers()
+			if err != nil {
+				t.Fatalf("img.Layers(): %v", err)
+			}
+
+			// Validate the digests and media type for each layer
```
smukherj1: Refactor this bit into another function to validate a given layer in a given image.

smukherj1: Period at the end of comments.

container/go/pkg/oci/reader_test.go
xiaohegong: I just realized `mt` is used two lines later, same for later cases.

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"testing"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/google/go-containerregistry/pkg/v1/validate"
+)
+
+var readertests = []struct {
+	// A descriptive name for this test case.
```
smukherj1: The convention is to comment fields starting with the name of the field. So something like
`// name is the name of this test case.`

and the same for every other field below.

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,123 @@
...
+	// The hash code of the config metadata file.
+	configDigest v1.Hash
+	// An array of hash codes for each layer in this image.
+	layerHashes []v1.Hash
+	// The media type of this image.
+	mediaType types.MediaType
+	// The relative path of this test case.
+	testPath string
+}{
+	// #1 This test index is the output of puller.go
```
smukherj1: As a reader of a test case, I'm also interested in a tl; dr of what exactly this test case is trying to do. So comment something like:
`// Ensure the image downloaded by puller.go from <image path> located in testdata/test_index1 can be loaded.`

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+import (
+	"testing"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/google/go-containerregistry/pkg/v1/validate"
+)
+
+var readertests = []struct {
```
smukherj1: Comment here explaining what each test case in readerTest does.

`// readerTests has test cases for the Read function that loads a docker image stored on disk serialized in OCI format by "Write"`

container/go/pkg/oci/reader_test.go
smukherj1: sgtm

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,123 @@
...
+			for i, layer := range layers {
+				validateLayer(layer, rt.layerHashes[i], i, t)
+			}
+
+		})
+	}
+}
+
+// validateLayer checks if the digests and media type matches for the given layer.
+func validateLayer(layer v1.Layer, layerHash v1.Hash, i int, t *testing.T) {
```
smukherj1: Convention in Go unit tests is to not pass in the `t *testing.T` argument to helper functions. Instead, make this function return an error and deal with the error in the caller.

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,123 @@
...
+			}
+
+			layers, err := img.Layers()
+			if err != nil {
+				t.Fatalf("img.Layers(): %v", err)
+			}
+
+			// Validate the digests and media type for each layer.
+			for i, layer := range layers {
+				validateLayer(layer, rt.layerHashes[i], i, t)
```
smukherj1: Avoid passing in the index here. validateLayer shouldn't necessarily care if the layer it's being given to validate is part of a list of layers of simple a layer in isolation. If you modify validateLayer to return an error, you can print the index in the message when you deal with the error here.

container/go/pkg/oci/reader_test.go
```diff
@@ -0,0 +1,129 @@
...
+					t.Fatalf("layers[%d] is invalid: %v", i, err)
+				}
+			}
+
+		})
+	}
+}
+
+// validateLayer checks if the digests and media type matches for the given layer.
+func validateLayer(layer v1.Layer, layerHash v1.Hash) (int, error) {
```
smukherj1: This return value is not used anywhere. Please remove it in a follow up PR

PR #899, opened 2019-06-18
run dockerfile examples tests

alex1545:


testing/e2e.sh
```diff
@@ -542,6 +534,36 @@ function test_py_image_deps_as_layers() {
...
+function test_dockerfile_image_apt_pkgs() {
+  cd "${ROOT}/testing/examples/run_instruction_apt_pkgs"
+  clear_docker
+  bazel test --host_force_python=PY2 ...
+}
+
+function test_dockerfile_image_arbitrary() {
+  cd "${ROOT}/testing/examples/run_instruction_arbitrary"
+  clear_docker
+  bazel test --host_force_python=PY2 ...
```
nlopezgi: add .bazelrc files to the projects with this flag

nlopezgi:
I'd rather we setup these new tests as GCB triggers if its not too much additional effort to do so.

testing/e2e.sh
alex1545: Done.
There is only one "internal" project for all examples, so added one such file.

alex1545:
These tests will run as GCB jobs. Removed the tests from the `e2e.sh` script and added `cloudbuild.yaml` files for each of the examples. Single `cloudbuild.yaml` file won't be able to run the examples in parallel because it looks like bazel can be launched at most 4 times at a time.

These tests won't run for now, until this repo is configured to use GCB triggers for its PRs.

contrib/dockerfile_build.bzl
```diff
@@ -89,7 +89,7 @@ exports_files(["{}"])
         if build_result.return_code:
             fail("Image build command failed: {} ({})".format(
                 build_result.stderr,
-                " ".join(command),
+                " ".join([str(e) for e in command]),
```
nlopezgi: why is this needed, shouldn't all elements in command already be strings? If not, its probably better to make them strings as you are putting them in the command instead of converting them here

testing/examples/WORKSPACE
```diff
@@ -16,9 +16,11 @@ workspace(name = "dockerfile_examples")
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive", "http_file")
 
-local_repository(
+http_archive(
```
nlopezgi: Why the change from local_repository? You should leave it as local repo otherwise you would be testing a committed version of rules_docker not the changes you have in the PR

testing/examples/basic/cloudbuild.yaml
```diff
@@ -0,0 +1,29 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Tests that verify the basic example of rules_docker vs Dockerfile.
+
+steps:
+  # Confirm bazel and docker are installed.
+  - name: "gcr.io/asci-toolchain/nosla-ubuntu16_04-bazel-docker-gcloud"
```
nlopezgi: use the bazel container (or do you need gcloud for some reason)?

testing/examples/basic/cloudbuild.yaml
```diff
@@ -0,0 +1,29 @@
...
+    entrypoint: "sh"
+    args: ["-c", "bazel version && docker version"]
+    id: "versions"
+    waitFor: ["-"]
+
+  # Test the basic example.
+  - name: "gcr.io/asci-toolchain/nosla-ubuntu16_04-bazel-docker-gcloud"
+    entrypoint: "sh"
+    args: ["-c", "cd testing/examples && bazel test //basic:all"]
+    waitFor: ["versions"]
```
nlopezgi: use 'dir:' to set the directory to testing/examples (Example: https://github.com/bazelbuild/bazel-toolchains/blob/master/tests/rbe_repo/examples/rbe_custom.yaml#L8)
That way you dont need an entrypoint and the only args you should pass are 
 - test 
 - //basic:all

testing/examples/basic/cloudbuild.yaml
```diff
@@ -0,0 +1,29 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Tests that verify the basic example of rules_docker vs Dockerfile.
+
+steps:
+  # Confirm bazel and docker are installed.
+  - name: "gcr.io/asci-toolchain/nosla-ubuntu16_04-bazel-docker-gcloud"
+    entrypoint: "sh"
+    args: ["-c", "bazel version && docker version"]
```
nlopezgi: No need to have a step just to check the bazel version and docker version

testing/examples/run_instruction_apt_pkgs/cloudbuild.yaml
```diff
@@ -0,0 +1,35 @@
...
+  - name: "gcr.io/asci-toolchain/nosla-ubuntu16_04-bazel-docker-gcloud"
+    entrypoint: "sh"
+    args: ["-c", "bazel version && docker version"]
+    id: "versions"
+    waitFor: ["-"]
+
+  # Test the run_instruction_apt_pkgs example.
+  - name: "gcr.io/asci-toolchain/nosla-ubuntu16_04-bazel-docker-gcloud"
+    entrypoint: "sh"
+    args: ["-c", "cd ../testing/examples && bazel --output_base=/workspace/apt_pkgs test //run_instruction_apt_pkgs:all"]
```
nlopezgi: Why do you need to set the output base?

contrib/dockerfile_build.bzl
alex1545: Done.
Some of the elements were of type `path`.

testing/examples/WORKSPACE
alex1545: Done.
Changed that when had trouble running targets (on GCB) in the "internal" bazel project within the rules_docker project. I figured it out now.

testing/examples/basic/cloudbuild.yaml
alex1545: Done.
No, just need bazel and docker.

testing/examples/basic/cloudbuild.yaml
alex1545: Done.

testing/examples/basic/cloudbuild.yaml
alex1545: Done.

testing/examples/run_instruction_apt_pkgs/cloudbuild.yaml
alex1545: Not a 100% sure, but this makes sure that Bazel generated files can be a accessed by multiple containers (even within the same steps). This is needed here and in the run_instruction_arbitrary example because these examples use the `download_pkgs`, `install_pkgs` and `container_run_and_commit` rules that launch additional containers.

nlopezgi: So do you get an error if you dont set this?
What type of error?
If its because these rules need to launch additional containers, there should be some other way to avoid issues; e.g., by passing `/var/run/docker.sock:/var/run/docker.sock` in `volumes` (https://cloud.google.com/cloud-build/docs/build-config)

alex1545: Ass discussed separately, added a comment why this is needed.

testing/examples/run_instruction_apt_pkgs/cloudbuild.yaml
```diff
@@ -0,0 +1,33 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Tests that verify the example installing apt packages using rules_docker
+# and Dockerfile.
+
+timeout: 21600s
```
nlopezgi: Does this really need all that time to run?

testing/examples/run_instruction_apt_pkgs/cloudbuild.yaml
```diff
@@ -0,0 +1,33 @@
...
+options:
```
nlopezgi: why does this need the highcpu machine?

testing/examples/run_instruction_arbitrary/cloudbuild.yaml
```diff
@@ -0,0 +1,26 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Tests that verify the example manipulating arbitrary commands using
+# rules_docker and Dockerfile.
+
+timeout: 21600s
```
nlopezgi: same q as above

testing/examples/run_instruction_apt_pkgs/cloudbuild.yaml
alex1545: Done.
Reduced timeout to 30 mins.

testing/examples/run_instruction_apt_pkgs/cloudbuild.yaml
alex1545: I thought it was worth it as it bring the time of this test from almost 13 mins to ~5 mins.

testing/examples/run_instruction_arbitrary/cloudbuild.yaml
alex1545: Timeout is reduced.
High CPU here brings the test time from almost 20 mins to just under 10 mins.

testing/examples/run_instruction_apt_pkgs/cloudbuild.yaml
nlopezgi: cool, leave with the highcpu then but remove the timeout then!

alex1545: Done.

PR #901, opened 2019-06-18
Create new starlark rules for new_container_pull

xwinxu:
Modified the starlark (`.bzl`) rules to account for the new puller binary being uploaded to the cloud.

Old puller is left intact, invoke `new_container_pull` to use new puller.

Current speculative container_import written to BUILD file in output directory, see comments for details. No compatibility issues with container_import as of now (??) but potentially will need to refactor puller.go to allow reader to ingest such pulled data.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -29,6 +29,9 @@ platforms:
     # TODO(xingao): Remove after https://github.com/bazelbuild/rules_scala/issues/644 is addressed.
     - "--incompatible_disallow_legacy_javainfo=false"
   ubuntu1804:
+    build_targets:
+    - "@new_alpine_linux_armv6//..."
+    - "@new_distroless_base//..."
     test_targets:
     # We are running the skipped targets remotely only.
     - "--"
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -29,9 +29,6 @@ platforms:
     # TODO(xingao): Remove after https://github.com/bazelbuild/rules_scala/issues/644 is addressed.
     - "--incompatible_disallow_legacy_javainfo=false"
   ubuntu1804:
-    build_targets:
-    - "@new_alpine_linux_armv6//..."
-    - "@new_distroless_base//..."
     test_targets:
     # We are running the skipped targets remotely only.
     - "--"
--- a/container/new_pull.bzl
+++ b/container/new_pull.bzl
@@ -81,17 +81,16 @@ def _impl(repository_ctx):
     # Add an empty top-level BUILD file.
     repository_ctx.file("BUILD", "")
 
-    # TODO(xwinxu): change container_import rule to have an oci attribute to account for new OCI Image Layout
-    # currently exclude:     exports_files(["image.digest", "digest"]) in BUILD
-    repository_ctx.file("image/BUILD", """
-package(default_visibility = ["//visibility:public"])
+    # Currently exports all files pulled by the binary and cannot be depended on by other rules_docker rules.
+    # We will implement a new_container_import rule to comprehend this oci layout.
+    repository_ctx.file("image/BUILD", """package(default_visibility = ["//visibility:public"])
 
 filegroup(
-    name = "image_new",
+    name = "image",
     srcs = glob(["image/**"]),
 )
-exports_files(["index.json"])
-    """)
+
+exports_files(glob(["**"]))""")
 
     args = [
         repository_ctx.path(repository_ctx.attr._puller),
--- a/container/new_pull.bzl
+++ b/container/new_pull.bzl
@@ -1,0 +1,169 @@
+# Copyright 2017 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""An implementation of container_pull based on google/containerregistry using google/go-containerregistry.
+
+This wraps the rulesdocker.go.cmd.puller.puller executable in a
+Bazel rule for downloading base images without a Docker client to
+construct new images.
+"""
+
+_container_pull_attrs = {
+    "architecture": attr.string(
+        default = "amd64",
+        doc = "(optional) Which CPU architecture to pull if this image " +
+              "refers to a multi-platform manifest list, default 'amd64'.",
+    ),
+    "cpu_variant": attr.string(
+        doc = "Which CPU variant to pull if this image refers to a " +
+              "multi-platform manifest list.",
+    ),
+    "digest": attr.string(
+        doc = "(optional) The digest of the image to pull.",
+    ),
+    "docker_client_config": attr.string(
+        doc = "A custom directory for the docker client config.json. " +
+              "If DOCKER_CONFIG is not specified, the value of the " +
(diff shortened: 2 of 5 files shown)


container/new_pull.bzl
```diff
@@ -0,0 +1,208 @@
...
+            ),
+        ]
+
+    kwargs = {}
+    if "PULLER_TIMEOUT" in repository_ctx.os.environ:
+        kwargs["timeout"] = int(repository_ctx.os.environ.get("PULLER_TIMEOUT"))
+
+    result = repository_ctx.execute(args, **kwargs)
+    if result.return_code:
+        fail("Pull command failed: %s (%s)" % (result.stderr, " ".join(args)))
```
xingao267: fail("Pull command failed: %s (%s)" % (result.stderr, " ".join([str(a) for a in args])))

I think Bazel 0.27.0 would complain you are joining path with string.

container/new_pull.bzl
```diff
@@ -0,0 +1,208 @@
...
+
+    updated_attrs = {
+        k: getattr(repository_ctx.attr, k)
+        for k in _container_pull_attrs.keys()
+    }
+    updated_attrs["name"] = repository_ctx.name
+
+    # TODO(xwinxu): if repository_ctx.attr.digest is supported by puller (currently only outputting OCI Layout, cannot parse json,
```
xingao267: I think the purpose of the following piece old in the old pull.bzl is to

1: Return the resolved digest to bazel (https://github.com/bazelbuild/rules_docker/commit/e5785ceaef4eb7e0cc28bdb909fd1b10d5b991c3) so it can be used by bazel workspace resolve file
2. Fail when the pulled image digest is different from the digest user specifies (https://github.com/bazelbuild/rules_docker/pull/734/files).

We should be able to support them as well. Can you check if there is any file in the OCI layout pulled down that contains the digest of the image in the registry, or contains the same value of `image/digest`  that the python puller would pull down?

container/new_pull.bzl
```diff
@@ -0,0 +1,208 @@
...
+        repository_ctx.attr.cpu_variant,
+        "-features",
+        " ".join(repository_ctx.attr.platform_features),
+    ]
+
+    # Use the custom docker client config directory if specified.
+    if repository_ctx.attr.docker_client_config != "":
+        args += ["-client-config-dir", "{}".format(repository_ctx.attr.docker_client_config)]
+
+    # TODO(xwinxu): There is currently no support for caches in rules_docker. Stay tuned...
```
xingao267: I would probably remove this part first. We still need to figure out the cache story for go-containerregistry.

container/new_pull.bzl
```diff
@@ -0,0 +1,208 @@
...
+)
+
+# Pulls a container image.
+
+# This rule pulls a container image into our intermediate format (OCI Image Layout).
+new_container_pull = repository_rule(
+    attrs = _container_pull_attrs,
+    implementation = _impl,
+    environ = [
+        "DOCKER_REPO_CACHE",
```
xingao267: Remove "DOCKER_REPO_CACHE"

container/new_pull.bzl
```diff
@@ -0,0 +1,208 @@
...
+    repository_ctx.file("BUILD", "")
+
+    # TODO(xwinxu): change container_import rule to have an oci attribute to account for new OCI Image Layout
+    # currently exclude:     exports_files(["image.digest", "digest"]) in BUILD, theoretical BUILD commented out below:
+    repository_ctx.file("image/BUILD", """
+package(default_visibility = ["//visibility:public"])
+
+load("@io_bazel_rules_docker//container:import.bzl", "container_import")
+
+container_import(
```
xingao267: Does this work? If it's broken.

I would probably make it a filegroup target like this
```
filegroup(
    name = "image",
    srcs = glob([
        "image/**",
    ]),
)
```

So the files can be referred.

xingao267:
Also add at least one instance of the new_container_pull in WORKSPACE and build in CI by adding @<your_target_name>//... in https://github.com/bazelbuild/rules_docker/blob/master/.bazelci/presubmit.yml

container/new_pull.bzl
xwinxu: What I currently have works (at least no errors when I called the rule, and it pulled correctly). But I can try the filegroup as well

container/new_pull.bzl
xwinxu: Yes we have index.json, which is an entrypoint and lists the manifests of the images in the layout pulled. We would have to take the code in index.json and go to the corresponding file in blobs/sha256 to get the same digest as image/digest pulled by python puller.

xwinxu: But we don't have a file like the python where the content is just the digest.

container/new_pull.bzl
```diff
@@ -0,0 +1,208 @@
...
+        args += [
+            "-name",
+            "{registry}/{repository}@{digest}".format(
+                registry = repository_ctx.attr.registry,
+                repository = repository_ctx.attr.repository,
+                digest = repository_ctx.attr.digest,
+            ),
+        ]
+    else:
+        tag_ = repository_ctx.attr.tag
```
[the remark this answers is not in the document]
xwinxu: Hmm I remember changing this for a reason but you're right, it's defaulted to latest so I will change it back

.bazelci/presubmit.yml
```diff
@@ -29,6 +29,9 @@ platforms:
     # TODO(xingao): Remove after https://github.com/bazelbuild/rules_scala/issues/644 is addressed.
     - "--incompatible_disallow_legacy_javainfo=false"
   ubuntu1804:
+    build_targets:
```
xingao267: No need for these as you added the tests and they are already run on CI.

container/new_pull.bzl
```diff
@@ -0,0 +1,169 @@
...
+    ),
+}
+
+def _impl(repository_ctx):
+    """Core implementation of container_pull."""
+
+    # Add an empty top-level BUILD file.
+    repository_ctx.file("BUILD", "")
+
+    # TODO(xwinxu): change container_import rule to have an oci attribute to account for new OCI Image Layout
```
xingao267: Remove these two lines. And add a comment stating that this is currently just a file_group target and cannot be used by other rules_docker rules. We will implement a new_container_import rule to understand the oci layout.

container/new_pull.bzl
```diff
@@ -0,0 +1,169 @@
...
+    # currently exclude:     exports_files(["image.digest", "digest"]) in BUILD
+    repository_ctx.file("image/BUILD", """
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "image_new",
```
xingao267: name = "image"

container/new_pull.bzl
```diff
@@ -0,0 +1,169 @@
...
+}
+
+def _impl(repository_ctx):
+    """Core implementation of container_pull."""
+
+    # Add an empty top-level BUILD file.
+    repository_ctx.file("BUILD", "")
+
+    # TODO(xwinxu): change container_import rule to have an oci attribute to account for new OCI Image Layout
+    # currently exclude:     exports_files(["image.digest", "digest"]) in BUILD
```
xingao267: Remove this line

container/new_pull.bzl
```diff
@@ -0,0 +1,169 @@
...
+    repository_ctx.file("image/BUILD", """
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
+    name = "image_new",
+    srcs = glob(["image/**"]),
+)
+exports_files(["index.json"])
```
xingao267: Export all files.

container/new_pull.bzl
```diff
@@ -0,0 +1,169 @@
...
+
+    # Add an empty top-level BUILD file.
+    repository_ctx.file("BUILD", "")
+
+    # TODO(xwinxu): change container_import rule to have an oci attribute to account for new OCI Image Layout
+    # currently exclude:     exports_files(["image.digest", "digest"]) in BUILD
+    repository_ctx.file("image/BUILD", """
+package(default_visibility = ["//visibility:public"])
+
+filegroup(
```
xingao267: ummmm maybe we don't need the filegroup at all. Just export the files.

Issue #903, opened 2019-06-19
container_test does not work with tarball image when driver is "docker"


xingao267:
ERROR: /usr/local/google/home/xingao/gitrepos/rules_docker/tests/docker/BUILD:361:2: in container_bundle_ rule //tests/docker:test_dockerfile_image_kaniko_latest.image: 
Traceback (most recent call last):
	File "/usr/local/google/home/xingao/gitrepos/rules_docker/tests/docker/BUILD", line 361
		container_bundle_(name = 'test_dockerfile_image_kaniko_latest.image')
	File "/usr/local/google/home/xingao/gitrepos/rules_docker/container/bundle.bzl", line 64, in _container_bundle_impl
		_incr_load(ctx, images, ctx.outputs.executable, s...)
	File "/usr/local/google/home/xingao/gitrepos/rules_docker/container/layer_tools.bzl", line 182, in _incr_load
		zip(image["diff_id"], image["unzipped_..."])
	File "/usr/local/google/home/xingao/gitrepos/rules_docker/container/layer_tools.bzl", line 182, in zip
		image["diff_id"]
key "diff_id" not found in dictionary

xingao267:
This is fixed by #905

PR #905, opened 2019-06-19
Add better error message when tar is used in container_test docker mode.

xingao267:


xingao267:
Addresses https://github.com/bazelbuild/rules_docker/issues/903. That use case is not supported. Add better error message and advise users to container_load the tarball first.

PR #907, opened 2019-06-20
Add new Docker tarball feature to Go puller binary

xwinxu:
Default is `"OCI"` format, can also specify Docker tarball only with `"Docker"` or both formats with `"Both"`. All images are still pulled into the image/ directory to remain consistent.

PR #910, opened 2019-06-20
Add back reference to cache in new go puller binary

xwinxu:
Now that this exists in the internal codebase.

### DIFF
--- a/container/go/cmd/puller/puller.go
+++ b/container/go/cmd/puller/puller.go
@@ -13,7 +13,6 @@
 // limitations under the License.
 //////////////////////////////////////////////////////////////////////
 // This binary pulls images from a Docker Registry using the go-containerregistry as backend.
-// The pulled image is in OCI Image Format and this binary can also accomodate manifest lists.
 // Unlike regular docker pull, the format this package uses is proprietary.
 
 package main
@@ -23,6 +22,8 @@ import (
 	"log"
 	ospkg "os"
 	"strings"
+
+	"github.com/pkg/errors"
 
 	"github.com/bazelbuild/rules_docker/container/go/pkg/oci"
 	"github.com/google/go-containerregistry/pkg/authn"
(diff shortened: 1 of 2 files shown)


container/go/cmd/puller/puller.go
```diff
@@ -60,6 +62,9 @@ func pull(imgName, dstPath string, platform v1.Platform) {
 	if err != nil {
 		log.Fatalf("reading image %q: %v", ref, err)
 	}
+	if cachePath != "" {
+		i = cache.Image(i, cache.NewFilesystemCache(cachePath))
```
smukherj1: You are not using `i` anywhere. Change this to just `cache.Image(i, cache.NewFilesystemCache(cachePath))`

container/go/cmd/puller/puller.go
```diff
@@ -46,8 +48,8 @@ var (
 
 // NOTE: This function is adapted from https://github.com/google/go-containerregistry/blob/master/pkg/crane/pull.go
 // with slight modification to take in a platform argument.
-// Pull the image with given <imgName> to destination <dstPath> with optional required platform specifications.
-func pull(imgName, dstPath string, platform v1.Platform) {
+// Pull the image with given <imgName> to destination <dstPath> with optional cache files and required platform specifications.
+func pull(imgName, dstPath, cachePath string, platform v1.Platform) {
```
smukherj1: I'm seeing calls to log.Fatalf inside this function. Convention in Go is to never panic or die inside functions. Return an error here instead and handle the error in a log.Fatalf call in the main function.

container/go/cmd/puller/puller.go
xwinxu: Sorry typo, supposed to be reassigning img, not i, will make sense once my tarball PR is merged.

container/go/cmd/puller/puller.go
xingao267: +1 I'm gonna approve and merge this one https://github.com/bazelbuild/rules_docker/pull/907/files and then you can fix all the returns in this PR.

container/go/cmd/puller/puller.go
```diff
@@ -44,27 +47,36 @@ var (
...
+// image was (probably) never tagged with this, but lets us avoid applying the
+// ":latest" tag which might be misleading.
+const iWasADigestTag = "i-was-a-digest"
+
 // NOTE: This function is adapted from https://github.com/google/go-containerregistry/blob/master/pkg/crane/pull.go
 // with slight modification to take in a platform argument.
-// Pull the image with given <imgName> to destination <dstPath> with optional required platform specifications.
-func pull(imgName, dstPath string, platform v1.Platform) {
+// Pull the image with given <imgName> to destination <dstPath> with optional cache files and required platform specifications.
+func pull(imgName, dstPath, cachePath string, platform v1.Platform) (err error) {
```
smukherj1: Just `func pull(imgName, dstPath, cachePath string, platform v1.Platform) error` will do because you are not really using the unnamed return.

container/go/cmd/puller/puller.go
```diff
@@ -94,7 +106,9 @@ func main() {
 		Features:     strings.Fields(*features),
 	}
 
-	pull(*imgName, *directory, platform)
+	if err := pull(*imgName, *directory, *cachePath, platform); err != nil {
+		log.Fatalln("Image pull was unsuccessful: %v", err)
```
smukherj1: %v does nothing when used with log.Fatalln. Either remove the %v or use log.Fatalf

container/go/cmd/puller/puller.go
```diff
@@ -79,12 +91,12 @@ func main() {
 	}
 
 	// If the user provided a client config directory, instruct the keychain resolver
-	// to use it to look for the docker client config.
+	// to use it to look for the docker client config
```
smukherj1: nit: period at the end of comment.

container/go/cmd/puller/puller.go
```diff
@@ -79,12 +91,12 @@ func main() {
...
 	if *clientConfigDir != "" {
 		ospkg.Setenv("DOCKER_CONFIG", *clientConfigDir)
 	}
 
-	// Create a Platform struct with given arguments.
+	// Create a Platform struct with arguments
```
smukherj1: nit: period at the end of comment.

denyska:
This seems to be only writing to cache and not reading from it. Is it accurate ?

PR #911, opened 2019-06-20
Implement new container_push in go

xiaohegong:
First version of new `container_push`:
 - New starlark rule `new_container_push`
 - pusher.go 

This version supports pushing an OCI layout or a Docker tarball. Some of the old features (e.g., push by layers) are not implemented  in this version.


container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// NOTE: This function is adapted from https://github.com/google/go-containerregistry/blob/master/pkg/crane/push.go
+// with modification for option to push OCI layout or Docker tarball format .
+// Push the image from <src> to destination <dst> with specified format (oci or docker).
+func push(dst, src string, isOCI bool) {
+	// Read an OCI index or a Docker tarball from src.
+	var img v1.Image
+	var err error
+	if isOCI {
```
smukherj1: Control flow can be simplified here. Pass in v1.Image to `push` and move the logic that determines the format to read to a separate function as follows:
```
func readImage(src, format string) (v1.Image, error) {
    if format == "oci" {
        return oci.Read(src)
    }
   if format == "docker" {
       return tarball.ImageFromPath(src, nil)
    }
    return nil, fmt.Errorf("unknown image format %q", format)
}
```

call the above from main as follows:
```
img, err := readImage(*src, *format);
if err != nil {
    log.Fatalf("Error reading image from %s as format %s: %v", *src, *format, err)
}
if err := push(img, dst); err != nil {
   ...
}
```

container/new_push.bzl
```diff
@@ -0,0 +1,161 @@
...
+    pusher_args += ["-dst", "{registry}/{repository}:{tag}".format(
+        registry = registry,
+        repository = repository,
+        tag = tag,
+    )]
+
+    # Find and set src to correct paths depending the image format to be pushed
+    if ctx.attr.format == "oci":
+        for f in ctx.files.image:
+            if f.basename == "index.json":
```
smukherj1: Need an error case if we never find index.json after looping over all the files in ctx.files.image. Use a variable to check if the if condition was hit and add something like
`fail("Did not find an index.json in the image attribute specified to {}".format(ctx.label))`

Ideally, we would also want to print the label name corresponding to `ctx.attr.image` in the error message to provide more context.

container/new_push.bzl
```diff
@@ -0,0 +1,161 @@
...
+                pusher_args += ["-src", "{index_dir}".format(
+                    index_dir = f.dirname,
+                )]
+                break
+    else:
+        pusher_args += ["-src", str(ctx.files.image[0].path)]
```
smukherj1: Check len(ctx.files.image) > 0 in case the filegroup didn't capture any files.

PR #915, opened 2019-06-21
Implement new container_push

xiaohegong:
First version of new `container_push`:
 - New starlark rule `new_container_push`
 - pusher.go 

This version supports pushing an OCI layout or a Docker tarball. Some of the old features (e.g., push by layers) are not implemented  in this version.


container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/google/go-containerregistry/pkg/name"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/remote"
+	"github.com/google/go-containerregistry/pkg/v1/tarball"
+	"github.com/pkg/errors"
+)
+
+var (
+	dst             = flag.String("dst", "", "The destination location including repo and digest/tag of the docker image to push. Supports fully-qualified tag or digest references.")
+	src             = flag.String("src", "", "The path of the source index relative to the execution workspace.")
```
smukherj1: The fact that the path is relative to the execution workspace is an implementation detail of this binary being run from Bazel. Over here, `src` can either be an absolute path or relative to the working directory from where the binary is being run. I would just say `Path to the directory which has the image index that will be pushed.`

container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,95 @@
...
+	format          = flag.String("format", "", "The form to push, oci or docker (A tarball).")
```
smukherj1: It's technically the format of the source image. The format of the pushed image is not really controlled by this binary. It's determined by GCR (I think).

So I would say `The format of the source image, oci|docker. The docker format should be a tarball of the image as generated by docker save.`

container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,95 @@
...
+	clientConfigDir = flag.String("client-config-dir", "", "The path to the directory where the client configuration files are located. Overiddes the value from DOCKER_CONFIG.")
+)
+
+func main() {
+	flag.Parse()
+	log.Println("Running the Image Puller to pull images from a Docker Registry...")
```
smukherj1: Change this to say pusher

container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	img, err := readImage(*src, *format)
+	if err != nil {
+		log.Fatalln("Error reading from %s: %v", *src, err)
+	}
+
+	if err := push(*dst, img); err != nil {
+		log.Fatalln("Error pushing image to %s: %v", *dst, err)
+	}
```
smukherj1: Suggest adding a log here that prints `Successfully pushed <format> image from <src> to <dst>`

container/new_push.bzl
```diff
@@ -0,0 +1,166 @@
...
+    if ctx.attr.format == "oci":
+        found = false
+        for f in ctx.files.image:
+            if f.basename == "index.json":
+                pusher_args += ["-src", "{index_dir}".format(
+                    index_dir = f.dirname,
+                )]
+                found = True
+        if not found:
+            fail("Did not find an index.json in the image attribute specified to {}".format(ctx.attr.image))
```
smukherj1: nit: here I would say
```
fail("Did not find an index.json in the image attribute {} specified to {}".format(ctx.attr.image, ctx.label))
```

Here "to" should refer to the `new_container_push` rule itself and the label of the image attribute will be printed right after the text "attribute" in the error log.

container/new_push.bzl
```diff
@@ -0,0 +1,166 @@
...
+    elif len(ctx.files.image) > 0:
```
smukherj1: We should check the number of images provided is exactly one when format is `docker`.
Instead of having the logic for format `docker` under the elif & else clauses, I would move this under a separate if condition for better readability. So something like:
```
if ctx.attr.format == "docker":
   if len(ctx.files.image) == 0:
       fail("Attribute image {} to {} did not contain an image tarball".format(ctx.attr.image, ctx.label))
   if len(ctx.files.images) > 1:
       fail("Attribute image {} to {} had {} files. Expected exactly 1".format(ctx.attr.image, ctx.label, len(ctx.files.images))
   pusher_args += ["-src", str(ctx.files.image[0].path)]
```

container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,97 @@
...
+
+	// If the user provided a client config directory, instruct the keychain resolver
+	// to use it to look for the docker client config.
+	if *clientConfigDir != "" {
+		os.Setenv("DOCKER_CONFIG", *clientConfigDir)
+	}
+
+	img, err := readImage(*src, *format)
+	if err != nil {
+		log.Fatalln("Error reading from %s: %v", *src, err)
```
smukherj1: Use Fatalf here

container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if err := push(*dst, img); err != nil {
+		log.Fatalln("Error pushing image to %s: %v", *dst, err)
```
smukherj1: Use Fatalf here

container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,97 @@
...
+// Push the given image to destination <dst>.
+func push(dst string, img v1.Image) error {
+	// Push the image to dst.
+	ref, err := name.ParseReference(dst)
+	if err != nil {
+		return errors.Wrapf(err, "parsing tag %q", dst)
+	}
+
+	if err := remote.Write(ref, img, remote.WithAuthFromKeychain(authn.DefaultKeychain)); err != nil {
+		return errors.Wrapf(err, "")
```
smukherj1: Add context here `errors.Wrapf(err, "unable to push image to %s", dst)`

container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,97 @@
...
+		log.Fatalln("Error pushing image to %s: %v", *dst, err)
+	}
+
+	log.Println("Successfully pushed %s image from %s to %s", *format, *src, *dst)
+}
+
+// NOTE: This function is adapted from https://github.com/google/go-containerregistry/blob/master/pkg/crane/push.go
+// with modification for option to push OCI layout or Docker tarball format .
+// Push the given image to destination <dst>.
+func push(dst string, img v1.Image) error {
```
smukherj1: Go function docstrings for functions, structs and struct fields begin with the name of the entity being document. So here, the comment would begin as follows:
```
// push pushes the given image to the given destination.
// Note: ...
```

container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,97 @@
...
+	}
+
+	if err := remote.Write(ref, img, remote.WithAuthFromKeychain(authn.DefaultKeychain)); err != nil {
+		return errors.Wrapf(err, "")
+	}
+
+	return nil
+}
+
+// Return a v1.Image after reading an OCI index or a Docker tarball from src.
```
smukherj1: `// readImage returns ...`

container/go/cmd/pusher/pusher.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// NOTE: This function is adapted from https://github.com/google/go-containerregistry/blob/master/pkg/crane/push.go
+// with modification for option to push OCI layout or Docker tarball format .
+// Push the given image to destination <dst>.
+func push(dst string, img v1.Image) error {
+	// Push the image to dst.
+	ref, err := name.ParseReference(dst)
+	if err != nil {
+		return errors.Wrapf(err, "parsing tag %q", dst)
```
smukherj1: Change `parsing tag %q` to `error parsing %q as an image reference`. Error messages should be complete statements on their own and add as much information as possible from the context available to them.

PR #918, opened 2019-06-21
Add extensive Write.go tests with small image

xwinxu:


container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,179 @@
...
+	_, err = f.Readdirnames(1)
+	if err == io.EOF {
+		return fmt.Errorf("blobs directory is empty: %v", err)
+	}
+	if err = compareSHA(path.Join(blobs, shaPth), "blobs file at "+shaPth, content["blobs/sha256/9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c"].Hex); err != nil {
+		return fmt.Errorf("sha256 hex code did not match %v", err)
+	}
+
+	// validate the index.json file
+	idx, err := os.Stat(path.Join(tmp, index))
```
xingao267: All these checks can be grouped into a single function, maybe called validateFile

xingao267: including the sha comparison

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,179 @@
...
+	}
+	if idx.Size() == 0 {
+		return fmt.Errorf("error occured: empty index.json, size: %d", idx.Size())
+	}
+	if err = compareSHA(path.Join(tmp, index), "oci-layout", content["index.json"].Hex); err != nil {
+		return fmt.Errorf("sha256 hex code did not match, %v", err)
+	}
+
+	// validate the oci-layout file
+	oci, err := os.Stat(tmp + ocif)
```
xingao267: use path.Join() for tmp + ocif, that way, you can also get rid of the /. And you don't need the tag in compareSHA function.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
+package oci
```
xingao267: Add license

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
...
+			img, err := tarball.ImageFromPath(wt.dataPath, nil)
+			if err != nil {
+				t.Errorf("error loading image from testdata: %v", err)
+			}
+
+			if err := assertImageIndexValid(tmp, img); err != nil {
+				t.Errorf("image written is not a valid index: %v", err)
+			}
+
+			if err = assertOCIFormat(tmp, img, wt.content, wt.shaPath); err != nil {
```
xingao267: do err := assert... here?

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
...
+				t.Errorf("image is not in oci format: %v", err)
+			}
+		})
+	}
+}
+
+func assertImageIndexValid(tmp string, img v1.Image) error {
+	if err := Write(img, tmp); err != nil {
+		return fmt.Errorf("error writing image to temp path %s", tmp)
```
xingao267: errors.Wrapf

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	written, err := layout.ImageIndexFromPath(tmp)
+	if err != nil {
+		return fmt.Errorf("failed to write image to temp path %s", tmp)
```
xingao267: errors.Wrapf

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	if err := validate.Index(written); err != nil {
+		return fmt.Errorf("validate.Index() = %v", err)
```
xingao267: errors.Wrapf

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
...
+
+			if err = assertOCIFormat(tmp, img, wt.content, wt.shaPath); err != nil {
+				t.Errorf("image is not in oci format: %v", err)
+			}
+		})
+	}
+}
+
+func assertImageIndexValid(tmp string, img v1.Image) error {
+	if err := Write(img, tmp); err != nil {
```
xingao267: It's a little weird having this write action in this function. I would probably put this Write() call in TestWrite above directly. And then have one single function called in the TestWrite to do all the checks, maybe called assertValidImageOCILayout() which contains the rest of assertImageIndexValid + assertOCIFormat. So basically, rename assertOCIFormat and move the rest of assertImageIndexValid there.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	if err := validate.Index(written); err != nil {
+		return fmt.Errorf("validate.Index() = %v", err)
+	}
+
+	return nil
+}
+
+func assertOCIFormat(tmp string, img v1.Image, content map[string]v1.Hash, shaPth string) error {
```
xingao267: doc

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
...
+	if err := Write(img, "/usr/local/google/home/xwinxu/Desktop/temp/temp"); err != nil {
```
xingao267: Leftover from local testing?

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
...
+	// Under blobs/, for each image:
+	// 	   one file for each layer, config blob, and manifest blob
+	blobs := path.Join(tmp, blob)
+	file := path.Join(blobs, "sha256", "9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c")
+	if err := validateDir(blobs, content["blobs/sha256/9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c"].Hex, tmp, file); err != nil {
+		return fmt.Errorf("failed to validate blobs/ directory: %v", err)
+	}
+
+	// validate the index.json file
+	if err := validateFile(path.Join(tmp, index), index, content["index.json"].Hex, tmp); err != nil {
```
xingao267: Try to put all these checks into a loop that iterates over the keys of your content map. Otherwise, we are duplicate and hardcode the keys again here.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,188 @@
...
+	}
+
+	// validate that the contents of blobs/ were written in the following format:
+	// Top level:
+	//     oci-layout, index.json, blobs/
+	// Under blobs/, for each image:
+	// 	   one file for each layer, config blob, and manifest blob
+	blobs := path.Join(tmp, blob)
+	file := path.Join(blobs, "sha256", "9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c")
+	if err := validateDir(blobs, content["blobs/sha256/9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c"].Hex, tmp, file); err != nil {
```
xingao267: And you don't have to combine the check for dir and file in one function here. You can check the dir only in one call. And then have a loop that checks all the files that you want to check that are defined in the content map.

PR #924, opened 2019-06-24
Prefer docker_run_flags from ctx when present

### LINKED ISSUE #923: docker_run_flags impossible to override with subsequent container_image targets
In #505, the behavior of `container_image` was changed to prefer a base image's `docker_run_flags` over the top-most `container_image`.

This seems backwards to me, as it's now impossible to override this value, and we end up duplicating our `container_image` definitions with the same `layers` attributes (and any others we want to set) just so we can change the `docker_run_flags` on them.

I think it makes sense to prefer the `docker_run_flags` of the base _only_ in the case where the current `container_image` is relying on the default value.

Are there folks relying on the current behavior? I would expect not, but maybe there's a deeper reason here I'm not understanding.

clintharrison:
This changes the behavior to prefer `docker_run_flags` from the "top-most" `container_image` target.

It also adds some tests to ensure that the default/custom value is correctly inherited between multiple `container_image`s.

Fixes #923.


testdata/BUILD
```diff
@@ -938,3 +938,22 @@ go_binary(
     name = "main",
     srcs = ["main.go"],
 )
+
+container_image(
```
xingao267: Looks good! Could you also add a test case the default one is used when docker_run_flags is not specified?

clintharrison:
Thanks for the quick review!

The `container_image` targets with `legacy_run_behavior = False` were also causing failures as a result of the `test_bazel_run_docker_build_incremental` (and similar) tests -- they can't actually get run as a container because they don't have a real entrypoint/cmd. I've excluded them from the `bazel query`, so it's back to green now.

clintharrison:
Oops, I thought it would automatically merge, but I seem to have messed up the workflow 🙂

PR #931, opened 2019-06-26
Add tests for new container_push 

xiaohegong:
- Added e2e tests for `new_container_push`
- Updated image file directory in `new_container_push` to resolve to runtime path.
  - Fixes the issue of "invalid file directory"


container/go/cmd/pusher/pusher.go
```diff
@@ -56,6 +57,11 @@ func main() {
 		os.Setenv("DOCKER_CONFIG", *clientConfigDir)
 	}
 
+	// Ensure src is a directory, trim basename index.json if not.
+	if filepath.Base(*src) == "index.json" {
+		*src = filepath.Dir(*src)
+	}
+
```
xiaohegong: This trims `src` to a directory path. I need to do this because ` get_runfile_path(ctx, index.json)` (what we used to find absolute path to an oci-layout in `new_pusher.bzl`) returns a path to `index.json`. I could not think of a good way for this.

container/new_push.bzl
```diff
@@ -65,7 +65,7 @@ def _impl(ctx):
         for f in ctx.files.image:
             if f.basename == "index.json":
                 pusher_args += ["-src", "{index_dir}".format(
-                    index_dir = f.dirname,
+                    index_dir = _get_runfile_path(ctx, f),
```
xingao267: you can write a new _get_runfile_dir_path that can give you the dir of f. 

```
def _get_runfile_dir_path(ctx, f):
    return "${RUNFILES}/%s" % runfile(ctx, f).dirname

```

container/go/cmd/pusher/pusher.go
xingao267: See comment below.

PR #933, opened 2019-06-26
Add extensive Write.go tests with small image

xwinxu:
Tests extensively for the OCI layout written.

PR #934, opened 2019-06-26
Refactor write.go test with small image

xwinxu:
Tests extensively for the OCI layout written.


container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+	"github.com/pkg/errors"
+)
+
+var (
+	blob     = "blobs/"
+	testPath = "testdata/test_write1"
+)
+
+var testCases = []struct {
+	// name is the name of the test case
```
smukherj1: All go comments should end with periods. Same for every other attribute.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+		"sha256/9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c",
+		map[string]v1.Hash{
+			"blobs/sha256/9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c": v1.Hash{Algorithm: "sha256", Hex: "9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c"},
+			"index.json": v1.Hash{Algorithm: "sha256", Hex: "ce6ed2e6817add8a5a0e94f4c2ac00360c7ca2cc98c4972b86530f14de622508"},
+			"oci-layout": v1.Hash{Algorithm: "sha256", Hex: "b66dbb27a73334db6ac9c030475837bd7f4472d835c72b2360534b203edce6cb"},
+		},
+	},
+}
+
+// TestWrite checks to ensure that the v1.Image written is in OCI format by validating image output format and content
```
smukherj1: All go comments should end with periods. Here and many more below.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+func TestWrite(t *testing.T) {
+	for _, wt := range testCases {
+		t.Run(wt.name, func(t *testing.T) {
+			tmp, err := ioutil.TempDir("", "write-oci-index-test")
+			if err != nil {
+				t.Errorf("Write(%s): %v", os.TempDir(), err)
```
smukherj1: this is not realated to "Write" really. I would recommend changing the message to "Unable to create a TempDir called `write-oci-index-test`"

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+			}
+
+			defer os.RemoveAll(tmp)
+
+			img, err := tarball.ImageFromPath(wt.dataPath, nil)
+			if err != nil {
+				t.Errorf("error loading image from testdata: %v", err)
```
smukherj1: Begin unit test errors with capital letters.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+			}
+
+			if err := Write(img, tmp); err != nil {
+				t.Errorf("error writing image to temp path: %v", err)
```
smukherj1: Begin unit test errors with capital letters and below as well.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+	defer f.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return fmt.Errorf("sha256 hashing failed: %v", err)
+	}
+	got := hex.EncodeToString(h.Sum(nil))
+	if strings.Compare(got, want) != 0 {
+		return fmt.Errorf("sha256 hex code did not match, wanted: %s got: %s", want, got)
```
smukherj1: got before want as per Go convention.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+func compareSHA(src, want string) error {
+	f, err := os.Open(src)
+	if err != nil {
+		return fmt.Errorf("failed to open file: %v", err)
+	}
+	defer f.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return fmt.Errorf("sha256 hashing failed: %v", err)
```
smukherj1: mention the name of the src file in the error for context.

container/go/pkg/oci/write_test.go
smukherj1: mention the name of the src file in the error for context.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	return nil
+}
+
+// validates a file being tested for by checking file is not empty or corrupt
+func validateFile(src, file, want, dir string) error {
+	f, err := os.Stat(src)
+	if err != nil {
+		return fmt.Errorf("could not obtain stat: %v", err)
```
smukherj1: mention the name of the src file in the error for context. Also we usually say "unable to stat %s".

Use errors.Wrapf here.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+	}
+
+	if os.IsNotExist(err) {
+		return fmt.Errorf("wrong OCI output format; got: %v, missing: %s", err, file)
```
smukherj1: errors.Wrapf here and below.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+	blob     = "blobs/"
+	testPath = "testdata/test_write1"
+)
+
+var testCases = []struct {
+	// name is the name of the test case
+	name string
+	// dataPath is where the test data is found
+	dataPath string
+	// the sha code for the manifest
```
smukherj1: The comment mentions "code" which makes me thing the value here is a digest but the variable name suggests its a path. I think this is actually the "manifestPath" which is relative to the "dataPath". So I would rename this variable to "manifestPath" and explain this in the comment. The fact that the manifestPath in OCI has a file named with the sha256 digest is an implementation detail of the OCI format.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,156 @@
...
+	shaPath string
+	// content is the oci layout content where the key is the file path and value is sha256 sum
```
smukherj1: I would explain this more using the other variables in this struct. content maps from paths relative to dataPath to the expected sha256 digest of that file. I would also change the name of this variable to `wantContent` to make it explicit that this the "expected" values section of your unit test.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+var testCases = []struct {
+	// name is the name of the test case.
+	name string
+	// dataPath is where the test data is found.
+	dataPath string
+	// path to the image manifest relative to the dataPath.
+	manifestPath string
+	// content is the OCI layout content where the key is the file path and value is the sha256 sum.
```
smukherj1: I would simplify this comment further to
```
// content is a mapping from path to a file expected to exist in the image relative to "dataPath" to
// the expected sha256 digest of that file.
```

This clearly states what the value of the key & data portions of this map are. The OCI bit is an implementation detail. I would mention the "OCI" part in the comments section for each testCase value below if you want.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,157 @@
...
+	// it contains a mapping of the sha256 code of the different files inside the OCI layout.
+	expectedContent map[string]v1.Hash
+}{
+	{
+		"alpine_linux",
+		"testdata/test_write1/small_linux.tar",
+		"sha256/9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c",
+		map[string]v1.Hash{
+			"blobs/sha256/9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c": v1.Hash{Algorithm: "sha256", Hex: "9a96f3888ebad00d46bca04ccb591e70d091624835998668af551f48512d9b5c"},
```
smukherj1: If you want, mention the fact that these keys correspond to the OCI format.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,157 @@
...
+// assertValidImageOCILayout checks that the index written is not corrupt
+// and that the contents of the OCI layout are all present and non-corrupt.
+func assertValidImageOCILayout(ociDir string, img v1.Image, content map[string]v1.Hash, shaPth string) error {
+	written, err := layout.ImageIndexFromPath(ociDir)
+	if err != nil {
+		return errors.Wrapf(err, "failed to write image to temp path %s", ociDir)
+	}
+
+	if err := validate.Index(written); err != nil {
+		return errors.Wrapf(err, "validate.Index() =")
```
smukherj1: error messages in Go should be standalone statements on their own and have as much context as possible. Here I would suggest saying:
`errors.Wrapf(err, "failed to validate the image written to %s using validate.Index()", ociDir)`

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	// validate that the contents of blobs/ were written in the following format:
+	// Top level:
+	//     oci-layout, index.json, blobs/.
+	// Under blobs/, for each image:
+	// 	   one file for each layer, config blob, and manifest blob.
+	for k := range content {
```
smukherj1: You are using both the key & value below so I would get them both here:
```
for wantFile, wantDigest := range wantContent {
    if err := validateFile(path.Join(ociDir, wantFile), wantFile, wantDigest, ociDir); err != nil {
```

The above variable naming makes the code more readable as it's clearer what `wantContent` uses as key & value without having to go to the documentation.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,157 @@
...
+			if err := assertValidImageOCILayout(tmp, img, wt.expectedContent, wt.manifestPath); err != nil {
+				t.Errorf("Image index written is not a valid OCI format: %v", err)
+			}
+		})
+	}
+}
+
+// assertValidImageOCILayout checks that the index written is not corrupt
+// and that the contents of the OCI layout are all present and non-corrupt.
+func assertValidImageOCILayout(ociDir string, img v1.Image, content map[string]v1.Hash, shaPth string) error {
```
smukherj1: I would name `content` as `wantContent` which makes the purpose of the variable clearer from the name.

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,157 @@
...
+		return errors.Wrapf(err, "failed to open file")
+	}
+	defer f.Close()
+
+	h := sha256.New()
+	if _, err := io.Copy(h, f); err != nil {
+		return errors.Wrapf(err, "sha256 hashing of file %s failed", src)
+	}
+	got := hex.EncodeToString(h.Sum(nil))
+	if strings.Compare(got, want) != 0 {
```
smukherj1:
```
if got := hex.EncodeToString(h.Sum(nil)); strings.Compare(got, want) != 0 {
```

container/go/pkg/oci/write_test.go
```diff
@@ -0,0 +1,157 @@
...
+)
+
+var testCases = []struct {
+	// name is the name of the test case.
+	name string
+	// dataPath is where the test data is found.
+	dataPath string
+	// path to the image manifest relative to the dataPath.
+	manifestPath string
+	// expectedContent is a mapping form a file expected to exist in the image
```
smukherj1: mapping from

xingao267:
This replaces https://github.com/bazelbuild/rules_docker/pull/933

Issue #935, opened 2019-06-26
Why does java_image rule needs to pull in golang rules?


neakor:
I have a simple example project setup just testing out the `java_image` rule. The project contains only 3 Java files.

My WORKSPACE is setup as
```
# Download the rules_docker repository at release v0.8.0
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "io_bazel_rules_docker",
    sha256 = "3556d4972571f288f8c43378295d84ed64fef5b1a875211ee1046f9f6b4258fa",
    strip_prefix = "rules_docker-0.8.0",
    urls = ["https://github.com/bazelbuild/rules_docker/archive/v0.8.0.tar.gz"],
)

# Java language rules.
load(
    "@io_bazel_rules_docker//repositories:repositories.bzl",
    container_repositories = "repositories",
)
container_repositories()
load(
    "@io_bazel_rules_docker//java:image.bzl",
    _java_image_repos = "repositories",
)
_java_image_repos()
load(
    "@io_bazel_rules_docker//container:container.bzl",
    "container_pull",
)
container_pull(
  name = "java_base",
  registry = "gcr.io",
  repository = "distroless/java",
  # 'tag' is also supported, but digest is encouraged for reproducibility.
  digest = "sha256:deadbeef",
)
```

My BUILD file is
```
load("@io_bazel_rules_docker//java:image.bzl", "java_image")

java_image(
    name = "ProjectRunner",
    srcs = glob(["src/main/java/com/example/ProjectRunner.java"]),
    deps = [":greeter", "//src/main/java/com/example/cmdline:runner"],
    # layers = [":java_image_library"],
    main_class = "com.example.ProjectRunner",
)

java_library(
    name = "greeter",
    srcs = ["src/main/java/com/example/Greeting.java"],
    visibility = ["//src/main/java/com/example/cmdline:__pkg__"],
)
```

When I attempt to run `bazel build //...` I get an error saying
```
ERROR: Analysis of target '//:ProjectRunner' failed; build aborted: error loading package '@io_bazel_rules_docker//': in /private/var/tmp/_bazel_yi.wang/d1962d194a91b276aac98fc45acdff1f/external/baze
[...]

PR #936, opened 2019-06-27
refactor go dependencies to a separate bzl file

nlopezgi:
* Also update readme with instructions.
* Addresses https://github.com/bazelbuild/rules_docker/issues/935
* Will need to partially revert https://github.com/bazelbuild/bazel-toolchains/pull/593/files once this is in.
* Also reduced the timeout of the cloudbuild to create go puller binary from 6h to 1h
* Also extended timeout for run_instruction_arbitrary/cloudbuild.yaml to 20 minutes


nlopezgi:
We will need to release 0.8.1 after this PR is in

WORKSPACE
```diff
@@ -386,17 +373,3 @@ load("@bazel_toolchains//rules:rbe_repo.bzl", "rbe_autoconfig")
 rbe_autoconfig(
     name = "buildkite_config",
 )
-
-# gazelle:repo bazel_gazelle
```
smukherj1: This needs to remain in WORKSPACE. Otherwise running `bazel run //:gazelle` will fail to update BUILD files.

go/image.bzl
```diff
@@ -39,10 +39,10 @@ load(":go.bzl", "DIGESTS")
 def repositories():
     """Import the dependencies of the go_image rule.
 
-    Call the core "repositories" function to reduce boilerplate. This is
+    Call the core "go_deps" function to reduce boilerplate. This is
     idempotent if folks call it themselves.
     """
-    _repositories()
+    _go_deps()
```
xingao267: Do you need both repositories from repositories.bzl and go_repositories.bzl?

WORKSPACE
nlopezgi: fixed

go/image.bzl
nlopezgi: container_repositories must be called from WORKSPACE directly (updated readme) for all image rules now

PR #938, opened 2019-06-27
migrating content from e2e.sh to separate, more granular, GCB builds

nlopezgi:
Moving cc_image and go_image from e2e.sh to separate GCB configs.
Will continue migrating more content from e2e tests in upcoming PRs.
Most content in new sh files is just copied from e2e.sh with minor adaptations to run safely on GCB.


testing/e2e.sh
```diff
@@ -246,53 +246,12 @@ EOF
   rm -f output.txt
```
xingao267: import the util.sh in this as well?

testing/e2e/cc_image.sh
```diff
@@ -0,0 +1,41 @@
...
+  cd "${ROOT}"
+  EXPECT_CONTAINS "$(bazel run "$@" testdata:cc_binary_as_image)" "Hello World"
+}
+
+function test_cc_image_wrapper() {
+  cd "${ROOT}"
+  EXPECT_CONTAINS "$(bazel run "$@" testdata:cc_image_wrapper)" "Hello World"
+}
+
+if [[ $# -ne 1 ]]; then
```
xingao267: Add some doc for this.

testing/e2e.sh
nlopezgi: done

testing/e2e/cc_image.sh
nlopezgi: fixed

clintharrison:
@nlopezgi is there a way to make the results public? The e2e tests were previously publicly viewable on Travis, but it seems that's no longer the case when they're run on GCB.

nlopezgi:
hi @clintharrison ,
apologies for the issue. I did not realize the results were not public. I'll try to figure out a make results public. In the meantime, you can run these test locally - https://cloud.google.com/cloud-build/docs/build-debug-locally - (are you able to at least see which trigger failed or do you just see the trigger id?)

clintharrison:
@nlopezgi thanks for looking into it, and that link is definitely helpful, too. Unfortunately it doesn't seem like we're able to see which trigger, the ID is the only public info :\

nlopezgi:
can you let me know the ID of the failing test? Also do let me know if you are able to run any of these locally with the instructions in that link.

clintharrison:
Yep, I can run them locally! I don't actually have a failing ID right now I need to see -- but generally I was just curious what the plan was.

PR #941, opened 2019-06-28
Even more tests moved to GCB from e2e

nlopezgi:
* Noted which test cannot move from e2e.sh for now and why.

### DIFF
--- a/testing/e2e.sh
+++ b/testing/e2e.sh
@@ -18,11 +18,18 @@ source ./testing/e2e/util.sh
 # Must be invoked from the root of the repo.
 ROOT=$PWD
 
+function stop_containers() {
+  docker rm -f $(docker ps -aq) > /dev/null 2>&1 || builtin true
+}
+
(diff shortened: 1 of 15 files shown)


testing/e2e.sh
```diff
@@ -69,54 +63,15 @@ EOF
...
-function test_launcher_image() {
-  cd "${ROOT}"
-  clear_docker
-  EXPECT_CONTAINS "$(bazel run "$@" testdata:launcher_image)" "Launched via launcher!"
-}
-
 function test_java_image() {
   cd "${ROOT}"
-  clear_docker
+  clear_docker_full
```
smukherj1: Why not use the new clear_docker everywhere instead of using clear_docker_full at all?

nlopezgi: Don't want to mess up with the existing e2e.sh functionality

smukherj1: sgtm

PR #942, opened 2019-06-28
refactor tests/docker -> tests/container

nlopezgi:
In preparation to migrating content to /docker I need to clear out the test/docker folder which was misnamed.


tests/container/BUILD
```diff
@@ -1,5 +1,499 @@
...
+    },
+)
+
+docker_push_all(
+    name = "test_docker_push_three_images_bundle",
+    bundle = ":three_images_bundle",
+)
+
+[container_test(
+    name = "test_dockerfile_image_%s" % driver,
```
xingao267: This is testing dockerfile_image from contrib. Probably should be moved to tests/contrib, together with the testing configs.

nlopezgi: done

PR #949, opened 2019-07-02
Add bzl_library for lang/image.bzl

clintharrison:
Stardoc currently requires all dependencies to also have a bzl_library, and it seems lang/image.bzl got missed when the rest of the repo has them, from what I can tell.


clintharrison:
yep, sorry! always forget that step :)

PR #954, opened 2019-07-03
Implement new interface for legacy intermediate format layouts

xiaohegong:
Implement `v1.Image` interface for legacy (MM) intermediate format layouts
 - Such layout requires the following structure
    - `digest`
    - `manifest.json`
    - `config.json`
    - `000.tar.gz`, ......, `N.tar.gz`

Update `new_container_push` and `pusher.go` to incorporate new format.


container/go/pkg/compat/image.go
```diff
@@ -0,0 +1,147 @@
...
+
+// RawManifest returns the serialized bytes of manifest.json metadata.
+func (li *layoutImage) RawManifest() ([]byte, error) {
+	li.manifestLock.Lock()
+	defer li.manifestLock.Unlock()
+	if li.rawManifest != nil {
+		return li.rawManifest, nil
+	}
+
+	b, err := ioutil.ReadFile(li.path.path(manifestFile))
```
xingao267: Add comment for this.

container/go/pkg/compat/index.go
```diff
@@ -0,0 +1,198 @@
...
+		return nil, err
+	}
+	return lp.ImageIndex()
+}
+
+// FromPath reads an legacy intermediate image index at path and constructs a layout.Path object.
+// Naively validates a legacy intermediate layout at <path> by checking if digest, config.json, and manifest.json all exist.
+func FromPath(path string) (Path, error) {
+	var err error
+	_, err = os.Stat(filepath.Join(path, manifestFile))
```
xingao267:
```
if _, err := os.Stat(filepath.Join(path, manifestFile)); err != nil {
    return "", err
}
```

container/go/pkg/compat/image.go
```diff
@@ -0,0 +1,147 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//////////////////////////////////////////////////////////////////////
+// Image for intermediate format used in python containerregistry.
+// Uses the go-containerregistry API as backend.
+
+package compat
```
xingao267: Mention that it's mostly from go-containerregistry's layout implementation with modification to understand rules_docker's legacy intermediate format.

container/go/pkg/compat/index.go
```diff
@@ -0,0 +1,198 @@
...
+// findDescriptor looks for the manifest with digest h in our "index.json" struct and returns its descriptor.
+func (i *intermediateLayout) findDescriptor(h v1.Hash) (*v1.Descriptor, error) {
+	im, err := i.IndexManifest()
+	if err != nil {
+		return nil, err
+	}
+
+	for _, desc := range im.Manifests {
+		if desc.Digest == h {
+			return &desc, nil
```
xingao267: will we always only have one desc in this list and that value is the same as the one in `digest` file?

container/go/pkg/compat/index.go
```diff
@@ -0,0 +1,198 @@
...
+				Digest:    manifestDigest,
+			},
+		},
+	}
+
+	return &index, nil
+}
+
+// RawManifest returns the serialized bytes of manifest.json metadata.
+func (i *intermediateLayout) RawManifest() ([]byte, error) {
```
xingao267: This seems the manifest of the image. But layout's RawManifest is the one for index.json.

xingao267: Return an error when it's called.

container/go/pkg/compat/index.go
```diff
@@ -0,0 +1,198 @@
+/// Copyright 2015 The Bazel Authors. All rights reserved.
```
xingao267: I don't think we need index.go or have enough information to create a index object, as we are not really getting a ImageIndex from the legacy format. We are just getting an image.

xingao267: You are trying to use the same oci.Read function in pusher, which takes an path to an ImageIndex directory. So you implement the whole index.go here. But it doesn't apply to this case. We should not call the same oci.Read function in pusher. But instead, write a new Read function in this compat package which takes a path to the directory that contains the legacy image, and directly read a v1.Image out of it.

xingao267: And that read function is used in pusher.go when the format is legacy.

container/go/pkg/compat/index.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	// We are missing index.json in this intermediate format.
+	// Since index.json is represented in IndexManifest structure, we will populate this struct with parsed info.
+	index := v1.IndexManifest{
+		SchemaVersion: manifest.SchemaVersion,
+		Manifests: []v1.Descriptor{
+			v1.Descriptor{
+				MediaType: manifest.MediaType,
+				Size:      int64(len(i.rawManifest)),
```
xingao267: This size is wrong. You are using the size from the Image Manifest file.

container/go/pkg/compat/image.go
```diff
@@ -0,0 +1,147 @@
...
+
+	for i, desc := range manifest.Layers {
+		if h == desc.Digest {
+			switch desc.MediaType {
+			case types.OCILayer, types.DockerLayer:
+				return partial.CompressedToLayer(&compressedBlob{
+					path: li.path,
+					desc: desc,
+					// Passed in index to look for the actual gzipped layer file due to intermediate format naming convention.
+					index: i,
```
smukherj1: instead of setting the index, I would recommend storing the name of the file here, ie layerFilename(i). This way the implementation of the `Compressed` function on `compressedBlob` won't need the control statement.

container/go/pkg/compat/path.go
```diff
@@ -0,0 +1,58 @@
...
+	"path/filepath"
+)
+
+// Expected metadata files in legacy layout.
+const manifestFile string = "manifest.json"
+const configFile string = "config.json"
+const digestFile string = "digest"
+
+// Path returns a full directory of this path concatenated with other <elem> paths.
+func Path(path string, elem ...string) string {
```
smukherj1: Don't see a need for this function. Just doing filepath.Join wherever this is called should suffice as this function is not really doing anything special to handle docker images.

container/new_push.bzl
```diff
@@ -70,6 +70,13 @@ def _impl(ctx):
                 found = True
         if not found:
             fail("Did not find an index.json in the image attribute {} specified to {}".format(ctx.attr.image, ctx.label))
+    if ctx.attr.format == "legacy":
+        print(ctx.files.image)
```
smukherj1: Probably forgot to remove this after debugging.

container/go/pkg/compat/path.go
```diff
@@ -0,0 +1,58 @@
...
+package compat
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// Expected metadata files in legacy layout.
+const manifestFile string = "manifest.json"
```
smukherj1: Should be able to do
```
const (
    manifestFile = "manifest.json"
    configFile = "config.json"
    digestFile = "digest"
)
```

container/go/cmd/pusher/pusher.go
```diff
@@ -58,7 +59,7 @@ func main() {
 	}
 
 	// Ensure src is a directory, trim basename index.json if not.
-	if filepath.Base(*src) == "index.json" {
+	if filepath.Base(*src) == "index.json" || filepath.Base(*src) == "manifest.json" {
```
smukherj1: I think this logic could be improved. Right now it will only work if the path given is to a index.json or manifest.json. Instead, I would recommend using [this](https://golang.org/pkg/os/#FileMode.IsDir) function to check if the given path is a directory and if not, trim the path. Also generate logs here because it's changing what the input is being interpreted as.

container/go/pkg/compat/reader.go
```diff
@@ -0,0 +1,80 @@
...
+	"github.com/google/go-containerregistry/pkg/v1/validate"
+	"github.com/pkg/errors"
+)
+
+// Read returns a docker image referenced by the legacy intermediate layout at src. The image index should have been outputted by container_pull.
+// NOTE: this only reads index with a single image.
+func Read(src string) (v1.Image, error) {
+	_, err := isValidLegacylayout(src)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Invalid legacy layout at %s, requires manifest.json, config.json and digest files.", src)
```
smukherj1: error messages aren't standalone sentences. They should begin with a small letter and shouldn't end with a period.

container/go/pkg/compat/reader.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	digest, err := getManifestDigest(src)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Invalid sha256 hash in digest file at %s. Should have format sha256:{digest}", src)
```
smukherj1: begin error message with small letter. Also the error here shouldn't include details of how `getManifestDigest` expects it. That detail should be included in the error returned by `getManifestDigest`. The error here should only indicate that it was unable to get the manifest digest and the location of the image it was trying to get the manifest digest of.

container/go/pkg/compat/reader.go
```diff
@@ -0,0 +1,80 @@
...
+
+	// Constructs and validates a v1.Image object.
+	legacyImg := &legacyImage{
+		path:   src,
+		digest: digest,
+	}
+
+	img, err := partial.CompressedToImage(legacyImg)
+	if err != nil {
+		return nil, errors.Wrapf(err, "unable to load image with digest %s obtained from the manifest", digest)
```
smukherj1: mention the location the image was loaded from

container/go/pkg/compat/reader.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	if err := validate.Image(img); err != nil {
+		return nil, errors.Wrapf(err, "unable to load image with digest %s due to invalid legacy layout format", digest)
```
smukherj1: mention the location the image was loaded from

container/go/pkg/compat/reader.go
```diff
@@ -0,0 +1,80 @@
...
+
+	return img, nil
+}
+
+// Get the hash of the image to read at <path> from digest file.
+func getManifestDigest(path string) (v1.Hash, error) {
+	// We expect a file named digest that stores the manifest's hash formatted as sha256:{Hash} in this directory.
+	digest, err := ioutil.ReadFile(filepath.Join(path, digestFile))
+	if err != nil {
+		return v1.Hash{}, fmt.Errorf("Failed to locate SHA256 digest file for image manifest: %v", err)
```
smukherj1: `Failed to locate ...` -> `failed to locate ...`

container/go/pkg/compat/image.go
```diff
@@ -0,0 +1,155 @@
...
+var _ partial.CompressedImageCore = (*legacyImage)(nil)
+
+// MediaType of this image's manifest from manifest.json.
+func (li *legacyImage) MediaType() (types.MediaType, error) {
+	manifest, err := li.Manifest()
+	if err != nil {
+		return "", err
+	}
+
+	if !isExpectedMediaType(manifest.MediaType, types.OCIManifestSchema1, types.DockerManifestSchema2) {
```
smukherj1: Given that there's only two media types and this check is only done once, I would remove this function as it hinders readability. Simply doing
`if manifest.MediaType != types.OCIManifestSchema1 && manifest.MediaType != types.DockerManifestSchema2`

is much more readable.

container/go/pkg/compat/reader.go
```diff
@@ -0,0 +1,80 @@
...
+	}
+
+	return v1.NewHash(string(digest))
+}
+
+// isExpectedMediaType returns whether the given mediatype mt is allowed.
+func isExpectedMediaType(mt types.MediaType, expected ...types.MediaType) bool {
```
smukherj1: This function shouldn't be needed after addressing another comment.

container/go/cmd/pusher/pusher.go
```diff
@@ -33,7 +34,7 @@ import (
 var (
 	dst             = flag.String("dst", "", "The destination location including repo and digest/tag of the docker image to push. Supports fully-qualified tag or digest references.")
 	src             = flag.String("src", "", "Path to the directory which has the image index that will be pushed.")
```
smukherj1: Update help text here. Suggest `Path to the manifest.json when -format is legacy, path to the index.json when -format is oci or path to the image .tar file when -format is docker`.

container/go/cmd/pusher/pusher.go
```diff
@@ -57,26 +58,29 @@ func main() {
 		os.Setenv("DOCKER_CONFIG", *clientConfigDir)
 	}
 
-	// Ensure src is a directory, trim basename index.json if not.
-	if filepath.Base(*src) == "index.json" {
+	// Check if src is a tarball when pushing a docker image. Trim basename index.json or manifest.json if src is a directory, since we are pushing a OCI/legacy index.
+	if *format == "docker" && filepath.Ext(*src) != ".tar" {
```
smukherj1: Suggest:
```
var imgSrc string
if *format == "docker" && filepath.Ext(*src) != ".tar" {
    log.Fatalf("Invalid value for argument -src for -format=docker, got %q, want path to tarball file with extension .tar.", src)
}
if *format == "legacy" && filepath.Base(*src) != "manifest.json" {
    log.Fatalf("Invalid value for argument -src for -format=legacy, got %q, want path to manifest.json", *src)
}
if *format == "oci" && filepath.Base(*src) != "index.json" {
    log.Fatalf("Invalid value for argument -src for -format=oci, got %q, want path to index.json", *src)
}
if *format == "oci" || *format == "legacy" {
   imgSrc = filepath.Dir(*src)
}
if format == "docker" {
   imgSrc = *src
}
log.Printf("Determined image source path to be %q based on -format=%q, -src=%q.", imgSrc, *format, *src)
```

Pass this `imgSrc` argument to every function that needs it. Avoid repurposing global variables like `*src` which are meant to store the parsed value of flags.

Also, log.Fatalf messages need to begin with capital letters and end with the appropriate punctuation because they are full sentences. The beginning with lower case and not ending in punctuation only applies to messages given to an error object being returned up the stack like `errors.Wrapf` or `fmt.Errorf`, etc.

PR #970, opened 2019-07-09
Refactor package_managers from base-images-docker to this repo

nlopezgi:
This is just moving functionality from base-images-docker to here.
Notable changes from code currently in base-images-docker include:
- fix for --incompatible_no_support_tools_in_action_inputs=false
- modify test_complex_packages to run in GCB
- READMEs + better documentation added in attrs in bzl files

Note rules are not setup to use properly toolchains yet. Will do that in upcoming PR.
This PR is not migrating from base-images-docker:
- Security script check (will migrate in upcoming PR)
- bootstrap image rule (will NOT be migrated from base-images-docker to this repo)


docker/package_managers/apt_key.bzl
```diff
@@ -0,0 +1,159 @@
+#Copyright 2017 Google Inc. All rights reserved.
```
xingao267: This should be changed to the Bazel version of it.

repositories/deps.bzl
```diff
@@ -0,0 +1,34 @@
...
+# to exist.
+"""
+Provides functions to pull all external package dependencies of this
+repository.
+"""
+
+load(":go_repositories.bzl", "go_deps")
+load(":py_repositories.bzl", "py_deps")
+
+def container_deps():
```
xingao267: why name it container_deps?

tests/docker/package_managers/BUILD
```diff
@@ -0,0 +1,134 @@
+package(default_visibility = ["//visibility:public"])
```
xingao267: We also need these case https://github.com/GoogleContainerTools/base-images-docker/blob/master/BUILD#L10 which are located at the root of the repository.

repositories/deps.bzl
nlopezgi: renamed to deps

tests/docker/package_managers/BUILD
nlopezgi: added to testing/download_pkgs_at_root

PR #973, opened 2019-07-09
Implement createImageConfig binary using Go containerregistry

xwinxu:
Implementations of modifying an image's config.json file with extension to also changing the manifest.json. Also implementations of the utils libraries and helper functions for manipulating object properties.


container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+	// empty file sha256 hex
+	b         = sha256.Sum256([]byte(""))
+	emptyFile = hex.EncodeToString(b[:])
+)
+
+// default architecture type based on legacy create_image_config.py
+const processorArchitecture = "amd64"
+
+// unix epoch 0, representation in 32 bits
+const defaultTimestamp = "1970-01-01T00:00:00Z"
```
smukherj1: Is this variable used anywhere? If not remove it? If yes:
Suggest doing 
```
const (
   // ...
   defaultProcArch = ...
  // ...
  defaultTimestamp = time.Unix(0, 0).Format(time.RFC3339)
)
```

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+
+// arrayFlags is a type that can handle multiple flags of the same name.
```
smukherj1: remove `is a type` which is evident from the declaration syntax.

Also, suggesting saying `that can be used to store multiple flags ...` but this is just a nit. "handle" is too generic a word for this purpose imo.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+		err := errors.New("environment variable sought after does not exist in evnrionemnet map")
+		if err != nil {
+			errors.Wrap(err, "failed to create new error")
+		}
+		return "", err
+	}
+	return os.Expand(value, mapper), nil
+}
+
+func main() {
```
smukherj1: This function is too long. Suggest refactoring into functions.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+		entrypoint = nil
+	}
+	if *nullCmd == "true" {
+		command = nil
+	}
+
+	// read config file into struct.
+	configPath, err := os.Open(*baseConfig)
+	if err != nil {
+		log.Fatalf("Failed to read the parent image's config file: %v", err)
```
smukherj1: base image. Change all references to "parent" to "base" as "base" is the standard term used for Docker images.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+		}
+		manifestFile := v1.Manifest{}
+		if err = json.Unmarshal([]byte(manifestPath), &manifestFile); err != nil {
+			log.Fatalf("Failed to successfully read manifest file contents: %v", err)
+		}
+	}
+
+	var createTime string
+	if *creationTime == "" {
+		creationTime = nil
```
smukherj1: Avoid modifying parsed flag variables. This repurposes them and can be confusing when their values differ from the provided command line arguments later.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+		if err != nil {
+			log.Fatalf("Failed to read the parent image's manifest: %v", err)
+		}
+		manifestFile := v1.Manifest{}
+		if err = json.Unmarshal([]byte(manifestPath), &manifestFile); err != nil {
+			log.Fatalf("Failed to successfully read manifest file contents: %v", err)
+		}
+	}
+
+	var createTime string
```
smukherj1: Add comment here explaining what this variable is supposed to store.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+	if *creationTime == "" {
+		creationTime = nil
+	} else {
```
smukherj1: This else should become unnecessary after addressing the comment above.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+		var unixTime int64
+		if createTime, err = stamp(*creationTime); err != nil {
+			log.Fatalf("Unable to format creation time from BUILD_TIMESTAMP macros: %v", err)
+		}
+		// if creationTime is parsable as a floating point type, assume unix epoch timestamp.
+		// otherwise, assume RFC 3339 date/time format.
+		unixTime, err := strconv.ParseInt(createTime, 10, 64)
+		if err != nil {
+			log.Fatalf("Unable to parse a floating point type from flag creationTime: %v", err)
+		} else {
```
smukherj1: This else is unnecessary because the body of the if condition terminates the program.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+			if unixTime > 1.0e+11 {
```
smukherj1: Add comment explaining why this is necessary.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+				unixTime = unixTime / 1000.0
+			}
+			// construct a RFC 3339 date/time from Unix epoch.
```
smukherj1: Begin comment with capital letter

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+			t := time.Unix(unixTime, 0)
+			creationTime := t.Format(time.RFC3339)
```
smukherj1: This creates a new variable called `creationTime` that goes out of scope when this control block ends. Is this intended?

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+			}
+			configFile.Created = v1.Time{rfcCreationTime}
+		}
+	}
+
+	configFile.Author = "Bazel"
+	configFile.OS = *operatingSystem
+	configFile.Architecture = processorArchitecture
+
+	// output['config'] = defaults.get('config', {}) line 161
```
smukherj1: Remove

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+	configFile.Config = v1.Config{}
+	if len(entrypoint) > 0 {
+		configFile.Config.Entrypoint = entrypoint
+	}
+	if len(command) > 0 {
+		configFile.Config.Cmd = command
+	}
+	if *user != "" {
```
smukherj1: This if condition is unnecessary. The body can be brought outside without affecting the functionality.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+	} else {
+		var unixTime int64
+		if createTime, err = stamp(*creationTime); err != nil {
+			log.Fatalf("Unable to format creation time from BUILD_TIMESTAMP macros: %v", err)
+		}
+		// if creationTime is parsable as a floating point type, assume unix epoch timestamp.
+		// otherwise, assume RFC 3339 date/time format.
+		unixTime, err := strconv.ParseInt(createTime, 10, 64)
+		if err != nil {
+			log.Fatalf("Unable to parse a floating point type from flag creationTime: %v", err)
```
smukherj1: The function above is parsing as an int so I think this error message is misleading.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+	}
+	if *user != "" {
+		configFile.Config.User = *user
+	}
+
+	if len(env) != 0 {
+		environMap := keyValueToMap(env)
+		var resolvedValue string
+		for k, v := range environMap {
+			if resolvedValue, err = resolveVariables(v, environMap); err != nil {
```
smukherj1: Is this invocation correct? I think you might need to pass in the current os environment for a recursive substitution to work correctly. For example, for something like `PATH=$PATH:foo` I think this will just replace `$PATH` with `$PATH:foo` resulting in the value of `PATH` becoming `$PATH:foo:foo`.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+				log.Fatalf("Unable to resolve environment variables from path %s: %v", v, err)
```
smukherj1: Mention the key here as well. Also I think the word `path` in this message is a little confusing.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+	if len(command) > 0 {
+		configFile.Config.Cmd = command
+	}
+	if *user != "" {
+		configFile.Config.User = *user
+	}
+
+	if len(env) != 0 {
+		environMap := keyValueToMap(env)
+		var resolvedValue string
```
smukherj1: Don't declare the variable here. Just split the if condition inside the for loop instead. This variable isn't needed outside the for loop

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+func (i *arrayFlags) Set(value string) error {
+	*i = append(*i, value)
+	return nil
+}
+
+// extractValue returns the contents of a file pointed to by value if it starts with a '@'.
+func extractValue(value string) (string, error) {
+	if strings.HasPrefix(value, "@") {
+		f, err := os.Open(value[1:])
+		if err != nil {
```
smukherj1: This check is unnecessary. ioutil.Readall would return something similar if the file can't be opened. I suggest just removing this function as the only thing it's doing is reading the contents of a file and just calling this directly at the call site.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+		configFile.Config.Env = mapToKeyValue(environMap)
+	}
+
+	labels := keyValueToMap(labelsArray)
+	for label, value := range labels {
+		if strings.HasPrefix(value, "@") {
+			if labels[label], err = extractValue(value); err != nil {
+				log.Fatalf("Failed to extract the contents of labels file: %v", err)
+			}
+		} else if strings.Contains(value, "{") {
```
smukherj1: Add a `continue` above and change this `else if` to `if`. Go convention is to avoid if else ladder if possible.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+		}
+		configFile.Config.Labels = labelsMap
+	}
+
+	if len(ports) > 0 {
+		if len(configFile.Config.ExposedPorts) == 0 {
+			configFile.Config.ExposedPorts = make(map[string]struct{})
+		}
+		for _, port := range ports {
+			if strings.Contains(port, "/") {
```
smukherj1: This check is too lax. I would use a regex to verify it's of the format \d+\\/(tcp)|(udp)

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+
+func main() {
+	flag.Var(&labelsArray, "labels", "Augment the Label of the previous layer.")
+	flag.Var(&ports, "ports", "Augment the ExposedPorts of the previous layer.")
+	flag.Var(&volumes, "volumes", "Augment the Volumes of the previous layer.")
+	flag.Var(&entrypointPrefix, "entrypointPrefix", "Prefix the Entrypoint with the specified arguments.")
+	flag.Var(&env, "env", "Augment the Env of the previous layer.")
+	flag.Var(&command, "command", "Override the Cmd of the previous layer.")
+	flag.Var(&entrypoint, "entrypoint", "Override the Entrypoint of the previous layer.")
+	flag.Var(&layer, "layer", "Layer sha256 hashes that make up this image.")
```
smukherj1: Name this variable `layerDigests` to be explicit about what it stores. For the flag, I would suggest `layerDigest`. I would also mention in the help text that the order the layer digests are specified in is important.

Or is this path to files storing the digest of the layer? In that case, name the variable layerDigestFiles, ...

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+		for _, volume := range volumes {
+			configFile.Config.Volumes[volume] = struct{}{}
+		}
+	}
+
+	if *workdir != "" {
+		configFile.Config.WorkingDir = *workdir
+	}
+
+	layers := []string{}
```
smukherj1: layerDigests

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,413 @@
...
+	rawConfig, err := json.Marshal(configFile)
+	if err != nil {
+		log.Fatalf("Unable to read config struct into json object: %v", err)
+	}
+	err = ioutil.WriteFile(*outputConfig, rawConfig, os.ModePerm)
+	if err != nil {
+		log.Fatalf("Writing config to %s was unsuccessful: %v", *outputConfig, err)
+	}
+
+	log.Println("Successfully created Image Config.")
```
smukherj1: nit: Move these messages to right after the call to write out the config and manifest respectively and add the name of the generated file paths in the messages.

container/go/cmd/create_image_config/createImageConfig.go
xwinxu: yep will refactor into a package under compat

container/go/cmd/create_image_config/createImageConfig.go
xwinxu: yes, we don't need the string rep of creationTime, just the time.Time object. But it's refactored now as a result of the comment about modifying/resassigning flag variables.

container/go/cmd/create_image_config/createImageConfig.go
xwinxu: seems correct when I tested it against the old one. They output the same content for config.

container/go/cmd/create_image_config/createImageConfig.go
xwinxu: ah the error with this is that resolveVariables is assigned to resolvedValue, so it is not guaranteed to succeed and hence must be defined separately otherwise the linter complains.

container/go/cmd/create_image_config/createImageConfig.go
xwinxu: it's a single layer digest. This create image config is called in a loop in the bzl, so we don't need to worry about order in this binary as we assume it is handed correctly to us.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,357 @@
...
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+var (
+	emptyFile = emptySHA256Digest([]byte(""))
```
smukherj1: I meant get rid of the `emptyFile` variable and calling `emptySHA256Digest` wherever `emptyFile` is currently used instead.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,357 @@
...
+
+const (
+	// defaultProcArch is the default architecture type based on legacy create_image_config.py.
+	defaultProcArch = "amd64"
+	// defaultTimeStamp is the unix epoch 0 time representation in 32 bits.
+	defaultTimestamp = "1970-01-01T00:00:00Z"
+)
+
+// emptySHA256Digest returns the sha256 sum of an empty digest.
+func emptySHA256Digest(emptyByte []byte) (empty string) {
```
smukherj1: This function shouldn't take an argument. Right now, this function caculates the digest of any given blob of bytes so it would be named `SHA256Digest`. So:
1. Change the function to not accept an argument.
2. Keep the name.
3. Change the impl to always return the digest of an empty byte slice.
4. Call this function instead of `emptyFile`.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,139 @@
...
+
+func main() {
+	flag.Var(&labelsArray, "labels", "Augment the Label of the previous layer.")
+	flag.Var(&ports, "ports", "Augment the ExposedPorts of the previous layer.")
+	flag.Var(&volumes, "volumes", "Augment the Volumes of the previous layer.")
+	flag.Var(&entrypointPrefix, "entrypointPrefix", "Prefix the Entrypoint with the specified arguments.")
+	flag.Var(&env, "env", "Augment the Env of the previous layer.")
+	flag.Var(&command, "command", "Override the Cmd of the previous layer.")
+	flag.Var(&entrypoint, "entrypoint", "Override the Entrypoint of the previous layer.")
+	flag.Var(&layerDigest, "layer", "Layer sha256 hashes that make up this image. The order that these layers are specified matters.")
```
smukherj1: I think these are files that contain the layerDigest (according to how it's being used in OverrideConfig), so I suggest naming the variable and the flag `layerDigestFile` instead and changing the description to say it's the file that contains the sha256 digest of the layer.

Also, I would explore the possibility of not having an `@` at the beginning of these names to avoid having to do what `extractValue` initially does to remove the `@` prefix.

container/image.bzl
```diff
@@ -123,74 +123,73 @@ def _image_config(
...
-        args += ["--user=" + ctx.attr.user]
+        args += ["-user=" + ctx.attr.user]
     if workdir:
-        args += ["--workdir=" + workdir]
+        args += ["-workdir=" + workdir]
 
     inputs = layer_names
     for layer_name in layer_names:
-        args += ["--layer=@" + layer_name.path]
+        args += ["-layer=@" + layer_name.path]
```
smukherj1: Try removing the `@` here so that the `extractValue` function becomes unnecessary

container/image.bzl
```diff
@@ -123,74 +123,73 @@ def _image_config(
...
+        args += ["-creationTime=%s" % creation_time]
     elif ctx.attr.stamp:
         # If stamping is enabled, and the creation_time is not manually defined,
         # default to '{BUILD_TIMESTAMP}'.
-        args += ["--creation_time={BUILD_TIMESTAMP}"]
+        args += ["-creationTime={BUILD_TIMESTAMP}"]
 
-    args += ["--labels=%s" % "=".join([key, value]) for key, value in labels.items()]
-    args += ["--env=%s" % "=".join([
+    args += ["-labels=%s" % "=".join([key, value]) for key, value in labels.items()]
```
smukherj1: Try removing the `@` here so that the `extractValue` function becomes unnecessary.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,357 @@
...
+	b := sha256.Sum256(emptyByte)
+	empty = hex.EncodeToString(b[:])
+	return
+}
+
+// extractValue returns the contents of a file pointed to by value if it starts with a '@'.
+func extractValue(value string) (string, error) {
```
smukherj1: See other comments about not passing in the `@` in the layer & label arguments. This function should become unnecessary then. You should just be able to do ioutil.ReadFile instead.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,139 @@
...
+
+var (
+	baseConfig         = flag.String("baseConfig", "", "The base image config.")
+	baseManifest       = flag.String("baseManifest", "", "The base image manifest.")
+	outputConfig       = flag.String("outputConfig", "", "The output image config file to generate.")
+	outputManifest     = flag.String("outputManifest", "", "The output manifest file to generate.")
+	creationTimeString = flag.String("creationTime", "", "The creation timestamp. Acceptable formats: Integer or floating point seconds since Unix Epoch, RFC, 3339 date/time.")
+	user               = flag.String("user", "", "The username to run the commands under.")
+	workdir            = flag.String("workdir", "", "Set the working directory of the layer.")
+	nullEntryPoint     = flag.String("nullEntryPoint", "False", "If True, Entrypoint will be set to null.")
```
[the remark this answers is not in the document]
xwinxu: don't think I can,  since in python it's True but in go it's true?

smukherj1:
beep boop I'm a bot

PR #974, opened 2019-07-09
Refactor new_container_pull to generate fully backward compatible format

xwinxu:
Added manifest.json symlink to fully parallel the legacy format.

### DIFF
--- a/container/go/pkg/compat/ociToLegacy.go
+++ b/container/go/pkg/compat/ociToLegacy.go
@@ -104,7 +104,7 @@ func LegacyFromOCIImage(img v1.Image, srcDir, dstDir string) error {
 		return errors.Wrap(err, "unable to get manifest from image index")
 	}
 	if len(manifest.Manifests) != 1 {
-		log.Fatalf("Image index read from %s had unexpected number of manifests: got %d want 1.", srcDir, len(manifest.Manifests))
+		log.Fatalf("Image index read from %s had unexpected number of manifests: got %d, want 1", srcDir, len(manifest.Manifests))
 	}
 	manifestHex := manifest.Manifests[0].Digest.Hex
 	dstLink = path.Join(dstDir, legacyManifestFile)
--- a/container/go/pkg/compat/ociToLegacy.go
+++ b/container/go/pkg/compat/ociToLegacy.go
@@ -24,6 +24,7 @@ import (
 	"strconv"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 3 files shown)


container/go/pkg/compat/ociToLegacy.go
```diff
@@ -90,5 +92,22 @@ func LegacyFromOCIImage(img v1.Image, srcDir, dstDir string) error {
...
+	}
+	manifest, err := imgIndex.IndexManifest()
+	if err != nil {
+		return errors.Wrap(err, "unable to get manifest from image index")
+	}
+	manifestHex := manifest.Manifests[0].Digest.Hex
+	dstLink = path.Join(dstDir, manifestExt)
+	manifestPath := path.Join(targetDir, manifestHex)
+	if err = generateSymlinks(manifestPath, dstLink); err != nil {
+		return errors.Wrapf(err, "failed to generate manifest.json symlink")
```
smukherj1: Use the `manifestExt` or the renamed variable in the error.

container/go/pkg/compat/ociToLegacy.go
```diff
@@ -24,12 +24,14 @@ import (
 	"strconv"
 
 	v1 "github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/layout"
 	"github.com/pkg/errors"
 )
 
 // Extension for layers and config files that are made symlinks
 const targzExt = ".tar.gz"
 const configExt = "config.json"
```
smukherj1: For config & manifest, these are not really extensions but the name of the files themselves. I would remove the "Ext" from the names of the variables. Suggestion `legacyConfigFile` & `legacyManifestFile` but feel free to choose something else.

container/go/pkg/compat/ociToLegacy.go
```diff
@@ -80,15 +83,35 @@ func LegacyFromOCIImage(img v1.Image, srcDir, dstDir string) error {
...
+	// named after its sha256 digest under blobs/sha256 to manifest.json
+	imgIndex, err := layout.ImageIndexFromPath(srcDir)
+	if err != nil {
+		return errors.Wrapf(err, "unable to open image as index from %s", srcDir)
+	}
+	manifest, err := imgIndex.IndexManifest()
+	if err != nil {
+		return errors.Wrap(err, "unable to get manifest from image index")
+	}
+	if len(manifest.Manifests) > 1 {
```
smukherj1: we want exactly 1 manifest. so len(manifest.Manifests) != 1

0 manifest (which is probably something layout.ImageIndexFromPath will never allow to happen but we want to check for it regardless) will cause the line `manifestHex := manifest.Manifests[0].Digest.Hex` to panic.

xwinxu: fixed.

container/go/pkg/compat/ociToLegacy.go
```diff
@@ -80,15 +83,35 @@ func LegacyFromOCIImage(img v1.Image, srcDir, dstDir string) error {
...
+	imgIndex, err := layout.ImageIndexFromPath(srcDir)
+	if err != nil {
+		return errors.Wrapf(err, "unable to open image as index from %s", srcDir)
+	}
+	manifest, err := imgIndex.IndexManifest()
+	if err != nil {
+		return errors.Wrap(err, "unable to get manifest from image index")
+	}
+	if len(manifest.Manifests) != 1 {
+		log.Fatal("error: image contains more than one manifest")
```
smukherj1: Update message `log.Fatalf("Image index read from %s had unexpected number of manifests: got %d, want 1.", srcDir, len(manifest.Manifests))`

PR #979, opened 2019-07-12
Use Python2 in image_util.sh

Kernald:
While `config_stripper` is wrapped as a Bazel binary and is ran with the proper toolchain, image_util.sh isn't, and is calling an inline Python script to format a command. The script isn't Python 3 compatible, so should be calling the `python2` binary.

### DIFF
--- a/docker/util/image_util.sh
+++ b/docker/util/image_util.sh
@@ -8,7 +8,7 @@ reset_cmd() {
     local old_cmd
     # docker inspect input cannot be piped into docker commit directly, we need to JSON format it.
     old_cmd=$(docker inspect -f "{{range .Config.Cmd}}{{.}} {{end}}" "${original_image_name}")
-    fmt_cmd=$(echo "$old_cmd" | python -c "import sys,json; print json.dumps(sys.stdin.read().strip().split())")
+    fmt_cmd=$(echo "$old_cmd" | python2 -c "import sys,json; print json.dumps(sys.stdin.read().strip().split())")
 
     # If CMD wasn't set, set it to a sane default.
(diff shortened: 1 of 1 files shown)


Kernald:
While the Travis issue seems to be a downloading issue unrelated to this change, the failure in `//tests/container:new_distroless_base_test_index_json` seems legit. I'm not exactly sure what this checksum represents though, and if it should just be updated or actually results from a change in the way a test container was generated.

nlopezgi:
Hi @Kernald ,
Thanks for creating this PR.
We just migrated these rules from base-images-docker to rules_docker. There are some rough edges, like this one, regarding how these rules access tools. In particular, the way this script is accessing python is not really compatible with how python toolchains should be used in Bazel. That is, the real fix here is probably to create a `py_binary` target to wrap around the functionality this is executing in the shell script and to pass that binary to this shell script. If you are willing to spend a bit more time to fix this I can help guide you. Otherwise, lets create an issue for now and I'll try to get to it soon as part of other fixes/improvements that I have planned for these rules.

Kernald:
Hi Nicolas,
Thanks for the response. It makes perfect sense. I'd be happy to take a deeper look at this, but not being familiar with Python at all and still learning how to use Bazel for anything else than _using_ existing rules, there's one thing I'm not sure of: what would be the proper way to execute such a `py_binary` target from a script that is itself ran by Bazel?

nlopezgi:
Hi @Kernald,
yeah, if you've never written Starlark rules before, then probably explaining in enough detail what needs to happen here might be more effort than preparing the PR. If you could please just create an issue to fix this and I'll get to it as part of other upcoming improvements. Thanks again!

Kernald:
Makes perfect sense, thanks! I just opened #985.

PR #984, opened 2019-07-12
Enable container_repro_test for toolchain_container installing deb packages

alex1545:
In order to build a `toolchain_container` target that installs deb packages inside a container (which is what container_repro_test does) we need:
1. To mount a temp directory from the host into the container.
2. Within the container, run the `bazel build` command with `--output_base` pointing at the mounted dir.

This allows mounting generated files into the other container that needs to be run (to install packages) inside the container.

### DIFF
--- a/contrib/repro_test.bzl
+++ b/contrib/repro_test.bzl
@@ -135,6 +135,7 @@ def _impl(ctx):
     # Commands to build the image targets and copy the files required for image
     # comparison to a known location in the container (being 'extract_path').
     cd_cmd = ["cd \$(cat /%s)" % proj_root.basename]
+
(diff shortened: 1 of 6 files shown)


contrib/repro_test.bzl
```diff
@@ -129,12 +129,24 @@ def _impl(ctx):
     extract_path = "/img_outs"
     img_tar = img_name + ".tar"
     img_digest = img_name + ".digest"
+    host_outs_path1 = "/tmp/" + name + "/bazel_out1"
```
nlopezgi: As mentioned offline, lets do our best to not have to mount anything under /tmp directory

alex1545: #1003 solves this.

PR #1001, opened 2019-07-16
use docker toolchains in docker/package_manager and docker/util rules

nlopezgi:


nlopezgi:
Will add todo in following PR that is also touching these rules. Thanks!

PR #1003, opened 2019-07-16
Test container repro test in gcb

nlopezgi:
* Fix also mounting of local directory onto containers that are running bazel commands. This enables running sibling container as part of process of building the container.

* Pending: discuss why container-diff is showing false positives. e.g., if I dont remove files in /var/cache/fontconfig/ the test fails, and shows that even files that were cleaned up differed - specifically etc/ssl/certs/java/cacerts. I double checked and the files did not exist in the container, but might exist in some form in some layer or manifest?


contrib/repro_test.bzl
```diff
@@ -130,17 +130,46 @@ def _impl(ctx):
...
+    # in the containers.
+    # This is necessary as the Bazel executions inside each container can
+    # in turn bring up a docker container and attempt to mount contents
+    # from their respective output_bases.
+    bazel_out1 = ctx.actions.declare_directory("bazel_out1")
+    bazel_out2 = ctx.actions.declare_directory("bazel_out2")
+
+    # dummy action to declare directories as outputs.
+    ctx.actions.run_shell(
+        command = "exit 0",
```
alex1545: Why not do
`command = "mkdir -p %s %s" % (bazel_out1.path, bazel_out2.path)`
instead?

contrib/repro_test.bzl
```diff
@@ -130,17 +130,46 @@ def _impl(ctx):
...
+    # We need to use absolute paths to the directory being mounted
+    # in order for containers spawned from inside the build to be able
+    # to access the output_base.
+    host_outs_path1 = "$(pwd)/" + bazel_out1.path
+    host_outs_path2 = "$(pwd)/" + bazel_out2.path
+
+    # Commands to build the image targets and copy files.
+    commands_image1 = cd_cmd + [
+        ("bazel --output_base=%s build " % host_outs_path1) + " ".join(build_targets),
+    ] + cp_cmds + ["rm -rdf %s" % host_outs_path1]
```
alex1545: Interesting that removing the mounted dir from within the container just works like this.
When I tried this, I had to first `umount` the mounted dir and for that this container had to also run with the `--privileged` flag. Did you encounter anything similar?

alex1545: Also, doesn't `rm -r` already does what `rm -d` is doing?

contrib/repro_test.bzl
nlopezgi: because the directories get created automatically by Bazel (seems kind of a problem with documentation), if I try to recreate them with mkdir I get an error.

contrib/repro_test.bzl
nlopezgi: removed unnecessary -d
This ends up removing the contents of the directory, not the directory, and does not complain.

PR #1005, opened 2019-07-17
Use distroless/static base for go containers when pure is "on"

### LINKED ISSUE #649: Build smaller Go images when pure is true
The default go base image should be `gcr.io/distroless/static` instead of `gcr.io/distroless/base` when pure is True.
It would be nice if rules_docker supported that by default.

See https://github.com/GoogleContainerTools/distroless/tree/master/base (which you probably already know very well).

**UPDATE** rereading this, it sounds incredibly condescending. I definitely didn't mean it that way. I'm very grateful for the work you do and I'd just like this to avoid setting different base images myself.

djfinlay:
Fixes #649

### DIFF
--- a/go/image.bzl
+++ b/go/image.bzl
@@ -34,7 +34,8 @@ load(
 )
 
 # Load the resolved digests.
-load(":go.bzl", "DIGESTS")
+load(":go.bzl", BASE_DIGESTS = "DIGESTS")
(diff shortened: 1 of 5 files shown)


djfinlay:
I could test for the existence of glibc and libssl but all the filenames have versions in them, which could make the tests flaky. Checking for /usr/bin/openssl is enough to differentiate.

PR #1011, opened 2019-07-19
Add e2e tests to check new_pusher compatibility

xiaohegong:
Added e2e tests for checking compatibility of new pusher.


tests/container/BUILD
```diff
@@ -320,6 +320,21 @@ container_push(
...
+    name = "alpine_with_custom_attr",
+    base = "@new_alpine_linux_amd64//image",
+    cmd = ["Hello World!"],
+    entrypoint = ["echo"],
+    files = [
+        "//testdata:bar",
+        "//testdata:foo",
+    ],
+    layers = [
+        "//testdata:env_layer",
```
smukherj1: Can you add another layer here that has files? I want to ensure the pusher works when there's more than one layer with files to upload. Also, for this case we want to run a `container_pull` followed by `container_test` to ensure files from both layers were uploaded to the registry correctly.

tests/container/BUILD
```diff
@@ -370,20 +385,70 @@ new_container_push(
...
     format = "legacy",
+    image = ":alpine_with_custom_attr",
+    registry = "localhost:5000",
+    repository = "docker/test",
+    tag = "test",
+    tags = ["manual"],
+)
+
+new_container_push(
+    name = "new_push_test_legacy_auth",
```
xingao267: Is the image also from container_img? If so, make the naming consistent.

tests/container/BUILD
```diff
@@ -370,20 +385,70 @@ new_container_push(
...
+    format = "legacy",
     image = ":set_cmd_and_entrypoint",
     registry = "localhost:5000",
     repository = "docker/test",
     tag = "test",
     tags = ["manual"],
 )
 
+new_container_push(
+    name = "new_push_test_pulled_legacy",
```
xingao267: new_push_test_legacy_from_old_puller

tests/container/BUILD
```diff
@@ -370,20 +385,70 @@ new_container_push(
     tags = ["manual"],
 )
 
+new_container_push(
+    name = "new_push_test_pulled_oci",
```
xingao267: new_push_test_oci_from_new_puller

tests/container/BUILD
xingao267: like new_push_test_legacy_from_container_img_with_auth

tests/container/BUILD
```diff
@@ -370,20 +385,70 @@ new_container_push(
...
+    format = "legacy",
+    image = "@new_alpine_linux_ppc64le//image",
+    registry = "localhost:5000",
+    repository = "docker/test",
+    tag = "test",
+    tags = ["manual"],
+)
+
+new_container_push(
+    name = "new_push_test_old_puller_compat",
```
xingao267: new_push_test_legacy_from_old_puller

PR #1017, opened 2019-07-23
Add a test rule to syntax check file update & dependency update spec

smukherj1:
This replaces #1010


contrib/automatic_container_release/run_checker.sh.tpl
```diff
@@ -0,0 +1,59 @@
...
+        echo "$( cd ${BASH_SOURCE[0]}.runfiles && pwd )"
+    else
+        # The current script is within some other script's runfiles.
+        mydir="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
+        echo $mydir | sed -e 's|\(.*\.runfiles\)/.*|\1|'
+    fi
+}
+
+RUNFILES="${PYTHON_RUNFILES:-$(guess_runfiles)}"
+
```
nlopezgi: Do a check for the docker_tool_path here as in other templates: https://github.com/bazelbuild/rules_docker/blob/master/docker/util/extract.sh.tpl#L6

smukherj1: Done

PR #1023, opened 2019-07-25
Modify container_push and container_image to use new digester

xiaohegong:
Updated `container_push`, the old pusher, and `container_image` to use newly implemented digester in go. 
- Added generating legacy format in `push.bzl` to understand the legacy intermediate format used in the old pusher
- Added generating legacy format in `image.bzl` to understand the legacy intermediate format used in the `container_image`


container/go/cmd/digester/digester.go
```diff
@@ -16,23 +16,28 @@
...
 
 var (
-	dst    = flag.String("dst", "", "The destination location of the digest file to write to.")
-	src    = flag.String("src", "", "Path to the config.json when -format is legacy, path to the index.json when -format is oci or path to the image .tar file when -format is docker.")
-	format = flag.String("format", "", "The format of the source image, (oci, legacy, or docker). The docker format should be a tarball of the image as generated by docker save.")
-	layers utils.ArrayStringFlags
+	dst     = flag.String("dst", "", "The destination location of the digest file to write to.")
+	src     = flag.String("src", "", "Path to the config.json when -format is legacy, path to the index.json when -format is oci or path to the image .tar file when -format is docker.")
+	format  = flag.String("format", "", "The format of the source image, (oci, legacy, or docker). The docker format should be a tarball of the image as generated by docker save.")
+	tarball = flag.String("tarball", "", "The path to a legacy base image in tarball form, optional.")
```
smukherj1: Put `Optional:` in the beginning of the help text. Also mention this is only relevant when format is legacy. Probably name the flag `legacyBaseImage` to make this more explicit in the flag name.

container/go/cmd/digester/digester.go
```diff
@@ -88,32 +93,92 @@ func main() {
...
-	if err != nil {
-		log.Fatalf("Error getting image digest: %v", err)
+	configPath := ""
+	if *format == "legacy"{
+		configPath = *src
 	}
-
-	err = writeDigest(digest, *dst)
+	err = writeDigest(img, *dst, configPath)
 	if err != nil {
```
smukherj1: `if err := writeDigest(...); err != nil {`

container/go/cmd/digester/digester.go
```diff
@@ -88,32 +93,92 @@ func main() {
...
 		log.Fatalf("Error outputting digest file to %s: %v", *dst, err)
 	}
 
 	log.Printf("Successfully generated image digest file at %s", *dst)
 }
 
-// writeDigest outputs digest to a "digest file" at dst. Digest is the image manifest's sha256 hash.
-func writeDigest(digest v1.Hash, dst string) error {
+// writeDigest outputs image's digest to a "digest file" at dst. Digest is the image manifest's sha256 hash.
```
smukherj1: Can be simplified to `writeDigest writes the sha256 digest of the manifest of the given image to the file given by dst. The image config is replaced with the config at the given path before generating the digest.`

container/go/cmd/digester/digester.go
```diff
@@ -88,32 +93,92 @@ func main() {
...
+func writeDigest(image v1.Image, dst, configPath string) error {
+	m, err := image.Manifest()
+	if err != nil {
+		return errors.Wrap(err, "error getting image manifest")
+	}
+
+	if configPath != "" {
+		rawConfig, err := ioutil.ReadFile(configPath)
+		if err != nil {
+			return errors.Wrap(err, "Error getting image raw config")
```
smukherj1: Error -> error. Also mention the config file name in the error.

container/go/cmd/digester/digester.go
```diff
@@ -88,32 +93,92 @@ func main() {
...
+
+	digest, _, err := v1.SHA256(bytes.NewReader((*out).Bytes()))
+	if err != nil {
+		return errors.Wrap(err, "unable to compute sha256 sum for digest")
+	}
+
 	rawDigest := []byte(digest.Algorithm + ":" + digest.Hex)
 
-	err := ioutil.WriteFile(dst, rawDigest, os.ModePerm)
+	err = ioutil.WriteFile(dst, rawDigest, os.ModePerm)
```
smukherj1: combine this with the if

container/go/cmd/digester/digester.go
```diff
@@ -88,32 +93,92 @@ func main() {
...
 	if err != nil {
 		return errors.Wrapf(err, "unable to write digest file to %s", dst)
 	}
 
 	return nil
 }
+
+func formatStruct(s interface{}) (*bytes.Buffer, error) {
```
smukherj1: Add function comment. Also, this function doesn't need to be this generic as it's only being used to format the manifest. I would suggest naming it `manifestToJSON` and accepting a `v1.Manifest` as the input parameter.

container/go/cmd/pusher/pusher.go
```diff
@@ -35,6 +35,7 @@ var (
 	src             = flag.String("src", "", "Path to the config.json when -format is legacy, path to the index.json when -format is oci or path to the image .tar file when -format is docker.")
 	format          = flag.String("format", "", "The format of the source image, (oci, legacy, or docker). The docker format should be a tarball of the image as generated by docker save.")
 	clientConfigDir = flag.String("client-config-dir", "", "The path to the directory where the client configuration files are located. Overiddes the value from DOCKER_CONFIG.")
+	tarball         = flag.String("tarball", "", "The path to a legacy base image in tarball form, optional.")
```
smukherj1: Same comment here as the digester.

container/go/pkg/compat/reader.go
```diff
@@ -57,6 +60,31 @@ func Read(src string) (v1.Image, error) {
 	return img, nil
 }
 
+// ReadWithBaseTarball returns a Image object read from src with tarball as base and layers appended from layersPath.
+func ReadWithBaseTarball(src, tarballPath string, layersPath []string) (v1.Image, error) {
```
smukherj1: It appears `src` is not used here? If so, remove the parameter.

PR #1024, opened 2019-07-26
Refactor createImageConfig.go and rectify image tests

xwinxu:
This PR contains all changes that correlate to the createImageConfig function of container_image:

* Implement `createImageConfig.go` to write out modified config file and empty/base manifest file, maintaining parity with the current behaviour
* Implement an `overrideConfig.go` binary to mutate the base config file to match the updates supplied in the flags. This also contains the `Stamp` utility function that the digester is dependent on.
* Modify the sha256 hashes in `image_test.py` to contain the updated hashes from using `go-containerregistry` backend
* Modify `image.bzl` to accept golang flags

### DIFF
--- a/container/go/cmd/create_image_config/createImageConfig.go
+++ b/container/go/cmd/create_image_config/createImageConfig.go
@@ -1,0 +1,163 @@
+// Copyright 2016 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+////////////////////////////////////////////////
+// This package manipulates v2.2 image configuration metadata.
+// It writes out both a config file and a manifest for the v2.2 image.
+
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+
+	"github.com/bazelbuild/rules_docker/container/go/pkg/compat"
+	v1 "github.com/google/go-containerregistry/pkg/v1"
+)
+
+var (
+	baseConfig         = flag.String("baseConfig", "", "The base image config.")
+	baseManifest       = flag.String("baseManifest", "", "The base image manifest.")
+	outputConfig       = flag.String("outputConfig", "", "The output image config file to generate.")
+	outputManifest     = flag.String("outputManifest", "", "The output manifest file to generate.")
+	creationTimeString = flag.String("creationTime", "", "The creation timestamp. Acceptable formats: Integer or floating point seconds since Unix Epoch, RFC, 3339 date/time.")
+	user               = flag.String("user", "", "The username to run the commands under.")
+	workdir            = flag.String("workdir", "", "Set the working directory of the layer.")
+	nullEntryPoint     = flag.String("nullEntryPoint", "False", "If True, Entrypoint will be set to null.")
+	nullCmd            = flag.String("nullCmd", "False", "If True, Cmd will be set to null.")
+	operatingSystem    = flag.String("operatingSystem", "linux", "Operating system to create docker image for, eg. linux.")
+	labelsArray        arrayFlags
+	ports              arrayFlags
+	volumes            arrayFlags
+	entrypointPrefix   arrayFlags
+	env                arrayFlags
+	command            arrayFlags
+	entrypoint         arrayFlags
+	layerDigestFile    arrayFlags
+	stampInfoFile      arrayFlags
+)
+
+const (
+	// defaultProcArch is the default architecture type based on legacy create_image_config.py.
+	defaultProcArch = "amd64"
+	// defaultTimeStamp is the unix epoch 0 time representation in 32 bits.
+	defaultTimestamp = "1970-01-01T00:00:00Z"
+)
+
+// arrayFlags can be used to store multiple flags the same name.
+// the resulting data parsed in will be an array data type.
+type arrayFlags []string
+
+func (i *arrayFlags) String() string {
+	return fmt.Sprintf("%s", strings.Join(*i, ", "))
+}
+
+// Get returns an empty interface that may be type-asserted to the underlying
+// value of type bool, string, etc.
+func (i *arrayFlags) Get() interface{} {
+	return ""
+}
+
+func (i *arrayFlags) Set(value string) error {
+	*i = append(*i, value)
+	return nil
+}
+
+func main() {
+	log.Println("Args before:", os.Args)
+	flag.Var(&labelsArray, "labels", "Augment the Label of the previous layer.")
+	flag.Var(&ports, "ports", "Augment the ExposedPorts of the previous layer.")
+	flag.Var(&volumes, "volumes", "Augment the Volumes of the previous layer.")
+	flag.Var(&entrypointPrefix, "entrypointPrefix", "Prefix the Entrypoint with the specified arguments.")
+	flag.Var(&env, "env", "Augment the Env of the previous layer.")
+	flag.Var(&command, "command", "Override the Cmd of the previous layer.")
+	flag.Var(&entrypoint, "entrypoint", "Override the Entrypoint of the previous layer.")
+	flag.Var(&layerDigestFile, "layerDigestFile", "Layer sha256 hashes that make up this image. The order that these layers are specified matters.")
+	flag.Var(&stampInfoFile, "stampInfoFile", "A list of files from which to read substitutions to make in the provided fields.")
+
+	flag.Parse()
+
+	log.Println("Running the Image Config manipulator...")
+
+	if *outputConfig == "" {
+		log.Fatalln("Required option -outputConfig was not specified.")
+	}
+	if *nullEntryPoint == "True" {
+		entrypoint = nil
+	}
+	if *nullCmd == "True" {
+		command = nil
+	}
+
+	// read config file into struct.
+	configFile := &v1.ConfigFile{}
+	if *baseConfig != "" {
+		// configPath, err := os.Open(*baseConfig)
+		configPath, err := ioutil.ReadFile(*baseConfig)
+
+		if err != nil {
+			log.Fatalf("Failed to read the base image's config file: %v", err)
+		}
+
+		configFile, err = v1.ParseConfigFile(bytes.NewReader(configPath))
+		if err != nil {
+			log.Fatalf("Failed to successfully parse config file json contents: %v", err)
+		}
+		// log.Print("Configfile \n")
+		// log.Printf("%+v", *configFile)
+		// log.Print("end")
+	} else {
+		// write out an empty config file.
+		log.Println("baseConfig is empty!!!")
+	}
+	// write out the updated config after overriding config content.
+	err := compat.OverrideContent(configFile, *outputConfig, *creationTimeString, *user, *workdir, *nullEntryPoint, *nullCmd, *operatingSystem, labelsArray[:], ports[:], volumes[:], entrypointPrefix[:], env[:], command[:], entrypoint[:], layerDigestFile[:], stampInfoFile[:])
+	if err != nil {
+		log.Fatalf("Failed to override values in old image config and write to dst %s: %v", err, *outputConfig)
+	}
+
+	log.Printf("Successfully created Image Config at %s.\n", *outputConfig)
+
+	// if *baseManifest != "" {
+	// read manifest file into struct if provided.
+	manifestFile := v1.Manifest{}
+	if *baseManifest != "" {
+		manifestPath, err := ioutil.ReadFile(*baseManifest)
+		if err != nil {
+			log.Fatalf("Failed to read the base image's manifest: %v", err)
+		}
+		if err = json.Unmarshal([]byte(manifestPath), &manifestFile); err != nil {
+			log.Fatalf("Failed to successfully read manifest file contents: %v", err)
+		}
+	}
+
+	type Q struct{}
+
+	// TODO(xwinxu): write out the updated manifest after updating it from compat pkg.
+	rawManifest, err := json.Marshal(Q{})
+	if err != nil {
+		log.Fatalf("Unable to read config struct into json object: %v", err)
+	}
+	err = ioutil.WriteFile(*outputManifest, rawManifest, os.ModePerm)
+	if err != nil {
+		log.Fatalf("Writing config to %s was unsuccessful: %v", *outputManifest, err)
+	}
+
+	log.Println("Successfully created Image Manifest at %s.\n", *outputManifest)
+}
--- a/container/go/cmd/create_image_config/createImageConfig.go
+++ b/container/go/cmd/create_image_config/createImageConfig.go
@@ -161,6 +161,7 @@ func main() {
 		// if err != nil {
 		// 	log.Fatalf("Unable to read config struct into json object: %v", err)
 		// }
+		// write manifest instead of rawManifest
 		err = ioutil.WriteFile(*outputManifest, manifest, os.ModePerm)
 		if err != nil {
 			log.Fatalf("Writing config to %s was unsuccessful: %v", *outputManifest, err)
--- a/container/go/cmd/create_image_config/createImageConfig.go
+++ b/container/go/cmd/create_image_config/createImageConfig.go
@@ -19,7 +19,6 @@ package main
 
 import (
 	"bytes"
-	"encoding/json"
 	"flag"
 	"io/ioutil"
 	"log"
@@ -38,8 +37,8 @@ var (
 	creationTimeString = flag.String("creationTime", "", "The creation timestamp. Acceptable formats: Integer or floating point seconds since Unix Epoch, RFC 3339 date/time.")
 	user               = flag.String("user", "", "The username to run the commands under.")
 	workdir            = flag.String("workdir", "", "Set the working directory of the layer.")
-	nullEntryPoint     = flag.String("nullEntryPoint", "False", "If True, Entrypoint will be set to null.")
-	nullCmd            = flag.String("nullCmd", "False", "If True, Cmd will be set to null.")
+	nullEntryPoint     = flag.Bool("nullEntryPoint", false, "If true, Entrypoint will be set to null.")
+	nullCmd            = flag.Bool("nullCmd", false, "If true, Cmd will be set to null.")
 	operatingSystem    = flag.String("operatingSystem", "linux", "Operating system to create docker image for, eg. linux.")
 	labelsArray        utils.ArrayStringFlags
 	ports              utils.ArrayStringFlags
@@ -60,7 +59,6 @@ const (
 )
 
 func main() {
-	log.Println("Args before:", os.Args)
 	flag.Var(&labelsArray, "labels", "Augment the Label of the previous layer.")
 	flag.Var(&ports, "ports", "Augment the ExposedPorts of the previous layer.")
 	flag.Var(&volumes, "volumes", "Augment the Volumes of the previous layer.")
@@ -95,7 +93,7 @@ func main() {
 		log.Println("baseConfig is empty!")
 	}
 
-	overrideConfig := compat.OverrideConfigOpts{
+	opts := compat.OverrideConfigOpts{
 		ConfigFile:         configFile,
 		OutputConfig:       *outputConfig,
 		CreationTimeString: *creationTimeString,
@@ -104,8 +102,8 @@ func main() {
 		NullEntryPoint:     *nullEntryPoint,
 		NullCmd:            *nullCmd,
 		OperatingSystem:    *operatingSystem,
-		CreatedByArg:       createdBy,
-		AuthorArg:          defaultAuthor,
+		CreatedBy:          createdBy,
+		Author:             defaultAuthor,
 		LabelsArray:        labelsArray[:],
 		Ports:              ports[:],
 		Volumes:            volumes[:],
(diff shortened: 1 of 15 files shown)


container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+)
+
+const (
+	// defaultProcArch is the default architecture type based on legacy create_image_config.py.
+	defaultProcArch = "amd64"
+	// defaultTimeStamp is the unix epoch 0 time representation in 32 bits.
+	defaultTimestamp = "1970-01-01T00:00:00Z"
+)
+
+// OverrideConfigOpts holds all configuration settings for the newly outputted config file.
```
smukherj1: Add comments for all fields.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	b := sha256.Sum256([]byte(""))
+	empty = hex.EncodeToString(b[:])
+	return
+}
+
+// extractValue returns the contents of a file pointed to by value if it starts with a '@'.
+func extractValue(value string) (string, error) {
+	if strings.HasPrefix(value, "@") {
+		f, err := os.Open(value[1:])
+		// f, err := os.Open(value)
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to read content from file %s", value)
+		}
+		return string(shaHash), nil
+	}
+	return value, errors.Wrapf(nil, "unexpected value, got: %s want: @{...}", value)
+}
+
+// keyValueToMap converts an array of strings separated by '=' into a map of key-value pairs.
+// if toFormat is set to True, adds a . in front of the value so that the string can be formatted.
```
smukherj1: Comment seems to be outdated

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+func keyValueToMap(value []string) map[string]string {
+	convMap := make(map[string]string)
+	var temp []string
+	for _, kvpair := range value {
+		temp = strings.Split(kvpair, "=")
+		key, val := temp[0], temp[1]
```
smukherj1: Check length of temp == 2. Return error if not.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+			}
+			formatArgs[key] = val
+		}
+		if err = scanner.Err(); err != nil {
+			return "", errors.Wrapf(err, "failed to read line from file %s", infofile)
+		}
+	}
+	// do string manipulation in order to mimic python string format.
+	// specifically, replace '{' with '{{.' and '}' with '}}'.
+	inpReformatted := strings.ReplaceAll(inp, "{", "{{.")
```
smukherj1: This is too error prone. I'm not sure but I think `{` could be used in contexts other than stamp variables and these replacements could corrupt those. Instead of doing this I suggest:
```
o := inp
for k, v := range formatArgs {
   o = strings.ReplaceAll(o, fmt.Sprintf("{%s}", key), value)
}
```
This should allow you to deprecate the type 'formattedString` and the function `formatWithMap`

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	return formatWithMap(inpReformatted, formattedString(formatArgs)), nil
+}
+
+// mapToKeyValue reverses a map to a '='-separated array of strings in {key}={value} format.
+func mapToKeyValue(kvMap map[string]string) []string {
+	keyVals := []string{}
+	concatenated := ""
+	for k, v := range kvMap {
+		concatenated = k + "=" + v
+		keyVals = append(keyVals, concatenated)
```
smukherj1: `keyVals := append(keyVals, fmt.Sprintf("%s=%s", k, v))`

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	}
+
+	sort.Sort(sort.StringSlice(keyVals))
```
smukherj1: `sort.Strings(keyVals)`

https://golang.org/pkg/sort/#Strings

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	return keyVals
+}
+
+// resolveVariables resolves the environment variables embedded in the given value using
+// provided map of environment variable expansions.
+// It handles formats like "PATH=$PATH:..."
+func resolveVariables(value string, environment map[string]string) (string, error) {
+	fmt.Printf("environment override %v", environment)
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	var errorMet = false
+	mapper := func(p string) string {
+		switch i.(type) {
+		case string:
+			return environment[p]
+		}
+		errorMet = true
+		return ""
+	}
+	if errorMet {
```
smukherj1: This will never be true here. This will only be true after `os.Expand` returns and a mapper invocation set `errorMet = true`. I suggest moving this check *after* os.Expand returns. Also, I would generate the error inside the mapper function and add the name of the bad key that was requested in the generated error.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	overrideInfo.ConfigFile.Author = "Bazel"
+	overrideInfo.ConfigFile.OS = overrideInfo.OperatingSystem
+	overrideInfo.ConfigFile.Architecture = defaultProcArch
+
+	var err error
+	// createTime stores the input creation time with macros substituted.
+	var createTime string
+	// creationTime is the RFC 3339 formatted time derived from createTime input.
+	creationTime, err := time.Parse(time.RFC3339, defaultTimestamp)
+	log.Printf("the very very first creationTime: %v", creationTime)
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	if err != nil {
+		return errors.Wrap(err, "Unable to parse the default unix epoch time 1970-01-01T00:00:00Z")
```
smukherj1: messages in error objects begin with small letters. Also, instead of harcoding the time here, just use `defaultTimestamp`

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	}
+
+	// Parse for specific time formats.
+	if overrideInfo.CreationTimeString != "" {
+		log.Printf("The CreationTimeString is: %s", overrideInfo.CreationTimeString)
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+		var unixTime float64
+		// Use stamp to as preliminary replacement.
+		if createTime, err = Stamp(overrideInfo.CreationTimeString, overrideInfo.StampInfoFile); err != nil {
+			return errors.Wrapf(err, "Unable to format creation time from BUILD_TIMESTAMP macros")
```
smukherj1: Unable -> unable

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+		log.Printf("The first createTime is: %v", createTime)
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+		// If creationTime is parsable as a floating point type, assume unix epoch timestamp.
+		// otherwise, assume RFC 3339 date/time format.
+		unixTime, err = strconv.ParseFloat(createTime, 64)
+		log.Printf("the unixTime after strconv is: %f", unixTime)
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+		// Assume RFC 3339 date/time format. No err means it is parsable as floating point.
+		if err == nil {
+			log.Println("nani")
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+			// Ensure that the parsed time is within the floating point range.
+			// Values > 1e11 are assumed to be unix epoch milliseconds.
+			if unixTime > 1.0e+11 {
+				unixTime = unixTime / 1000.0
+			}
+			// Construct a RFC 3339 date/time from Unix epoch.
+			sec, dec := math.Modf(unixTime)
+			log.Printf("sec: %v, dec: %v", sec, dec)
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+			creationTime = time.Unix(int64(sec), int64(dec*(1e9))).UTC()
+		} else {
+			creationTime, err = time.Parse(time.RFC3339, createTime)
+			if err != nil {
+				return errors.Wrapf(err, "failed to parse the %s into float, so assuming RFC3339", createTime)
+			}
+		}
+	}
+	log.Printf("the assigned CreationTime is: %v", creationTime)
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+			if err != nil {
+				return errors.Wrapf(err, "Unable to perform substitutions to Env variable %s", entry)
+			}
+			overrideInfo.Entrypoint[i] = stampedEntry
+		}
+		overrideInfo.ConfigFile.Config.Entrypoint = overrideInfo.Entrypoint
+	}
+
+	if overrideInfo.NullCmd == "True" {
+		log.Println("hellooooo")
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+			stampedCmd, err := Stamp(cmd, overrideInfo.StampInfoFile)
+			if err != nil {
+				return errors.Wrapf(err, "Unable to perform substitutions to Env variable %s", cmd)
+			}
+			overrideInfo.Command[i] = stampedCmd
+		}
+		overrideInfo.ConfigFile.Config.Cmd = overrideInfo.Command
+	}
+
+	log.Printf("the stamped Command: %v", overrideInfo.Command)
```
smukherj1: Remove

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	if err != nil {
+		errors.Wrapf(err, "Unable to perform substitutions to user %s", overrideInfo.User)
+	}
+	overrideInfo.ConfigFile.Config.User = stampedUser
+
+	environMap := keyValueToMap(overrideInfo.Env)
+	for key, valToBeStamped := range environMap {
+		stampedValue, err := Stamp(valToBeStamped, overrideInfo.StampInfoFile)
+		if err != nil {
+			return errors.Wrapf(err, "Error stamping value %s", valToBeStamped)
```
smukherj1: `Error` -> `error` and similar for many others below

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+			overrideInfo.Command[i] = stampedCmd
+		}
+		overrideInfo.ConfigFile.Config.Cmd = overrideInfo.Command
+	}
+
+	log.Printf("the stamped Command: %v", overrideInfo.Command)
+
+	stampedUser, err := Stamp(overrideInfo.User, overrideInfo.StampInfoFile)
+	if err != nil {
+		errors.Wrapf(err, "Unable to perform substitutions to user %s", overrideInfo.User)
```
smukherj1: Unable -> unable and similar for many others below

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	for key, valToBeStamped := range environMap {
+		stampedValue, err := Stamp(valToBeStamped, overrideInfo.StampInfoFile)
+		if err != nil {
+			return errors.Wrapf(err, "Error stamping value %s", valToBeStamped)
+		}
+		environMap[key] = stampedValue
+	}
+	// perform any substitutions of $VAR or ${VAR} with environment variables.
+	if len(environMap) != 0 {
+		var baseEnvMap map[string]string
```
smukherj1: `baseEnvMap := make(map[string]string)`

Can just remove the else clause below.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+	err = WriteConfig(overrideInfo.ConfigFile, overrideInfo.OutputConfig)
+	if err != nil {
+		return errors.Wrap(err, "Failed to create updated Image Config.")
+	}
+
+	return nil
+}
+
+// WriteConfig writes a json representation of a config file to outPath.
+func WriteConfig(configFile *v1.ConfigFile, outPath string) error {
```
smukherj1: Does this function need to be exported?

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+// and replaces it according to the map of parameters to values in `params`.
+func formatWithMap(format string, params formattedString) string {
+	msg := &bytes.Buffer{}
+	template.Must(template.New("").Parse(format)).Execute(msg, params)
+	return msg.String()
+}
+
+// Stamp provides the substitutions of variables inside {} using info in file pointed to
+// by stampInfoFile.
+func Stamp(inp string, stampInfoFile []string) (string, error) {
```
smukherj1: Does this function need to be exported?

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+type OverrideConfigOpts struct {
+	ConfigFile         *v1.ConfigFile
+	OutputConfig       string
+	CreationTimeString string
+	User               string
+	Workdir            string
+	NullEntryPoint     string
+	NullCmd            string
+	OperatingSystem    string
+	CreatedByArg       string
```
smukherj1: No need to have the `Arg` suffix for CreatedBy & Author

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+		// length of history is expected to match the length of diff_ids.
+		history := []v1.History{}
+		if len(overrideInfo.ConfigFile.History) > 0 {
+			history = overrideInfo.ConfigFile.History
+		}
+		var historyToAdd v1.History
+
+		for _, l := range layerDigests {
+			var createdBy = overrideInfo.CreatedByArg
+			if createdBy == "" {
```
smukherj1: Based on how OverrideConfig is called, createdBy can never be empty. I would suggest making this and the check for `Author` below a validate function on `OverrideConfigOpts`. Something like

```
func (o *OverrideConfigOpts) validate() error {
   if o.CreatedBy == "" {
       return errors.New("CreatedBy was not specified")
   }
   ...
}
```

and just call
```
if err := overrideInfo.validate(); err != nil {
   return errors.Wrap(err, "unable to create a new config because the given options to override the existing image config/generate a new config failed validation")
}
```

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+		}
+		overrideInfo.ConfigFile.History = history
+	}
+
+	if len(overrideInfo.EntrypointPrefix) != 0 {
+		newEntrypoint := append(overrideInfo.ConfigFile.Config.Entrypoint, overrideInfo.EntrypointPrefix...)
+		overrideInfo.ConfigFile.Config.Entrypoint = newEntrypoint
+	}
+
+	err = WriteConfig(overrideInfo.ConfigFile, overrideInfo.OutputConfig)
```
smukherj1: Combine with next line

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,453 @@
...
+		if err != nil {
+			errors.Wrap(err, "failed to create new error")
+		}
+		return "", err
+	}
+	return os.Expand(value, mapper), nil
+}
+
+// OverrideImageConfig updates the current image config file to reflect the given changes.
+func OverrideImageConfig(overrideInfo *OverrideConfigOpts) error {
```
smukherj1: This function is way too long. Suggest calling sections of the body into helper functions.

container/go/pkg/compat/overrideConfig.go
xwinxu: Yes, Heddy uses it.

container/go/pkg/compat/overrideConfig.go
xwinxu: cannot generate error in mapper function because os.Expand only takes a function that doesn't return any errors.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,545 @@
...
+	var baseEnvMap = make(map[string]string)
+	var err error
+	if len(overrideInfo.ConfigFile.Config.Env) > 0 {
+		baseEnvMap, err = keyValueToMap(overrideInfo.ConfigFile.Config.Env)
+		if err != nil {
+			return []string{}, errors.Wrapf(err, "error converting Config env array %v to map", overrideInfo.ConfigFile.Config.Env)
+		}
+	}
+
+	for k, v := range environMap {
```
smukherj1: This loop is modifying the `baseEnvMap` as it's looping through it. I think this is problematic because iteration through Go maps does not have any pre-defined order. So imagine a baseEnvMap with
```
"PATH": "$PATH:a",
"FOO": "$PATH:b",
```

If the map iterates over "PATH" followed by "FOO" we get:
```
"PATH": ":a",
"FOO": ":a:b",
```

However, if the map iterates over "FOO" followed by "PATH" we get:
```
"PATH": ":a",
"FOO": ":b",
```

The iteration order can change arbitrarily depending on the contents of the map. I think the correct implementation would not modify the baseEnvMap while resolving the variables. It should create a separate output map that contains the resolved values. So something like:
```
result := make(map[string]string)
for k, v := range environMap {
  var expanded string
  if expanded, err = resolveVariables(v, baseEnvMap); err != nil {
    return nil, errors.Wrapf(err, "unable to resolve environment variables in %s with content mapping at %s", k, v)
  }
  result[k] = expanded
```

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,545 @@
...
+		var expanded string
+		if expanded, err = resolveVariables(v, baseEnvMap); err != nil {
+			return []string{}, errors.Wrapf(err, "unable to resolve environment variables in %s with content mapping at %s", k, v)
+		}
+		if _, ok := environMap[k]; ok {
```
smukherj1: I don't think this condition can ever fail given the for loop is iterating over the keys of environMap

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,545 @@
...
+// OverrideImageConfig updates the current image config file to reflect the given changes.
+func OverrideImageConfig(overrideInfo *OverrideConfigOpts) error {
+	overrideInfo.ConfigFile.Author = "Bazel"
+	overrideInfo.ConfigFile.OS = overrideInfo.OperatingSystem
+	overrideInfo.ConfigFile.Architecture = defaultProcArch
+
+	var err error
+
+	// creationTime is the RFC 3339 formatted time derived from createTime input.
+	creationTime, err := time.Parse(time.RFC3339, defaultTimestamp)
```
smukherj1: This value is always overridden by the line `if creationTime, err = getCreationTime(overrideInfo); err != nil {` so it can be removed.

You probably need to update the `getCreationTime` function to return `time.Parse(time.RFC3339, defaultTimestamp)` if the override value is blank.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,545 @@
...
+			if err != nil {
+				return time.Unix(0, 0), errors.Wrapf(err, "failed to parse the %s into float, so assuming RFC3339", createTime)
+			}
+		}
+	}
+	return creationTime, nil
+}
+
+// updateConfigInfo returns an updated version of the input infoToStamp.
+func updateConfigInfo(overrideInfo *OverrideConfigOpts, infoToStamp []string) ([]string, error) {
```
smukherj1: `updateConfigInfo` is too generic a function name. This function is stamping the given slice of strings using the stamp file fields in the given overrideInfo. Suggest naming this function as follows:
```
func stampValues(values []string, stampInfoFiles []string) ([]string, error) {
```

Taking overrideInfo here is making this function accept way more input than it needs to.

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,545 @@
...
+
+// resolveVariables resolves the environment variables embedded in the given value using
+// provided map of environment variable expansions.
+// It handles formats like "PATH=$PATH:..."
+func resolveVariables(value string, environment map[string]string) (string, error) {
+	var i interface{}
+	i = "this is a string type"
+	var errorMet = false
+	mapper := func(p string) string {
+		switch i.(type) {
```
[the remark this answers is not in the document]
xwinxu: yes that is the purpose. It is a hack for the mapper function that is used in os.Expand because I do not want to loop through all the contents of the environment variables and add each one as a switch statement. So I am using the fact that they're all strings to do so.

container/go/cmd/create_image_config/createImageConfig.go
```diff
@@ -0,0 +1,143 @@
...
+	if *baseManifest != "" {
+		manifestBlob, err = ioutil.ReadFile(*baseManifest)
+		if err != nil {
+			log.Fatalf("Failed to read manifest file from %s: %v", *baseManifest, err)
+		}
+	} else {
+		// generating an empty manifest.
+		manifestBlob = []byte("{}")
+	}
+	if err = ioutil.WriteFile(*outputManifest, manifestBlob, os.ModePerm); err != nil {
```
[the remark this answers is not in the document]
xwinxu: actually can't, because we are using manifestBlob outside of if statements, and we need to also error check, so we can't assign to both at once else the compiler says that manifestBlob is unused.

container/go/pkg/compat/overrideConfig.go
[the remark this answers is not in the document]
xwinxu: The above will not work, as it requires the switch statements to work.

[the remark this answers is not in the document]
smukherj1: Not sure what you mean it won't work. This works for me https://play.golang.org/p/f2dZBZini50

PR #1032, opened 2019-07-29
Support stamping in new pusher

xiaohegong:
- Modified `new_push.bzl` and `pusher.go` to support stamping
   - which is enabled when a supported attribute contains a python format placeholder like `{BUILD_USER}`, by substituting in the placeholders with values from stamp info files. 
- Added e2e test `test_new_container_push_with_stamp` verifying the placeholders are resolved by testing pusher with two formats


container/go/pkg/utils/utils.go
```diff
@@ -44,3 +49,49 @@ func ReadImage(src, format string) (v1.Image, error) {
 
 	return nil, errors.Errorf("unknown image format %q", format)
 }
+
+// TODO: REMOVE these two functions copied from Winnie's PR, refactor after her PR is merged.
```
xiaohegong: The functions added in this file are from Winnie's PR, will be removed after her PR gets merged.

container/go/cmd/pusher/pusher.go
```diff
@@ -102,11 +104,22 @@ func main() {
 		log.Fatalf("Error reading from %s: %v", imgSrc, err)
 	}
 
-	if err := push(*dst, img); err != nil {
+	// Infer stamp info if provided and perform substitutions in the provided tag name.
+	formattedDst := *dst
+	if len(stampInfoFile) > 0 {
```
smukherj1: This if condition shouldn't be necessary

container/go/pkg/utils/utils.go
smukherj1: State which two functions are being referred to here and link the PR instead of saying `Winnie's PR`.

container/go/pkg/utils/utils.go
```diff
@@ -44,3 +49,49 @@ func ReadImage(src, format string) (v1.Image, error) {
...
+
+// formateWithMap takes all variables of format {{.VAR}} in the input string `format`
+// and replaces it according to the map of parameters to values in `params`.
+func formatWithMap(format string, params formattedString) string {
+	msg := &bytes.Buffer{}
+	template.Must(template.New("").Parse(format)).Execute(msg, params)
+	return msg.String()
+}
+
+// Stamp provides the substitutions of variables inside {} using info in file pointed to
```
smukherj1: Make this comment more information by giving an example and referring to the function parameters.

container/go/pkg/utils/utils.go
```diff
@@ -44,3 +49,49 @@ func ReadImage(src, format string) (v1.Image, error) {
...
+		return inp, nil
+	}
+	formatArgs := make(map[string]interface{})
+	for _, infofile := range stampInfoFile {
+		f, err := os.Open(infofile)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to open file %s", infofile)
+		}
+		defer f.Close()
+		// scanner reads line by line and discards '\n' character already
```
smukherj1: Seems like this comment is unfinished?

container/go/pkg/utils/utils.go
```diff
@@ -44,3 +49,49 @@ func ReadImage(src, format string) (v1.Image, error) {
...
+// and replaces it according to the map of parameters to values in `params`.
+func formatWithMap(format string, params formattedString) string {
+	msg := &bytes.Buffer{}
+	template.Must(template.New("").Parse(format)).Execute(msg, params)
+	return msg.String()
+}
+
+// Stamp provides the substitutions of variables inside {} using info in file pointed to
+// by stampInfoFile.
+func Stamp(inp string, stampInfoFile []string) (string, error) {
```
smukherj1: The implementation of this function is too complex imo. I think you should be able to do just the following:
Read the stamp file line by line as you are doing. Then, instead of using templates which is overkill here, simply replace all occurrences of `"{" + key + "}"` with the stamp value in the given input string.

Also, it might be a good idea to create a stamper struct that is initialized with a list of stamp files. The stamper would load all the files into maps of key value pairs. This stamper could then be given values to stamp and it would go through the pre-processed maps and replace any keys it finds. This way, you avoid read the stamp files for *every* value.

container/go/cmd/pusher/pusher.go
```diff
@@ -106,11 +108,20 @@ func main() {
...
+	formattedDst, err := compat.Stamp(*dst, stampInfoFile)
+	if err != nil {
+		log.Fatalf("Error resolving stamp info to destination %s: %v", *dst, err)
+	}
+	if formattedDst != *dst {
+		log.Printf("Destination %s was resolved to %s based on inferred stamp info.", *dst, formattedDst)
+	}
+
+	if err := push(formattedDst, img); err != nil {
 		log.Fatalf("Error pushing image to %s: %v", *dst, err)
```
smukherj1: Print `formattedDst` here

Issue #1051, opened 2019-08-03
Digester appears to be returning a different digest


dicarlo2:
This commit switched the digester to use go-containerregistry: https://github.com/bazelbuild/rules_docker/commit/1033b04bfd119cc18c9dd08e08f948f8008d9097. After testing out rules_docker with this commit, we noticed that the digest produced by the digester is different than what is reported by the container registry. Specifically, we pushed an image built with rules_docker to gcr and the digest reported is different from the one that we get locally with the digester.

Is this expected? Is there an alternate way to get the image digest without pushing the image?

nlopezgi:
hi @dicarlo2,
as part of the recent refactoring that we are working on to move from python to go containerregistry, the latest release of rules_docker includes changes that impact the produced digests of images.
The expectation of this change is that it should make it so that images produced with older versions of rules_docker and images produced with the latest version will have a different digest. However, once you push the new image to a registry, the digest should match the one produced by the new version of the rules.
Is it the case that the image you are producing has a digest that mismatches an image produced in the past and pushed to the registry. Or is it the case that a recently build image has a given digest, and once that is pushed the digest in the registry is not the same?
If the latter, then we definitely have a regression in our migration process. If so, could you provide repro instructions so we can look in detail at what is the issue?

@smukherj1 to track in case we do have a regression.

smukherj1:
@dicarlo2 Thanks for reporting this issue. It's mostly confirmed that [1033b04](https://github.com/bazelbuild/rules_docker/commit/1033b04bfd119cc18c9dd08e08f948f8008d9097) introduced a regression where the digest reported by `container_push` is different from what GCR will report for the uploaded image. This is because the python pusher used by `container_push` and the new Go digester are using slightly different manifests. The manifest used by the python pusher is the one that gets uploaded to GCR and thus the digest reported by GCR differs from the one reported by the Go digester. This shouldn't be an issue for the `new_container_push` rule. However, we are going to roll back the use of the Go digester in `container_image` and `container_push` to fix this for now.

nlopezgi:
closing for now as culprit has been reverted. We will try to roll forward those changes again in the future and take care the digest does not get messed up. Please let me know to reopen if you see any other issues.

PR #1058, opened 2019-08-06
Revert digester deps for container_push and container_image 

xiaohegong:
Reverted changes in `container_push` and `container_image` to use new digester in go. 
  - The digest differs since  `container_push` and `container_image` still rely on manifest created in python containerregistry, which results in a difference in the uploaded manifest and the manifest in `digester.go` used to compute sha256 sum. 
  - Resolves issue #1051


smukherj1:
This fixes #1051

PR #1062, opened 2019-08-07
Implement createImageConfig for use in container_image targets

xwinxu:
* Implement `createImageConfig.go` to write out modified config file and empty/base manifest file, maintaining parity with the current behaviour
* Implement an `overrideConfig.go` binary to mutate the base config file to match the updates supplied in the flags. This also contains the `Stamp` utility function that the digester is dependent on.

### DIFF
--- a/container/go/pkg/compat/overrideConfig.go
+++ b/container/go/pkg/compat/overrideConfig.go
@@ -1,0 +1,530 @@
+// Copyright 2016 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+////////////////////////////////////////////////
+// This package manipulates v2.2 image configuration metadata.
+// It writes out both a config file and a manifest for the v2.2 image.
+
+package compat
+
+import (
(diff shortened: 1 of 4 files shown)


container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,530 @@
...
+	return convMap, nil
+}
+
+// Stamp provides the substitutions of variables inside {} using info in file pointed to
+// by stampInfoFile.
+func Stamp(inp string, stampInfoFile []string) (string, error) {
+	if len(stampInfoFile) == 0 || inp == "" {
+		return inp, nil
+	}
+	formatArgs := make(map[string]interface{})
```
smukherj1: `formatArgs := make(map[string]string)`

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,530 @@
...
+			formatArgs[key] = val
+		}
+		if err = scanner.Err(); err != nil {
+			return "", errors.Wrapf(err, "failed to read line from file %s", infofile)
+		}
+	}
+
+	o := inp
+	for k, v := range formatArgs {
+		if vStr, ok := v.(string); ok {
```
smukherj1: Don't need this type conversion anymore after changing formatArgs to map from string to string

container/go/pkg/compat/overrideConfig.go
```diff
@@ -0,0 +1,530 @@
...
+}
+
+// keyValueToMap converts an array of strings separated by '=' into a map of key-value pairs.
+func keyValueToMap(value []string) (map[string]string, error) {
+	convMap := make(map[string]string)
+	var temp []string
+	for _, kvpair := range value {
+		temp = strings.Split(kvpair, "=")
+		if len(temp) != 2 {
+			return convMap, errors.New("value in array are not of format key=value")
```
smukherj1: include the vaue of kvpair in the error message for context. Something like `"%q is not in format key=value", kvpair`

PR #1063, opened 2019-08-07
Update image.bzl to optionally utilize new createImageConfig Go binary

xwinxu:
* Modify `image.bzl` to accept Golang flags

Note: the image_test.py will fail when using the new Go createImageConfig binary until the sha256 sums are updated to reflect these new changes. See https://github.com/bazelbuild/rules_docker/pull/1024 for how to update the tests accordingly.

### DIFF
--- a/container/image.bzl
+++ b/container/image.bzl
@@ -110,7 +110,7 @@ def _add_go_args(
         base_config,
         base_manifest,
         operating_system):
-    args = [
+    args += [
         "-outputConfig",
         "%s" % config.path,
     ] + [
@@ -154,7 +154,7 @@ def _add_go_args(
     if workdir:
         args += ["-workdir", workdir]
 
-    inputs = layer_names
+    inputs += layer_names
     for layer_name in layer_names:
         args += ["-layerDigestFile", "@" + layer_name.path]
 
@@ -203,7 +203,7 @@ def _add_legacy_args(
         base_config,
         base_manifest,
         operating_system):
-    args = [
+    args += [
         "--output=%s" % config.path,
     ] + [
         "--manifestoutput=%s" % manifest.path,
@@ -242,7 +242,7 @@ def _add_legacy_args(
     if workdir:
         args += ["--workdir=" + workdir]
 
-    inputs = layer_names
+    inputs += layer_names
     for layer_name in layer_names:
         args += ["--layer=@" + layer_name.path]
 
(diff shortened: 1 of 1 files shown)


container/image.bzl
```diff
@@ -468,12 +552,22 @@ def _impl(
 _attrs = dicts.add(_layer.attrs, {
     "base": attr.label(allow_files = container_filetype),
     "cmd": attr.string_list(),
+    "legacy_create_image_config": attr.bool(
+        default = True,
+        doc = ("If set to False, the Go create_image_config binary will be ran instead."),
```
smukherj1: nit: ran -> run

container/image.bzl
```diff
@@ -124,84 +124,168 @@ def _image_config(
...
+            for x in ["/" + ctx.file.launcher.basename]:
+                args += ["-entrypointPrefix", "%s" % x]
+            args += ctx.attr.launcher_args
+        ctx.actions.run(
+            executable = ctx.executable.go_create_image_config,
+            arguments = args,
+            inputs = inputs,
+            outputs = [config, manifest],
+            use_default_shell_env = True,
+            mnemonic = "Go ImageConfig",
```
smukherj1: nit: Convention is to not have spaces in the mnemonic. Suggest making this GoImageConfig

container/image.bzl
```diff
@@ -124,84 +124,168 @@ def _image_config(
...
-        args += ["--stamp-info-file=%s" % f.path for f in stamp_inputs]
-        inputs += stamp_inputs
-
-    if ctx.attr.launcher_args and not ctx.attr.launcher:
-        fail("launcher_args does nothing when launcher is not specified.", attr = "launcher_args")
-    if ctx.attr.launcher:
-        args += [
-            "--entrypoint_prefix=%s" % x
-            for x in ["/" + ctx.file.launcher.basename] + ctx.attr.launcher_args
+    if ctx.attr.legacy_create_image_config:
```
smukherj1: Suggest creating two functions that take "args" & "inputs" and adds the relevant arguments & inputs for the python & go cases. This way, the core impl function will be smaller. So something like:
```
args = []
inputs = []
exec = None
if ctx.attr.legacy_create_image_config:
  _add_legacy_args(args, inputs, manifest, entrypoint, ...)
  exec = ctx.executable.create_image_config
else:
  _add_go_args(args, inputs, manifest, entrypoint, ...)
  exec = ctx.executable.go_create_image_config
ctx.actions.run(
        executable = exec,
        arguments = args,
        inputs = inputs,
        outputs = [config, manifest],
        use_default_shell_env = True,
        mnemonic = "ImageConfig",
    )
```

container/image.bzl
xwinxu: I'm from the south, what can I say

PR #1082, opened 2019-08-12
README.md: add the K8s Container Image Promoter to adopters

listx:
Add the Container Image Promoter project to the list of adopters.

Technically the promoter is not a "company" but it certainly uses rules_docker.


nlopezgi:
Not running gcb checks as this is a doc only change PR.

Issue #1106, opened 2019-08-25
container_push skip tag if sha256 exists


prestonvanloon:
Context: https://groups.google.com/d/msg/bazel-discuss/UW3tDLHnKmo/xbreLVCOGAAJ

Is there some way to skip container_push if the docker image digest is already on the remote and already has a tag? The issue is that using the git commit hash in the image tag will push new tags even if the image has not changed. We have some digests with dozens of tags and would like to prevent pipeline triggers on new tags when there is nothing changed.

smukherj1:
Hi @prestonvanloon,

Thanks for opening the issue. As far as I can tell, `container_push` will always re-tag the image if the specified tag is different even if the digest is unchanged. We are open to a contribution that adds a boolean attribute to `container_push` which would skip tagging the image if the digest is unchanged. Also, note that we are currently in the process of migrating our rules from using the python containerregistry library to the Go containerregistry library. So if you do decide to implement this feature, I would suggest doing it for [new_container_push](https://github.com/bazelbuild/rules_docker#new_container_push) instead of `container_push`. Some important pointers:
1. Update the list of attributes accepted by the `new_container_push` rule [here](https://github.com/bazelbuild/rules_docker/blob/8650f276d214e84e43c1253bd839b838303d4197/container/new_push.bzl#L167). The attribute should default to false for backwards compatibility.
2. Modify this [list](https://github.com/bazelbuild/rules_docker/blob/8650f276d214e84e43c1253bd839b838303d4197/container/new_push.bzl#L40) to add a flag if the new attribute is set.
3. Add support for the new flag in the pusher Go binary [here](https://github.com/bazelbuild/rules_docker/blob/8650f276d214e84e43c1253bd839b838303d4197/container/go/cmd/pusher/pusher.go). The logic that skips the upload if the digest is unchanged would probably go just before the line [here](https://github.com/bazelbuild/rules_docker/blob/8650f276d214e84e43c1253bd839b838303d4197/container/go/cmd/pusher/pusher.go#L138).

zbarahal:
@prestonvanloon if you have time can you take a pass at the linked PR and make sure it addresses your use case.

prestonvanloon:
@zbarahal Thanks for quick resolution. The new container push is not backwards compatible so using new container push as a drop in with docker_push doesn't work.

I'll wait until the migration is complete I suppose. The big issue is that new container doesn't support bundle which is my main use case.

smukherj1:
[...]

PR #1112, opened 2019-08-27
Add a flag to control pushing new tags on unchanged images

zbarahal:
My team ran into a use case where due to Bazel pushing a new tag to ECR for each service every time the monorepo changes we ran into the 100 tag per image cap on ECR for services that don't change often. This is because ECR groups tags onto images by digest so if your image isn't changing but you keep pushing new tags you end up with a bunch of tags on one image.

Per https://github.com/bazelbuild/rules_docker/issues/1106 I added a new flag to the new_push functionality and updated the pusher. I may explore adding this to the old fast pusher as well.

Closes https://github.com/bazelbuild/rules_docker/issues/1106


zbarahal:
@smukherj1 Integration tests are added and CI is passing. The PR suggests adding @alex1545 as a reviewer but since we have been discussing it should you be the reviewer? Either way is fine by me but this is ready for review now.

Still working on the CLA but we'll have that done soon.

zbarahal:
There are also a lot of CI builds attached to this PR and I'm not sure which are required. Travis is passing and I've cleaned up the formatting for buildifier. Let me know if other CIs are failing that are required.

zbarahal:
@googlebot I signed it!

smukherj1:
@googlebot rescan

zbarahal:
Thanks @smukherj1!

Issue #927, opened 2019-06-25
new_container_push: Error reading from xxx-layer.tar: file manifest.json not found in tar


factuno-db:
Using new_container_push (introduced yesterday with commit 5647f4f7a6b7c247e788675963e2e03a6e7156e1) results in

`2019/06/25 18:28:14 Running the Image Pusher to push images to a Docker Registry...
2019/06/25 18:28:14 Error reading from bazel-out/k8-fastbuild/bin/[...]/web_image-layer.tar: open bazel-out/k8-fastbuild/bin/[...]/web_image-layer.tar: no such file or directory
`

The file actually exists.

nlopezgi:
fyi @smukherj1

factuno-db:
Commit ac23af6914f0d9fc4dd6b853f6d016ba3a08006a results in:

2019/06/26 20:07:12 Running the Image Pusher to push images to a Docker Registry...
2019/06/26 20:07:12 Error reading from /[...]/1597cabd3fc5ac127946332b9db4062c/execroot/__main__/bazel-out/k8-fastbuild/bin/[...]/web_image_push.runfiles/__main__/[...]/web_image-layer.tar: file manifest.json not found in tar

xiaohegong:
Hi @factuno-db, thank you for reporting the issue.  The `new_container_push` supports pushing tarballs of the `docker save` and our old `container_pull` formats. In addition, we also support pushing image indices in [OCI layout](https://github.com/opencontainers/image-spec/blob/master/image-layout.md#oci-image-layout-specification). Do you mind letting us know how you got your tarball?

factuno-db:
hi @xiaohegong, via `go_image`.

factuno-db:
If I'm seeing this correctly, the old push.bzl calls `get_from_target` from layer_tools.bzl, which calls `_extract_layers`. `_extract_layers` defines the manifest file and calls extract_config.py which generates it.

xiaohegong:
Thank you for pointing this out. We are still in the process of implementing `new_container_push`, so some of the features are not fully supported yet. We will let you know when it is ready. 

As a temporary workaround, you can `docker save` the loaded `go_image` and push that tarball. Hope that helps.

factuno-db:
Alright, so I am closing this issue as new_container_push is WIP. Thanks a lot!

smukherj1:
rules_docker v0.10.0 includes the fix for this issue. Please let us know if that's not the case.

Issue #1075, opened 2019-08-10
layers from manifest don't match image configuration


adamyi:
I got `layers from manifest don't match image configuration` when pulling a container pushed with `container_push` to gcr. Might be related to https://github.com/docker/distribution/issues/1439

Repro:
```
WORKSPACE
new_container_pull(
    name = "tomcat-jython",
    registry = "index.docker.io",
    repository = "adamyi/tomcat-jython",
    tag = "latest",
)
BUILD
container_image(
    name = "image",
    base = "@tomcat-jython//image",
    visibility = ["//visibility:public"],
)
```

adamyi:
This seems to be an issue with `new_container_pull`. I switched back to `container_pull` and it no longer triggers this error.

smukherj1:
Hi @adamyi,

Thanks for opening this issue. `new_container_pull` is a new rule using a new Go backend and might have issues. We will take a look at why this is happening as this might be a bug in the new Go implementation. In the meantime, we suggest using `container_pull`. Keeping this issue open to track looking into why `new_container_pull` is unable to pull this image.

smukherj1:
The fix for this is included in rules_docker release v0.10.0. Please let us know if that's not the case.

PR #1118, opened 2019-08-29
Make pusher & digester backwards compatible.

smukherj1:
Changes:
1. --format arg dictates the format of the generated manifest to be
   pushed to the remote registry instead of the format of the incoming
   image.
2. Simplify the image source argument from overloading the "--src" flag to
   having separate flags for config & tarball files.
3. Implements converting an arbitrary v1.Image to a v1.Image in OCI format. Using this to add support in the pusher & digester for --format=OCI.


smukherj1:
This fixes #927 and #1075

smukherj1:
This closes #927

container/go/pkg/oci/image.go
```diff
@@ -0,0 +1,192 @@
...
+
+	"github.com/google/go-containerregistry/pkg/v1"
+	"github.com/google/go-containerregistry/pkg/v1/types"
+	"github.com/pkg/errors"
+)
+
+// ociLayer extends a v1.Layer pretenting to be a layer in the equivalent OCI
+// format.
+type ociLayer struct {
+	v1.Layer
```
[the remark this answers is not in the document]
smukherj1: This avoids from from having to provide implementations for functions already implemented by v1.Layer. This embedding is the reason I didn't have to define the functions in your other comment "I was wondering where functions RawConfigFile and ConfigName (below) are defined for the ociImage struct."

container/go/pkg/oci/image.go
```diff
@@ -0,0 +1,192 @@
...
+// encoded string.
+func sha256Digest(blob []byte) string {
+	rawDigest := sha256.Sum256(blob)
+	return hex.EncodeToString(rawDigest[:])
+}
+
+// buildManifest generates a manifest in the OCI schema 1 format for the
+// underlying v1.Image.
+func (i *ociImage) buildManifest() error {
+	rawConfig, err := i.RawConfigFile()
```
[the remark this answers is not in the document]
smukherj1: This is answered by your other question "Is there any special purpose/advantage defining an anonymous field here?
Same question for the ociImage struct." :)

container/go/pkg/utils/utils.go
```diff
@@ -31,28 +31,21 @@ func (f *ArrayStringFlags) Set(value string) error {
 	return nil
 }
 
-// ReadImage returns a v1.Image after reading an legacy layout, an OCI layout or a Docker tarball from src.
-func ReadImage(src, format, configPath, tarballBase string, layersPath []string) (v1.Image, error) {
-	if format == "oci" {
-		return oci.Read(src)
+// ReadImage returns a v1.Image after reading an image in OCI or Docker format.
```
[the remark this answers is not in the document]
smukherj1: Done

container/new_push.bzl
```diff
@@ -59,65 +59,45 @@ def _impl(ctx):
...
+    blobs = image.get("zipped_layer", [])
+    config = image["config"]
+    legacy_files = [config] + blobs
+    pusher_input += legacy_files
+    digester_input += legacy_files
+    tarball = image.get("legacy")
+    if tarball:
+        print("Pushing an image based on a tarball can be very " +
+              "expensive.  If the image is the output of a " +
+              "docker_build, consider dropping the '.tar' extension. " +
```
nlopezgi: update this doc (docker_build is an alias for container_image)

container/new_push.bzl
```diff
@@ -59,65 +59,45 @@ def _impl(ctx):
...
+              "If the image is checked in, consider using " +
+              "docker_import instead.")
```
nlopezgi: container_import

container/new_push.bzl
```diff
@@ -169,13 +145,12 @@ def _impl(ctx):
 new_container_push = rule(
     attrs = dicts.add({
         "format": attr.string(
-            default = "legacy",
+            default = "Docker",
             values = [
-                "oci",
-                "docker",
-                "legacy",
+                "OCI",
```
nlopezgi: are we losing any functionality by not supporting legacy anymore? If so, should we document this better, and even possibly cut a release before this.

tests/container/BUILD
```diff
@@ -378,7 +378,7 @@ py_test(
 # BEGIN_DO_NOT_IMPORT
 new_container_push(
     name = "new_push_test_skip_unchanged_digest_unchanged_tag_1",
-    format = "docker",
+    format = "Docker",
```
nlopezgi: cant we just remove this since its the default?

container/go/pkg/oci/image.go
```diff
@@ -0,0 +1,192 @@
+/// Copyright 2015 The Bazel Authors. All rights reserved.
```
nlopezgi: Do we have any level of unit testing for this code (and other bits of go code you are modifying)? Should we (maybe in a different PR given the potential effort / size)?

container/new_push.bzl
smukherj1: Done

container/new_push.bzl
smukherj1: Done

container/new_push.bzl
smukherj1: No we are not losing any functionality. This change just treats the attributes differently. What was supported by `legacy` before is now supported by `Docker`. This makes `new_container_push` a drop-in replacement for `container_push`. Since `new_container_push` is experimental, I don't see a reason to cut a release because the API is subject to change and we didn't *remove* any functionality.

tests/container/BUILD
smukherj1: Instead, I made the attribute mandatory. My goal is to make `new_container_push` a drop-in replacement for `container_push` and this attribute is mandatory in `container_push`.

container/go/pkg/oci/image.go
smukherj1: This code is end to end tested. We spin up a local docker registry and push images in OCI format to that registry. The docker registry is in the best position to validate an image. However, I have an upcoming change that is going to fix a performance issue in the current Go code & functionality gaps which I plan to add more tests for. I might make those a combination of `container_test` for *unit-test* style testing & end to end tests for more thorough docker image schema validation. We can discuss more in that PR (will be after one more PR where I add a `container_flatten` implementation in Go).

But in general yes, I acknowledge this comment and in general, I plan to add some tests before the big switch.

PR #1122, opened 2019-09-03
Implement a container flattener in Go

smukherj1:
Keeping the python implementation as the default flattener but an attribute allows using the Go flattener. Added a test for the Go flattener as well.


container/image_test.py
```diff
@@ -350,6 +350,11 @@ def test_flattened(self):
       self.assertTarballContains(tar, [
         '.', '/usr', '/usr/bin', '/usr/bin/java', './foo'])
 
+  def test_flattened(self):
```
nlopezgi: add short doc for this def, its otherwise not clear what differs from the one above

container/go/cmd/flattener/flattener.go
```diff
@@ -0,0 +1,84 @@
...
+	"github.com/bazelbuild/rules_docker/container/go/pkg/utils"
+	"github.com/google/go-containerregistry/pkg/v1/mutate"
+)
+
+var (
+	// The argument names match the arguments of https://github.com/google/containerregistry/blob/master/tools/fast_flatten_.py
+	// for backwards compatibility. Many of these arguments are not actually used
+	// but they are defined to make the Go flattener a drop-in replacement for the
+	// python flattener.
+	imgConfig          = flag.String("config", "", "Path to the image config file.")
```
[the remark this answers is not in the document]
smukherj1: These flags are refactored to be used in #1123 to improve performance.

container/image_test.py
smukherj1: Done

PR #1123, opened 2019-09-03
Refactor image reader used in Go binaries fixing foreign layer, tarball loading and avoiding digest recomputation

smukherj1:
This PR should be reviewed only after #1122 has been merged.

Optimize the image reader by using all available layer parts. This avoids recomputing the digest of the compressed & uncompressed layer blobs in the Go binaries if they are already available.
    
Refactored the pusher, digester, flattener & join_layers binaries to use all available layer parts to serve the layer digest & diffIDs. The previous implementation would read the entire layer and actually compute the digest instead of using the pre-computed digests.
    
Refactor the Starlark argument & input processing code for an image stored in the intermediate format into a single function and reused this function to provide the arguments to all the Go binaries.

Also fix a couple of issues:
1. The Go pusher, digester & flattener didn't support loading images with foreign layers which is fixed by this PR.
2. The Go pusher, digest & flattener wouldn't include the layers from the base image if the base image was specified as a tarball.


container/go/pkg/compat/reader.go
```diff
@@ -16,69 +16,314 @@
...
+	}
+	m, err := v1.ParseManifest(mf)
+	if err != nil {
+		return errors.Wrapf(err, "unable to parse base image manifest from %s", r.parts.BaseManifest)
+	}
+	r.baseManifest = m
+	return nil
+}
+
+// foreignLayer represents a foreign layer usually present in windows images.
```
nlopezgi: Capitalize Windows here and in other comments

container/go/pkg/compat/reader.go
```diff
@@ -16,69 +16,314 @@
...
+type foreignLayer struct {
+	// digest is the digest of this foreign layer.
+	digest v1.Hash
+	// diffID is the diffID of this foreign layer.
+	diffID v1.Hash
+	// size is the size of the foreign layer.
+	size int64
+	// urls are the urls where the actual contents of the foreign layer can
```
nlopezgi: urls are the URLs

container/go/pkg/compat/reader.go
```diff
@@ -16,69 +16,314 @@
...
+}
+
+// Compressed returns a blank reader for this foreign layer.
+func (l *foreignLayer) Compressed() (io.ReadCloser, error) {
+	return l.Uncompressed()
+}
+
+// Size returns the number of bytes in the compressed version of this layer.
+func (l *foreignLayer) Size() (int64, error) {
+	return 0, nil
```
nlopezgi: does not seem right to return here 0 unconditionally?

container/image_test.py
```diff
@@ -358,6 +358,11 @@ def test_flattened_go(self):
       self.assertTarballContains(tar, [
         '.', '/usr', '/usr/bin', '/usr/bin/java', './foo'])
 
+  def test_flattened_from_tarball_base(self):
```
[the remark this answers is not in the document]
smukherj1: Done

container/go/pkg/compat/reader.go
smukherj1: Done

container/go/pkg/compat/reader.go
smukherj1: Yes this is intentional. This returns the number of bytes in the blob returned by the `Compressed` function on this layer. Since the `Compressed` function returns an empty blob, the size needs to match or the image tarball will fail to build. The correct size and the location to download the image from is added to the generated manifest in the `genManifest` function in `compat/image.go` file. Docker's image loader is aware of this discrepancy and uses the manifest to process foreign layers instead of the actual layer blob included in the image.

smukherj1: The `container_test` target `//tests/container:basic_windows_image_test_go_join_layers` tests this case.

container/go/pkg/compat/reader.go
smukherj1: Done

container/go/cmd/digester/digester.go
```diff
@@ -21,33 +21,35 @@ import (
...
+	imgTarball   = flag.String("tarball", "", "Path to the image tarball as generated by docker save. Required if --config was not specified.")
+	imgConfig    = flag.String("config", "", "Path to the image config JSON file. Required if --tarball was not specified.")
+	baseManifest = flag.String("manifest", "", "Path to the manifest of the base image. This should be the very first image in the chain of images and is only really required for Windows images with a base image that has foreign layers.")
+	format       = flag.String("format", "", "The format of the uploaded image (Docker or OCI).")
+	layers       utils.ArrayStringFlags
 )
 
 func main() {
-	flag.Var(&layers, "layer", "The list of paths to the compressed layers of this docker image, only with --format=Docker.")
+	flag.Var(&layers, "layer", "One or more layers with the following comma separated values (Compressed layer tarball, Uncompressed layer tarball, digest file, diff ID file). e.g., --layer layer1.tar.gz,layer1.tar,<file with digest>,<file with diffID>.")
```
nlopezgi: should this be layer001.tar.gz,layer001.tar,  ?

container/go/cmd/flattener/flattener.go
```diff
@@ -22,49 +22,39 @@ import (
...
+	outConfig    = flag.String("metadata", "", "Path to the output image config.")
+	layers       utils.ArrayStringFlags
 )
 
 func main() {
-	flag.Var(&digests, "digest", "The list of layer digest filenames in order, e.g., --digest=layer1.sha256 --digest=layer2.sha256.")
-	flag.Var(&layers, "layer", "The list of compressed layer filenames in order, e.g., --layer=layer1.tar.gz --layer=layer2.tar.gz.")
-	flag.Var(&uncompressedLayers, "uncompressed_layer", "The list of uncompressed layer filenames in order, e.g., --uncompressed_layer=layer1.tar --uncompressed_layer=layer2.tar.")
-	flag.Var(&diffIDs, "diff_id", "The list of layer diff IDs filenames in order, e.g., --diff_id=diff_id1.sha256 --diff_id=diff_id2.sha256.")
+	flag.Var(&layers, "layer", "One or more layers with the following comma separated values (Compressed layer tarball, Uncompressed layer tarball, digest file, diff ID file). e.g., --layer layer1.tar.gz,layer1.tar,<file with digest>,<file with diffID>.")
```
nlopezgi: same as above

container/go/cmd/pusher/pusher.go
```diff
@@ -45,7 +46,7 @@ var (
 )
 
 func main() {
-	flag.Var(&layers, "layer", "The list of paths to the layers of this docker image, only used for legacy images.")
+	flag.Var(&layers, "layer", "One or more layers with the following comma separated values (Compressed layer tarball, Uncompressed layer tarball, digest file, diff ID file). e.g., --layer layer1.tar.gz,layer1.tar,<file with digest>,<file with diffID>.")
```
nlopezgi: same

container/go/pkg/compat/image.go
```diff
@@ -20,241 +20,257 @@ package compat
...
-// We assume the layer files are named in the format of e.g., 000.tar.gz in this path, following the order they appear in manifest.json.
-func (li *legacyImage) LayerByDigest(h v1.Hash) (partial.CompressedLayer, error) {
-	manifest, err := li.Manifest()
-	if err != nil {
-		return nil, err
-	}
+// legacyImage is the image in legacy intermediate format. Implements
+// partial.CompressedImageCore in go-containerregistry.
+type legacyImage struct {
+	// config is the path to the image config.
```
nlopezgi: configPath

container/go/pkg/compat/image.go
```diff
@@ -20,241 +20,257 @@ package compat
...
+		if err != nil {
+			return errors.Wrap(err, "unable to get digest of layer")
+		}
+		size, err := l.Size()
+		if err != nil {
+			return errors.Wrap(err, "unable to get size of layer")
+		}
+		var urls []string
+		if fl, ok := l.(*foreignLayer); ok {
+			// The size returned by the Size method on foreign is always zero.
```
nlopezgi: on foreign layers

container/go/pkg/compat/reader.go
```diff
@@ -16,69 +16,314 @@
...
+	}
+	return LayerParts{
+		CompressedTarball:   split[0],
+		UncompressedTarball: split[1],
+		DigestFile:          split[2],
+		DiffIDFile:          split[3],
+	}, nil
+}
+
+// ImageParts contains paths to a Docker image config and the invidual layer
```
nlopezgi: individual

container/go/pkg/compat/reader.go
nlopezgi: If that is the case please update the docs to explain a bit more why this is returning a fixed value.

container/go/cmd/digester/digester.go
smukherj1: Removed the number instead as discussed offline

container/go/cmd/pusher/pusher.go
smukherj1: Done

container/go/pkg/compat/image.go
smukherj1: Done

container/go/pkg/compat/image.go
smukherj1: Done

container/go/pkg/compat/reader.go
smukherj1: Done

container/go/cmd/flattener/flattener.go
smukherj1: Done

PR #1128, opened 2019-09-05
Refactor puller & loader to directly generate legacy format.

### LINKED ISSUE #1125: unable to pull mysql image successfully
I've tried to download the mysql image at sha256:1a121f2e7590f949b9ede7809395f209dd9910e331e8372e6682ba4bebcc020b that had successfully worked on earlier versions. I'm on macOS and bazel 0.29.0.

I have a reproduction that you can use for this ticket in branch `docker_mysql` in https://github.com/jmhodges/bazel_bugs ([branch in the web UI](https://github.com/jmhodges/bazel_bugs/tree/docker_mysql)). I've set it up to work like item 3 below. You can reproduce it by running `bazel run //dockertest:dockertest`.

On v0.9.0 and HEAD (165cd95365684f5fa1f29abc71cf1006e06c581a), there are different errors.

1. On v0.9.0 with `container_pull`, the sha256 of the image downloaded doesn't match the one asked for in `digest`. The error is similar to 3's.

2. On v0.9.0 with `new_container_pull`, the `docker load` step fails on an "invalid diffID for layer 2". The sha256 requested for layer 2 is correct, but the 3rd layer ("layer 2") actually loaded is layer 10. This is because the files on disk are, in ASCII order, "0.tar.gz.nogz", "1.tar.gz.nogz", and then "10.tar.gz.nogz". To load these layers in the correct order, a numeric sort on the prefix has to be used by something (I'm not sure what) instead of the ASCII sort it does.

3. On HEAD with both `new_container_pull` and `container_pull`, the error is "SHA256 of the image specified does not match SHA256 of the pulled image". The expected digest (the one entered into `digest` in new_container_pull) doesn't match what is downloaded somehow.

3. On HEAD with `new_container_pull` and using the latest `puller` in HEAD, the error is that the `digest` file doesn't exist in `external/mysql/image` when `cat` is run. (I imagine something similar will happen with `container_pull` but this was as long as I could spend on the bug)

### LINKED ISSUE #1127: new_container_pull loads layers in wrong order
Something within `new_container_pull` is loading layers in the wrong order in both HEAD (4109d47b8656464e3075f3af1f6c8705e0eec867) and v0.9.0. Instead of loading layers in their numeric order, its loading them in their ASCII ordering. That is, it's loading 0.tar.gz.nogz, then 1.tar.gz.nogz, and then 10.tar.gz.nogz and then 2.tar.gz.nogz, and so on, instead of loading 2.tar.gz.nogz third and loading 10.tar.gz.nogz in the eleventh spot.

In short, the loading is happening in ASCII order of the file names instead of numeric order.

This is showing up during the docker load stage of `bazel run`'ing a `container_image` target for mysql 5.7.25 (and others) as `"invalid diffID for layer 2: expected "sha256:9f77b78f01a7965ee78a5480ce5a592f992873e7b34400eb2d49a56598d2028f", got "sha256:2ebc9c2f59ec0b90db5ae34d2718dff7f1b5a0443062ca7cf3dd7742c5947ca7"`

I've got a reproduction up in the `docker_mysql_layers` branch of https://github.com/jmhodges/bazel_bugs ([branch in the web UI](https://github.com/jmhodges/bazel_bugs/tree/docker_mysql_layers))

Running `bazel run dockertest:dockertest` will demonstrate the error

(This ticket was broken out from #1125 and the reproduction is in a different branch in my bazel_bugs repo than was given in #1125)

smukherj1:
This will eventually fix #1127 by naming the compressed layer blobs padded to 3
characters e.g., 000.tar.gz, 001.tar.gz. This makes the generated layer
files ASCII sortable. Since docker layers are limited to something like
128, the layer files always sort correctly now.

This PR only changes the puller/pusher binary. Since `container_pull` and `container_load` are repository rules using a hosted GCS file as the binary, they won't see the changes until this PR is merged. Once this PR is merged, the new puller & pusher Go binaries will be uploaded to GCS and there will be a follow up PR updating the bzl files to use the new binaries.


container/go/cmd/puller/puller.go
```diff
@@ -86,10 +76,11 @@ func getTag(ref name.Reference) name.Reference {
 	return tag
 }
 
-// NOTE: This function is adapted from https://github.com/google/go-containerregistry/blob/master/pkg/crane/pull.go
-// with slight modification to take in a platform argument.
-// Pull the image with given <imgName> to destination <dstPath> with optional cache files and required platform specifications.
-func pull(imgName, dstPath, format, cachePath string, platform v1.Platform) error {
+// pull pulls the given given to the given destination directory. A cached
```
nlopezgi: fix?

PR #1130, opened 2019-09-06
Fix apt-get being too slow in download_pkgs in some ubuntu images

smukherj1:
Set a ulimit on the number of opened files instead of the default
unlimited.


smukherj1:
Well this didn't help

PR #1142, opened 2019-09-11
Use Go impl instead of python in all remaining container rules

smukherj1:
Also update tests with new digests & other fixes.


container/flatten.bzl
```diff
@@ -74,7 +74,7 @@ container_flatten = rule(
         ),
         # TODO (smukherj1): Remove once migration in #580 is done.
         "use_legacy_flattener": attr.bool(
-            default = True,
+            default = False,
             doc = "Use the legacy python flattener to generate the image " +
                   "filesystem tarball. Uses the experimental Go implementation" +
```
nlopezgi: r/ experimental

container/image.bzl
```diff
@@ -352,23 +351,16 @@ def _repository_name(ctx):
...
-    args.add_all(blobsums, format_each = "--digest=%s")
-    blobs = image.get("zipped_layer", [])
-    args.add_all(blobs, format_each = "--layer=%s")
-    args.add(image["config"], format = "--config=%s")
-    args.add(output_digest, format = "--output-digest=%s")
-
-    if image.get("legacy"):
-        args.add(image["legacy"].path, format = "--tarball=%s")
+    args, inputs = _gen_img_args(ctx, image)
+    args.append("--dst=%s" % output_digest.path)
```
nlopezgi: create an args object and add_all the args produced by _gen_img_args but then you dont use append to add the output_digest, which should be slightly better wrt efficiency

container/layer_tools.bzl
```diff
@@ -185,7 +185,7 @@ def _add_join_layers_go_args(args, inputs, images):
         inputs += image["blobsum"]
 
         if image.get("legacy"):
-            args.add(image["legacy"].path, "--tarball=%s")
+            args.add("--tarball", image["legacy"])
```
nlopezgi: iirc, I had some tests complain that image["legacy"] was a DirectDependency type and not a File type, but if tests are passing here then this is better.

container/push.bzl
```diff
@@ -11,16 +11,16 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-"""An implementation of container_push based on google/containerregistry.
-
-This wraps the containerregistry.tools.fast_pusher executable in a
+"""An new implementation of container_push based on google/go-containerregistry.
```
nlopezgi: s/ An new implementation / An implementation

container/push.bzl
```diff
@@ -148,10 +140,10 @@ container_push = rule(
                 "OCI",
                 "Docker",
             ],
-            doc = "The form to push: Docker or OCI.",
+            doc = "The form to push: Docker or OCI, default to 'Docker'.",
         ),
         "image": attr.label(
-            allow_single_file = [".tar"],
+            allow_files = True,
```
nlopezgi: why change from allows_single_file to allow_files?

container/flatten.bzl
smukherj1: Done

container/image.bzl
smukherj1: Done

container/layer_tools.bzl
smukherj1: Yes tests pass. I don't think this particular line was previously tested as it failed without my change.

container/push.bzl
smukherj1: Done

container/push.bzl
smukherj1: Fixed. I didn't explicitly change this. I just renamed new_push.bzl to push.bzl.

PR #1145, opened 2019-09-13
Allow callers to specify entrypoint in py_image macro

NathanHowell:
Different base images have `python` installed in different locations, we should be able to specify the correct path:

```
❯ docker run --rm -it python:alpine which python
/usr/local/bin/python
```

### DIFF
--- a/python/image.bzl
+++ b/python/image.bzl
@@ -72,7 +72,7 @@ def py_layer(name, deps, filter = "", **kwargs):
     native.py_library(name = binary_name, deps = deps, **kwargs)
     filter_layer(name = name, dep = binary_name, filter = filter)
 
-def py_image(name, base = None, deps = [], layers = [], **kwargs):
+def py_image(name, base = None, deps = [], layers = [], entrypoint = None, **kwargs):
     """Constructs a container image wrapping a py_binary target.
 
     Args:
@@ -106,7 +106,7 @@ def py_image(name, base = None, deps = [], layers = [], **kwargs):
     app_layer(
         name = name,
         base = base,
-        entrypoint = ["/usr/bin/python"],
+        entrypoint = [entrypoint or "/usr/bin/python"],
         binary = binary_name,
         visibility = visibility,
         tags = tags,
--- a/python3/image.bzl
+++ b/python3/image.bzl
@@ -66,7 +66,7 @@ DEFAULT_BASE = select({
     "//conditions:default": "@py3_image_base//image",
 })
 
-def py3_image(name, base = None, deps = [], layers = [], **kwargs):
+def py3_image(name, base = None, deps = [], layers = [], entrypoint = None, **kwargs):
     """Constructs a container image wrapping a py_binary target.
 
   Args:
@@ -101,7 +101,7 @@ def py3_image(name, base = None, deps = [], layers = [], **kwargs):
     app_layer(
         name = name,
         base = base,
-        entrypoint = ["/usr/bin/python"],
+        entrypoint = [entrypoint or "/usr/bin/python"],
         binary = binary_name,
         visibility = visibility,
         tags = tags,


NathanHowell:
Do you have suggestions on how I should structure tests for this?

nlopezgi:
Would it be possible to just use a container_image target that adds a symlink in the right location and just use that as base? Example:
(add in WORKSPACE a container pull for alpine_python)
```
container_image (
    name = "alpine_python_w_symlink",
    base = "@alpine_python//image:image",
    symlinks { "/usr/bin/python" : "/usr/local/bin/python"},
)

py_image(
    name = "your_image",
    base = ":alpine_python_w_symlink",
    ...
)
```
If so then lets prefer that to extending the API for these rules.

NathanHowell:
@nlopezgi sure, it's the only way to make things work though it wasn't obvious or well documented. so I was curious if there is a better way. this removes the only hardcoded string in the macro.

nlopezgi:
@NathanHowell,
thanks for your contributions. 
Our current docs about usage of xx_image rules state that the recommended path is to use a custom base to modify any attrs / properties of the image you would need to use in the xx_image rules (https://github.com/bazelbuild/rules_docker#go_image-custom-base). Not sure we should add content there for specific use cases such as this one, but I'm open to suggestions.
As for this being the only hardcoded string in the python.bzl rule, I'm not sure having hardcoded info there is an issue (we do similarly for most other xx_image rules) and the challenge with your proposed solution is that it extend the API and test surface. Making this change would require proper testing, and would open the door for other xx_image rules to also request this feature (and thus increase even more the API and test surface for these rules). For this reason, for xx_image rules, we tend to push back on most requests that increase the API surface when there is a viable workaround available with a container_image base.
Thanks again for your contribution, if you feel the docs above can use some additional comments that would have helped you, please send a PR with suggested changes.

NathanHowell:
@nlopezgi would it make sense to pull the interepreter path from something like [container_py_toolchain](https://github.com/bazelbuild/rules_docker/blob/master/toolchains/python/BUILD#L23)? at first glance, it looks like the nodejs images may do this.

in this particular case, a distroless image for python 3.7 would have been sufficient. but it's been generally difficult to discover the translation from Dockerfiles to the various Bazel container rules (among other challenges).

nlopezgi:
I'm not sure it would be advisable to pull that path from the container_py_toolchain. @brandjon What do you think about this issue. tl;dr we have a hardcoded path that is used to launch the py wrapper that bazel produces as part of a py_image target. How should this be configurable via toolchains, if at all, in your opinion?

As for translating a dockerfile to rules_docker, we have some examples here: https://github.com/bazelbuild/rules_docker/tree/master/testing/examples. But as you mention, the ecosystem is complex, getting a distroless py 3.7 image would be nice, but its is something that requires coordination with other teams to assess viability / timelines (i.e., we just use what other teams make available here). But in general, I understand your concern and please open other issues if you find yourself stuck at any moment when using these rules.

brandjon:
> How should this be configurable via toolchains, if at all, in your opinion?

I'm not entirely sure of the usage here. In principle, if your py wrapper target needs to run under a Python interpreter (on its target platform), you'd want the py wrapper rule to declare a dependency on the Python toolchain type, and retrieve the interpreter from the resolved toolchain's `PyRuntimeInfo`.

`py_binary` also passes along the `PyRuntimeInfo` it's using as a provider to any other rules that depend on it. So you might be able to use that, as long as your target platform hasn't changed in between.

nlopezgi:
Thanks for your comments @brandjon ,
so my initial feel is that implementing this the right way, via toolchains (and ideally for all xx_image rules, not just py_image/py3_image), is more effort than what it actually buys us at this time, given that we have the available workaround of adding the symlink to any base image.
@NathanHowell if you think this is something that would make your life significantly easier, and are willing to put in the effort to do it properly (which would involve becoming knowledgeable on how python toolchains work) I'm willing to put in the time to help you get this working and properly tested.

PR #1147, opened 2019-09-13
Implement Get function for struct of type Value

alex1545:


contrib/go/cmd/metadata_merge/metadata_merge.go
```diff
@@ -47,6 +47,12 @@ func (l *strArgList) Set(value string) error {
 	return nil
 }
 
+// Get returns an empty interface that may be type-asserted to the underlying
+// value of type []string.
+func (l *strArgList) Get() interface{} {
+	return l.Args
```
smukherj1: I think the convention here is to return an empty value. So `return []string{}`

alex1545: As discussed over the chat, from others' implementation of this method, it seems that the convention is to return the actual field of the underlying struct.

PR #1157, opened 2019-09-18
more use ctx.args

nlopezgi:
Following up on https://github.com/bazelbuild/rules_docker/pull/1156


nlopezgi:
also tested locally with legacy_create_image_config with default value set to True

PR #1159, opened 2019-09-19
More use ctx args

nlopezgi:
Includes fixes that led to https://github.com/bazelbuild/rules_docker/pull/1158


nlopezgi:
tested:
Ran //testdata:py_image_complex.tar (note the .tar is important to ensure join_layers actually runs) after flipping [this](https://github.com/bazelbuild/rules_docker/blob/5b4c97b0322414f754bf7e0cb7ec384d2462d55d/container/layer_tools.bzl#L346)

PR #1166, opened 2019-09-24
fix json print command for py3

mattymo:
Python3 is preferred for running bazel, but somehow some code was committed using incompatible code.


alex1545:
Thank you for sending this PR.
Will run the CI and approve.

Issue #1105, opened 2019-08-22
Incompatible Python toolchain


mdvorsky:
When I include `rules_docker` in my `WORKSPACE`, my default python toolchain suddenly changes and no longer works on MacOSX:

```
INFO: ToolchainResolution: Looking for toolchain of type @bazel_tools//tools/python:toolchain_type...
INFO: ToolchainResolution:   For toolchain type @bazel_tools//tools/python:toolchain_type, possible execution platforms and toolchains: {@bazel_tools//platforms:host_platform -> @io_bazel_rules_docker//toolchains/python:default_container_py_runtime_pair}
```

https://github.com/bazelbuild/rules_docker/blob/master/toolchains/python/BUILD#L29

```
py_runtime(
    name = "default_container_py3_runtime",
    interpreter_path = "/usr/bin/python3",
    python_version = "PY3",
)
```

```
$ ls /usr/bin/python3
ls: /usr/bin/python3: No such file or directory
```

The default python toolchain `bazel_tools//tools/python:_autodetecting_py_runtime_pair` which I get if I don't include `rules_docker`:

```
INFO: ToolchainResolution:   For toolchain type @bazel_tools//tools/python:toolchain_type, possible execution platforms and toolchains: {@bazel_tools//platforms:host_platform -> @bazel_tools//tools/python:_autodetecting_py_runtime_pair}
```

works correctly.

```
$ bazel version
Build label: 0.28.1-homebrew
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Fri Jul 19 16:09:50 2019 (1563552590)
Build timestamp: 1563552590
Build timestamp as int: 1563552590
```

nlopezgi:
This is likely a duplicate of https://github.com/bazelbuild/rules_docker/issues/1093. Closing but let me know to reopen if the pointed issue (and its recommended solution) does not work for you.

nlopezgi:
See also https://github.com/bazelbuild/rules_docker/issues/1087

mdvorsky:
Thank you!

I'm not sure I understand what is the recommended solution. Do I need to define my own python runtime and then use `--host_platform` and `--platform`? 

[...]

PR #1173, opened 2019-10-02
register toolchain only for container exec and use it in py*_image

nlopezgi:
The python toolchain we are registering for execution inside the container seems to be messing up with execution outside the container as documented here https://github.com/bazelbuild/rules_docker#known-issues-1. `target_compatible_with` should now be supported on all targets, so restricting our toolchain using `target_compatible_with` should make this error go away.
See https://github.com/bazelbuild/rules_docker/issues/1105


.bazelci/presubmit.yml
```diff
@@ -6,7 +6,7 @@ buildifier:
...
   # TODO (suvanjan): Re-enable once issues are fixed.
-  warnings: "attr-cfg,attr-license,attr-non-empty,attr-output-default,attr-single-file,confusing-name,constant-glob,ctx-actions,ctx-args,depset-iteration,depset-union,dict-concatenation,duplicated-name,filetype,function-docstring,git-repository,http-archive,integer-division,load,load-on-top,module-docstring,name-conventions,native-build,native-package,no-effect,out-of-order-load,output-group,package-name,package-on-top,positional-args,redefined-variable,repository-name,same-origin-load,string-iteration,unreachable,unsorted-dict-items,unused-variable"
+  warnings: "attr-cfg,attr-license,attr-non-empty,attr-output-default,attr-single-file,confusing-name,constant-glob,ctx-actions,ctx-args,depset-iteration,depset-union,dict-concatenation,duplicated-name,filetype,function-docstring,git-repository,http-archive,integer-division,load,load-on-top,module-docstring,name-conventions,native-build,native-package,no-effect,out-of-order-load,output-group,package-name,package-on-top,redefined-variable,repository-name,same-origin-load,string-iteration,unreachable,unsorted-dict-items,unused-variable"
```
nlopezgi: removed "positional-args"

PR #1189, opened 2019-10-08
Remove references to python binaries using python containerregistry

smukherj1:


nlopezgi:
undo approval until its ready and passing presubmit

PR #1196, opened 2019-10-09
Fix parsing of environment variables

nickjmeyer:
Fix parsing of environment variables when the value contains an `=` character.


nlopezgi:
This seems like it's the same as https://github.com/bazelbuild/rules_docker/pull/1192?

nickjmeyer:
It sure does.  I'll close this.

PR #1201, opened 2019-10-11
Fix writing uncompressed images & add test

smukherj1:


smukherj1:
This was discovered by someone trying to pull nixery images that are uncompressed by default [comment](https://github.com/google/nixery/pull/64#issuecomment-540830197) . Nixery was updated by this [PR](https://github.com/google/nixery/pull/64) to always compress images now but there's no reason why rules_docker shouldn't be able to pull uncompressed images.

PR #1217, opened 2019-10-15
Set `container_test` to be `testonly`

virusdave:
Currently, one cannot use `container_test` targets that depend on other test targets,
as `container_bundle` is not a `testonly=True` target.  This change allows propagation
of the `testonly` arg to support this use case.

### DIFF
--- a/contrib/test.bzl
+++ b/contrib/test.bzl
@@ -167,6 +167,7 @@ def container_test(name, image, configs, driver = None, verbose = None, **kwargs
             images = {
                 loaded_name: image,
             },
+            **kwargs
         )
 
     _container_test(
--- a/contrib/test.bzl
+++ b/contrib/test.bzl
@@ -167,7 +167,7 @@ def container_test(name, image, configs, driver = None, verbose = None, **kwargs
             images = {
                 loaded_name: image,
             },
-            **kwargs
+            testonly = True,
         )
 
     _container_test(
--- a/contrib/test.bzl
+++ b/contrib/test.bzl
@@ -160,7 +160,7 @@ def container_test(name, image, configs, driver = None, verbose = None, **kwargs
         # Include the package name in the new image tag to avoid conflicts on naming
         # when running multiple container_test on images with the same target name
         # from different packages.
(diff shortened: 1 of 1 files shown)


contrib/test.bzl
```diff
@@ -160,13 +160,14 @@ def container_test(name, image, configs, driver = None, verbose = None, **kwargs
         # Include the package name in the new image tag to avoid conflicts on naming
         # when running multiple container_test on images with the same target name
         # from different packages.
-        sanitized_name = (native.package_name() + image).replace(":", "").replace("@", "").replace("/", "")
+        sanitized_name = (native.package_name() + image).replace(":", "").replace("@", "").replace("/", "").lower()
```
nlopezgi: why the need to lowe case this?

nlopezgi:
friendly ping @virusdave, let me know about the question I asked so we can get this in (or preferably remove the change to the lower case and if you're interested in that we can get it in a different PR)

contrib/test.bzl
virusdave: Ah, because it gets used in a container's name i believe and uppercase characters are not allowed there.  So if you have a `container_test(name = "badPractices")`, it wouldn't work as-is.

virusdave:
> friendly ping @virusdave, let me know about the question I asked so we can get this in (or preferably remove the change to the lower case and if you're interested in that we can get it in a different PR)

Ah, because it gets used in a container's name i believe and uppercase characters are not allowed there. So if you have a container_test(name = "badPractices"), it wouldn't work as-is.

nlopezgi:
> > friendly ping @virusdave, let me know about the question I asked so we can get this in (or preferably remove the change to the lower case and if you're interested in that we can get it in a different PR)
> 
> Ah, because it gets used in a container's name i believe and uppercase characters are not allowed there. So if you have a container_test(name = "badPractices"), it wouldn't work as-is.

Can you please remove that change. Changing the name to lowercase could be problematic for some use cases. If you want we can talk about that in a separate issue/PR

nlopezgi:
friendly ping @virusdave , can you remove the change to the lowecase and we can get this change in, otherwise we can close this?

PR #1223, opened 2019-10-17
Support compressed tarballs in the loader

siddharthab:
When the loader binary is updated, I can send in another PR that will
pipe this to container_load.


siddharthab:
Also submitted https://github.com/google/go-containerregistry/pull/578.

This PR will not be needed if that gets accepted.

container/go/cmd/loader/loader.go
```diff
@@ -31,11 +33,24 @@ import (
 var (
 	outDir     = flag.String("directory", "", "Where to save the image's files.")
 	imgTarball = flag.String("tarball", "", "The path to the tarball to load.")
+	isGzip     = flag.Bool("gzip", false, "The tarball is gzip compressed.")
```
smukherj1: Recommend detecting this automatically using [IsGzipped](https://github.com/google/go-containerregistry/blob/093adf2740a9617fbb6e6fe81ec37b00d3eb1b43/pkg/v1/v1util/zip.go#L82). This way if we want to support more compressed formats in the future, we won't have to add a flag + attribute for each of them.

siddharthab:
> Thanks a lot for your PR & adding tests.
Thank you for all the work on this project, specially the migration to go tools. When we initially wrote container_load using the python based tools, we did not support compressed tarballs because of efficiency reasons. Uncompressing every time to read each file was much slower, and so it made sense to require users to uncompress once to pass to the tools.

I will wait for some response on the go-containerregistry and then make updates here accordingly.

nlopezgi:
any updates on this PR? is it still needed / being worked on?

siddharthab:
The PR on go-containerregistry was not accepted. So we can continue with the review of this PR.

zoidbergwill:
It looks like there are some unresolved comments that need to be addressed?

container/go/cmd/loader/loader.go
siddharthab: Done. That method is now moved to an internal package so can't be used directly from there. So I have added a new method to this PR which is similar.

siddharthab:
> It looks like there are some unresolved comments that need to be addressed?

I think all comments are now resolved.

Issue #1207, opened 2019-10-11
Unable to set DOCKER_HOST and DOCKER_TLS_VERIFY properties with docker toolchain


bajacondor:
`rules_docker` with the docker toolchain appears unable to connect to a remote docker daemon   Setting a custom config seems to have no effect for these properties.

nlopezgi:
hi @bajacondor,
I don't think we have implemented features to connect to a remote docker host (to be honest I'm not familiar atm with use of these env vars). We would be happy to work with you if you want to contribute this feature.

bajacondor:
Hi @nlopezgi, Thanks for your reply.  I was able to hack the `toolchain_configure` rules `docker_path` argument, and I set all the remote host and tls values as flags to the docker command.  This worked up until `install_pkgs()` specifically `https://github.com/bazelbuild/rules_docker/blob/master/docker/package_managers/run_install.sh.tpl#L21` where there are some volume mounts going on.  In the case of a remote docker daemon, it seems the local side of the volume mount is the host machine.

I would be happy to contribute to this project.  I've been using `rules_docker` to create a library of procedurally built container images for use in CI/CD deployments at my job with Nordstrom.  You're welcome to dm me anytime to work out a plan for what I can contribute.

I must say that my development experience with Bazel is somewhat limited, but I'm getting the hang of reading and understanding rules and ctx's and whatnot, so I think I could be of some value to this project.

Heres a question:   In `download_pkgs()`  The template uses `docker run` with the expanded commands concated to the end of the command.  in `install_pkgs()` you volume mount the script instead.  Why is that?

Thank you for your involvement with this project!

David

bajacondor:
Here is another thing to note.  The `install_pkgs()` rule ends up running a container with a `-v` (volume mount) like this:
`-v /root/.cache/bazel/_bazel_root/2deb5d73cb7bfe00ce415d97f6b6c0a6/sandbox/linux-sandbox/27/execroot/__main__/bazel-out/k8-fastbuild/bin/ca-certificates.tar:/tmp/bazel-out/k8-fastbuild/bin/ca-certificates.tar` 
Even if your client and dockerd images can share a context, it's unlikely that they share the rootfs.  Is there a way to tell Bazel to use a different dir for files it wants to eventually mount on a container?

nlopezgi:
[...]

PR #1228, opened 2019-10-18
Remote docker host

bajacondor:
This allows flags like `-H=some.other.docker.host:2375` to be set with `docker_path` in  docker `toolchain_configure()` so that rules_docker works with a remote docker host.

Related issues:
https://github.com/bazelbuild/rules_docker/issues/1207

Regarding the CLA for d****@nordstrom.com:
I attest that I approve the CLA for both my personal email and the above mentioned nordstrom.com email.  Both addresses are mine.

Thank you for your consideration


bajacondor:
@googlebot I fixed it.

bajacondor:
Can someone tell me which email address is causing the `cla:no` flag?  All my commits have the email address that has an individual cla attached.  I'm not sure how to fix this.

Thank you.
David

smukherj1:
Hi @bajacondor,

Thanks for the PR. Regarding the CLA, the bot found the following commit authors:
1. david.v.schile@nordstrom.com (No CLA)
2. bajacondor@gmail.com (Has CLA)

bajacondor:
@smukherj1,
That is so annoying.  Here is my situation:
Both the gmail and the nordstrom addresses are mine.  I have a personal cla for the gmail address.  The nordstrom address is tied to another google  account so I can't add it to my personal alternate emails.  When I attempt to create a CLA for the nordstrom account, I get directed and can't sign in.  Who can I speak to about this?   I tried rewriting the history to change email addresses, and somehow the bot still knows about the n.com address!

nlopezgi:
thanks for your contributions @bajacondor , given the situation, you can just confirm here in the text that you agree with the CLA as both **@nordstrom.com and your other account and we can manually skip that check.

nlopezgi:
please fix the issues with the buildkite presubmit and let me know when this is ready for review

bajacondor:
I changed directions on this and the argument for flags to the `toolchain_configure()` rule.  I think I've got all the presubmit and buildifier tests corrected too, but I don't work at Google, so I'm not totally sure.  Let me know if I've missed something.

docker/util/image_util.sh
```diff
@@ -4,16 +4,17 @@ reset_cmd() {
     local original_image_name=$1
     local container_id=$2
     local output_image_name=$3
+    local docker_command=$4
```
nlopezgi: it seems to me you need to change the call to reset_cmd in https://github.com/bazelbuild/rules_docker/blob/master/docker/package_managers/run_install.sh.tpl#L25 to pass the expected docker_command?

bajacondor: I did that.  Its here: https://github.com/bazelbuild/rules_docker/blob/ba0795d0fda5dcdd68107f37c0355145c10fd24e/docker/package_managers/run_install.sh.tpl#L26  
Am I missing something?

bajacondor: RE Tests:
Can I get some more clarification on this?  It appears the tests you have are running from this script: https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e.sh  And they are running targets that are actually in the `//tests/` directory.  You want me to put my target in the BUILD file next to my WORKSPACE in `//testing/custom_toolchain_flags`?  

Also, How do I run `file_test()` against bazel files like the one that ends up here: `bazel-out/darwin-fastbuild/bin/base_image.build`?

Thank you!

nlopezgi: > I did that. Its here:
Thanks for the clarification, not sure why I missed it.




> RE Tests:

So our test setup is quite complex. Lets try to only explain the bits that you actually have to move.
- You need to create a new directory under testing (similar to custom_toolchain_flags, but not the same one, call it custom_docker_flags)
- Directory has to have its own WORKSPACE and BUILD file to test your new features
- Tests will be run via GCB using a cloudbuild.yaml similar to this one: https://github.com/bazelbuild/rules_docker/blob/master/testing/download_pkgs_at_root/cloudbuild.yaml (and its basically as simple as the one pointed to here, just it lives under a different directory and has a different `dir` attr)
- You can test locally simply by running `bazel test ...` from this directory
- I will be doing the GCB setup (triggers), once the cloudbuild.yaml file is created in your PR

> Also, How do I run file_test() against bazel files like the one that ends up here: bazel-out/darwin-fastbuild/bin/base_image.build?

for this file the corresponding target should be `"//:base_image.build"`

Let me know if you have more questions. Thanks for your contributions!

bajacondor: Excellent.  Thank you for the clarification and your patience.  Another quick question.  I see in `toolchain_configure()` that the `DOCKER_CONFIG` environment variable is used if set.  My question is: How is the rule getting the value of that env var?  According to the docs, the rule must specify that with the `environ` attr of the rule, but I don't see that happening.

bajacondor: More clarification.. When you say:

> for this file the corresponding target should be "//:base_image.build"

This doesn't seem to be an output of `install_pkgs()`  it is generated but the final `ctx.actions.run()` only has `output_tar` as an output.  So running a test like: ```file_test(
    name = "test_custom_flag_install_pkgs_build",
    file = ":test_install_pkgs.build",
    regexp = "DOCKER_FLAGS=\"--log-level=info\"",
)``` returns errors like : ```//:test_custom_flag_install_pkgs_build: missing input file '//:test_install_pkgs.build'```

bajacondor:
I wrote a few tests, but I'm stumped on a couple. (`container/incremental_load.sh.tpl` and `docker/util/commit.sh.tpl` (because those files don't seem to be available to the `file_test()` rule since they aren't specified as outputs of any rule.  Let me know if I'm doing something wrong here.

bajacondor:
I don't think my current tests are running from their own WORKSPACE location.  I'm not sure what is even kicking off the buildkite tests.  I'm not familiar with this pipeline.

Thank you for helping me through this.

docker/util/image_util.sh
nlopezgi: > This doesn't seem to be an output of `install_pkgs()` it is generated but the final `ctx.actions.run()` only has `output_tar` as an output.

Yes, I just noticed this is not an explicit output so you cant really refer to it. You will need to make it an explicit output of `install_pkgs` by adding it here https://github.com/bazelbuild/rules_docker/blob/master/docker/package_managers/install_pkgs.bzl#L187 and changing the assignment of script [here](https://github.com/bazelbuild/rules_docker/blob/master/docker/package_managers/install_pkgs.bzl#L88) to point to your new output

nlopezgi:
> I wrote a few tests, but I'm stumped on a couple. (`container/incremental_load.sh.tpl` and `docker/util/commit.sh.tpl`) Those files don't seem to be available to the `file_test()` rule since they aren't specified as outputs of any rule. Let me know if I'm doing something wrong here.

If they are not available for file_test then they need to be made explicit outputs of the rules (sorry for the additional hassle, testing features in starlark is not easy)

nlopezgi:
> I don't think my current tests are running from their own WORKSPACE location. I'm not sure what is even kicking off the buildkite tests. I'm not familiar with this pipeline.
> 
> Thank you for helping me through this.

They are most definitely not running on buildkite (buildkite is not setup to run these tests). These will run with GCB, and I will set up the trigger, once you get them passing locally and we are happy with the coverage they provide, in the meantime, no need to worry about that.

nlopezgi:
And thanks again for your contributions!

bajacondor:
> They are most definitely not running on buildkite (buildkite is not setup to run these tests). These will run with GCB, and I will set up the trigger, once you get them passing locally and we are happy with the coverage they provide

According to buildkite output here:  https://buildkite.com/bazel/rules-docker-docker/builds/3241#44edcadb-07e9-4a38-88d1-3f5ccd63dbbf/260-365 they are running currently.  How are they being initiated?

> iirc, the DOCKER_CONFIG env variable is used directly by the docker binary without needing to read it explicitly. 

Thats correct.  DOCKER_CONFIG gets picked up by the go code.

> If they are not available for file_test then they need to be made explicit outputs of the rules (sorry for the additional hassle, testing features in starlark is not easy)

No kidding.  I'm working on it.  In the case of `container/incremental_load.sh.tpl` its made more difficult because the function that creates that template is not a rule,  It's a support function to other rules like `container_image()`  The script however, is in the actual rule name as the executable.  Is there a way to test the executable?

nlopezgi:
> According to buildkite output here: https://buildkite.com/bazel/rules-docker-docker/builds/3241#44edcadb-07e9-4a38-88d1-3f5ccd63dbbf/260-365 they are running currently. How are they being initiated?

ugh, these should not run on buildkite, you need to add the directory to https://github.com/bazelbuild/rules_docker/blob/master/.bazelignore



> No kidding. I'm working on it. In the case of `container/incremental_load.sh.tpl` its made more difficult because the function that creates that template is not a rule, It's a support function to other rules like `container_image()` The script however, is in the actual rule name as the executable. Is there a way to test the executable?

the executable should be the main output of the rule (i.e., should be the file corresponding to the name of the target). You can also use bazel query to list the outputs: https://docs.bazel.build/versions/master/query-how-to.html#What_files_are_generated_by_rule

bajacondor:
> the executable should be the main output of the rule (i.e., should be the file corresponding to the name of the target). 

Thats what got me stumped, I can't seem to test that as a target!

> You can also use bazel query to list the outputs:

In the case of a `container_image()` rule target named `test_container_image`, `bazel query 'kind("generated file". //:*)'` gives me:

```//:test_container_image_rule.sh
//:test_container_image.tar
//:test_container_image.json
//:test_container_image.digest
//:test_container_image-layer.tar```

The executable should be called `test_container_image`  It's there in bazel-out/... but not available as a test target.  Seems like tests should be be able to check this file for content.

**Comment by nlopezgi** [comment]
> > the executable should be the main output of the rule (i.e., should be the file corresponding to the name of the target).
> 
> Thats what got me stumped, I can't seem to test that as a target!
> 
> > You can also use bazel query to list the outputs:
> 
> In the case of a `container_image()` rule target named `test_container_image`, `bazel query 'kind("generated file". //:*)'` gives me:
> 
> ```shell
> //:test_container_image.tar
> //:test_container_image.json
> //:test_container_image.digest
> //:test_container_image-layer.tar
> ```
> 
> The executable should be called `test_container_image` It's there in bazel-out/... but not available as a test target. Seems like tests should be be able to check this file for content.

Ugh, Bazel rules are hard to test, now that I think about it more, even though it sounds stupid, if you want to test the contents of the executable, i think, you have to make it an additional explicit output of the rule (call it `.executable`). Sorry again for all the hoops to jump needed to get this tested.

**Comment by bajacondor** [comment]
I'm attempting to solve the executable file testing that comes from the `incremental_load.tpl` in `layer_tools.bzl`  with a symlink to the `ctx.output.executable` from `container_image()`  This seems to work for me locally.  Now I'm getting buildkite fails in what seem to be totally unrelated tests: https://buildkite.com/bazel/rules-docker-docker/builds/3247#4c182525-61bd-43ad-a446-753ab7bbe938/1997-2241  These seem to have to do with the `--host_force_python=PY2` flag.  I still don't see where these tests are being initiated.   

Thanks for any input you can provide with this.

**Comment by nlopezgi** [comment]
> I'm attempting to solve the executable file testing that comes from the `incremental_load.tpl` in `layer_tools.bzl` with a symlink to the `ctx.output.executable` from `container_image()` This seems to work for me locally. Now I'm getting buildkite fails in what seem to be totally unrelated tests: https://buildkite.com/bazel/rules-docker-docker/builds/3247#4c182525-61bd-43ad-a446-753ab7bbe938/1997-2241 These seem to have to do with the `--host_force_python=PY2` flag. I still don't see where these tests are being initiated.
> 
> Thanks for any input you can provide with this.

The key message in your Buildkite run is this one:

```
(03:00:28) ERROR: file 'tests/docker/package_managers/ubuntu_gpg_image.executable' is generated by these conflicting actions:
--
  | Label: //tests/docker/package_managers:ubuntu_gpg_image
  | RuleClass: add_apt_key rule
  | Configuration: 6b53e59eacbd7840dbd2abe35d79ee9c
  | Mnemonic: Action
  | Action key: 6387cc75dd010878b670d6084bad7192, 34c96bfeae339eb830d16c9b33e41a35
  | Progress message: Action tests/docker/package_managers/ubuntu_gpg_image.executable
```
This means that you have two actions that are producing an output with the same name.
Specifically the two actions are:
1. https://github.com/bazelbuild/rules_docker/blob/master/docker/package_managers/apt_key.bzl#L71

2. https://github.com/bazelbuild/rules_docker/blob/master/docker/package_managers/apt_key.bzl#L105

Unfortunately, to get this fixed, you will have to pipe down the name of the output executable, so it can be overridden. See for instance how `output_digest` is passed with different values in the two calls above and how its handled in [image.bzl](https://github.com/bazelbuild/rules_docker/blob/master/container/image.bzl#L339). You might also likely have to pass the executable in the container_parts here: https://github.com/bazelbuild/rules_docker/blob/master/container/image.bzl#L433

 Sorry for all the complexity, Starlark is hard, but you'll be close to an expert by the time this PR is done!

**Comment by bajacondor** [comment]
Thanks for the coaching!  I will try this.

Quick question: For tests like `container_test()` that are run from .yaml specs, is there a way to get more verbose output from bazel about the individual tests defined in the yaml?  Bazel seems to turn everything in the yaml as one test with no information about whats being tested in the yaml.
Thanks again!

**Comment by nlopezgi** [comment]
> Thanks for the coaching! I will try this.
> 
> Quick question: For tests like `container_test()` that are run from .yaml specs, is there a way to get more verbose output from bazel about the individual tests defined in the yaml? Bazel seems to turn everything in the yaml as one test with no information about whats being tested in the yaml.
> Thanks again!

I think there should be a test log that has some details of the test execution, but yes, each container_test target as a whole is a single pass/fail. You can always split up the .yaml into several of them if its too big and do several container_test targets. Does this answer your question?

**Comment by bajacondor** [comment]
> Does this answer your question?

Yes, Thank you!

**Comment by bajacondor** [comment]
Making my way through this. :| 
Tests are failing at `//tests/docker/package_managers:install_pkgs_reproducibility_test` yet they pass locally when I add the `--host_force_python=PY2` flag.  I'm not sure what's happening with that.  I don't see how my changes would have affected a python script. Any insight?  Thanks again, @nlopezgi

**Comment by bajacondor** [comment]
Hello!  @nlopezgi,
It seems this has stalled.  Can I get some input?  Thank you!

**Comment by smukherj1** [comment]
Hi @bajacondor,

Could you please print the output you get when you run `bazel test --test_output=all //tests/docker/package_managers:install_pkgs_reproducibility_test`? This might help us figure out where the error is coming from. It's possible the tests added in this PR are running a new python script that was missed out when we migrated all our python scripts from python 2 to python 3.

**Comment by nlopezgi** [comment]
> Making my way through this. :|
> Tests are failing at `//tests/docker/package_managers:install_pkgs_reproducibility_test` yet they pass locally when I add the `--host_force_python=PY2` flag. I'm not sure what's happening with that. I don't see how my changes would have affected a python script. Any insight? Thanks again, @nlopezgi

Sorry for the super long delay getting back. These issues are hard to debug and I knew this would actually require me to pull your PR and debug myself, which I was delaying as much as I could.

The problem you have now is that you have multiple outputs from an image target which are being interpreted by `compare_ids_test.bzl` as all being tar files.
The simplest solution is to add this code:
```
    tar_files = []
    for file in ctx.files.images:
        if file.short_path.endswith("tar"):
            tar_files += [file]
```
instead of the line [here](https://github.com/bazelbuild/rules_docker/blob/master/contrib/compare_ids_test.bzl#L43)

**Comment by bajacondor** [comment]
@nlopezgi , Thanks again for the guidance.  

> The problem you have now is that you have multiple outputs from an image target

Corrected the issue as you suggested.  Seem to have passing tests now.

File: testing/custom_toolchain_flags/cloudbuild.yaml [hunk 2]
```diff
@@ -0,0 +1,22 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# Tests that verify that download_pkgs targets can be declared
+# in the top level BUILD file.
+
+steps:
+  # Test the download_pkgs examples.
+  - name: "l.gcr.io/google/bazel"
+    args: ["test", "--host_force_python=PY2", "//..."]
```
nlopezgi: can you remove the --host_force_python=PY2 flag here?

docker/package_managers/install_pkgs.bzl
```diff
@@ -93,7 +93,7 @@ def _impl(ctx, image_tar = None, installables_tar = None, installation_cleanup_c
         template = ctx.file._image_util_tpl,
         output = image_util,
         substitutions = {
-            "%{docker_tool_path}": toolchain_info.tool_path,
+            "%{docker_tool_path}": "%s %s" % (toolchain_info.tool_path, " ".join(toolchain_info.docker_flags)),
```
nlopezgi: Dont join here, make sure you sub in the template the two things separately.
```
"%{docker_tool_path}": toolchain_info.tool_path,
"%{docker_flags}": " ".join(toolchain_info.docker_flags),

```

testing/e2e.sh
```diff
@@ -59,7 +59,7 @@ EOF
   EXPECT_CONTAINS "$(cat output.txt)" "Second: 5"
   EXPECT_CONTAINS "$(cat output.txt)" "Third: 6"
   EXPECT_CONTAINS "$(cat output.txt)" "Fourth: 7"
-  EXPECT_CONTAINS "$(cat bazel-bin/testdata/py3_image_with_custom_run_flags)" "-i --rm --network=host -e ABC=ABC"
+  EXPECT_CONTAINS "$(cat bazel-bin/testdata/py3_image_with_custom_run_flags.executable)" "-i --rm --network=host -e ABC=ABC"
```
nlopezgi: Does this fail w/o the change?
i.e., this should not be necessary, as the default executable should still have these same contents.

bajacondor: Yes, it still seems to fail.  I think because `.executable` has become the new default executable?

nlopezgi: so what are the contents of `bazel-bin/testdata/py3_image_with_custom_run_flags` after this change?
What happens if you run `bazel run //testdata:py3_image_with_custom_run_flags`? Does it work as expected or does it fail?

nlopezgi:
/testing/e2e/test_docker_run_flags.yaml
/testing/e2e/test_bazel_run_docker.yaml
/testing/e2e/test_java_image.yaml
are failing (I suspect the check they are making is looking for the old executable, not the new one)
Your new tests (/testing/custom_toolchain_flags/cloudbuild.yaml) are failing with the following error message:
```
ERROR: /workspace/testing/custom_toolchain_flags/BUILD:36:1: Action test_install_pkgs.tar.unstripped failed (Exit 126) test_install_pkgs.build failed: error executing command bazel-out/k8-fastbuild/bin/test_install_pkgs.build

Use --sandbox_debug to see verbose messages from the sandbox
+ DOCKER=/usr/bin/docker
+ DOCKER_FLAGS=--log-level=info
+ [[ -z /usr/bin/docker ]]
+ TO_JSON_TOOL=bazel-out/host/bin/external/io_bazel_rules_docker/docker/util/to_json
+ source bazel-out/k8-fastbuild/bin/image_util.sh
++ bazel-out/host/bin/external/io_bazel_rules_docker/contrib/extract_image_id bazel-out/k8-fastbuild/bin/external/debian_base/image/image.tar
+ image_id=6786f6a45544215e9e4411db2318c7f561009b38fcf7c88f2f5f656256ba72f8
+ /usr/bin/docker --log-level=info load -i bazel-out/k8-fastbuild/bin/external/debian_base/image/image.tar
+++ pwd
+++ pwd
++ /usr/bin/docker --log-level=info run -d -v /builder/home/.cache/bazel/_bazel_root/0812a39b4bd30217bd2cd31fb98be2ff/sandbox/linux-sandbox/24/execroot/e2e_testing/bazel-out/k8-fastbuild/bin/test_download_pkgs.tar:/tmp/bazel-out/k8-fastbuild/bin/test_download_pkgs.tar -v /builder/home/.cache/bazel/_bazel_root/0812a39b4bd30217bd2cd31fb98be2ff/sandbox/linux-sandbox/24/execroot/e2e_testing/bazel-out/k8-fastbuild/bin/test_install_pkgs.install:/tmp/installer.sh --privileged 6786f6a45544215e9e4411db2318c7f561009b38fcf7c88f2f5f656256ba72f8 /tmp/installer.sh
/usr/bin/docker: Error response from daemon: OCI runtime create failed: container_linux.go:345: starting container process caused "exec: \"/tmp/installer.sh\": permission denied": unknown.
+ cid=830dd18a8bcfaab3522d0b6e1d6858efed0134b2ad7d7a36db39c4543e831f2b
```

bajacondor:
> Your new tests (/testing/custom_toolchain_flags/cloudbuild.yaml) are failing

Ok, this is a sad side-effect of  a remote docker.  The problem is that because the install script is volume mounted, when the daemon is remote it's trying to volume mount it from the remote machine where it does not exist.  I see a couple solutions: 
1. we could run the script from the docker command line
2. we could `docker cp` the script to the container and then run it.
I'll look into solutions and add to this MR.  
I'm glad the tests caught this! :/

nlopezgi:
> > Your new tests (/testing/custom_toolchain_flags/cloudbuild.yaml) are failing
> 
> Ok, this is a sad side-effect of a remote docker. The problem is that because the install script is volume mounted, when the daemon is remote it's trying to volume mount it from the remote machine where it does not exist. I see a couple solutions:
> 
> 1. we could run the script from the docker command line
> 2. we could `docker cp` the script to the container and then run it.
> 
> I'll look into solutions and add to this MR.
> I'm glad the tests caught this! :/

hmm, these docker-in-docker issues are fun. So, generally, what we do to resolve them, is control Bazel's output_base so that it maps to an absolute path that is identical in the host env and the docker container running Bazel. See for example https://github.com/bazelbuild/rules_docker/blob/master/testing/examples/run_instruction_apt_pkgs/cloudbuild.yaml#L48
(note you should not need to copy rules docker to a sub directory as in step 0 above, you should be fine with just setting an output_base)

emranbm:
I have the same issue with `install_pkgs`. Seems this PR will resolve that.
Thanks a lot. But what's the status now?

nlopezgi:
friendly ping, any plans on finishing this PR?

bajacondor:
> friendly ping, any plans on finishing this PR?

Ah yes.  I finally have a little time for it.  Sorry for the absence. I'll pick this up and see if I can get it merged first.

bajacondor:
@nlopezgi, Strugging a little with this: 
```text
ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/docker_config/BUILD:21:1: in docker_toolchain rule @docker_config//:toolchain:
	File "/var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/docker_config/BUILD", line 21
		docker_toolchain(name = 'toolchain')
	File "/workdir/toolchains/docker/toolchain.bzl", line 38, in _docker_toolchain_impl
		platform_common.ToolchainInfo(<1 more arguments>)
	File "/workdir/toolchains/docker/toolchain.bzl", line 39, in platform_common.ToolchainInfo
		DockerToolchainInfo(<7 more arguments>)
duplicate keyword 'client_config' in call to DockerToolchainInfo
```
I don't see where the dupe of client_config is.

toolchains/docker/toolchain.bzl
```diff
@@ -36,10 +37,12 @@ DockerToolchainInfo = provider(
 def _docker_toolchain_impl(ctx):
     toolchain_info = platform_common.ToolchainInfo(
         info = DockerToolchainInfo(
+            docker_flags = ctx.attr.docker_flags,
             client_config = ctx.attr.client_config,
             gzip_path = ctx.attr.gzip_path,
             gzip_target = ctx.attr.gzip_target,
             tool_path = ctx.attr.tool_path,
+            client_config = ctx.attr.client_config,
```
nlopezgi: this is duplicated from line 41

bajacondor: > this is duplicated from line 41

derp.  I was having some merge issues.  Turns out my buildifier command was screwing up files.  I was using `buildifier --lint=fix --warnings=all -r ./` and it did some weird things.  What does your local buildifier command look like?

nlopezgi:
I run
`buildifier --lint=fix $(find . -name BUILD -o -name WORKSPACE -o -name "*.bzl" -type f)`
which fixes some issues automatically and then 
`buildifier --lint=warn $(find . -name BUILD -o -name WORKSPACE -o -name "*.bzl" -type f)`
to find out any other issues I need to fix manually

nlopezgi:
travis tests are failing, can you check out what's up and let me know if you need a hand fixing?

bajacondor:
> travis tests are failing, can you check out what's up and let me know if you need a hand fixing?

Got a couple green checks there!  What next?

nlopezgi:
> > travis tests are failing, can you check out what's up and let me know if you need a hand fixing?
> 
> Got a couple green checks there! What next?

Awesome, kicked off gcb tests. I suspect some will fail due to changes in output names, so should be a matter of debugging those and then we should be good to go.

nlopezgi:
https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e/test_docker_run_flags.yaml is failing with the following error message:
```
Step #0: + echo 'FAILURE: Expected '\''-i --rm --network=host bazel/testdata:docker_run_flags_use_default'\'' not found in '\'''\'''
Step #0: FAILURE: Expected '-i --rm --network=host bazel/testdata:docker_run_flags_use_default' not found in ''
Step #0: + exit 1
```
related to [this](https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e/docker_run_flags.sh#L28) line


https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e/test_bazel_run_docker.yaml is failing with  the following error  message:
```
Step #6: + ./bazel-bin///testdata/workdir_with_tar_base
Step #6: ./testing/e2e/bazel_run_docker.sh: line 95: ./bazel-bin///testdata/workdir_with_tar_base: No such file or directory
```
related to [this](https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e/bazel_run_docker.sh#L95) line

https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e/test_java_image.yaml is failing with  the following error  message: 
```
Step #0: + grep -Fsq -- '-i --rm --network=host -e ABC=ABC'
Step #0: FAILURE: Expected '-i --rm --network=host -e ABC=ABC' not found in ''
Step #0: + fail 'Expected '\''-i --rm --network=host -e ABC=ABC'\'' not found in '\'''\'''
Step #0: + echo 'FAILURE: Expected '\''-i --rm --network=host -e ABC=ABC'\'' not found in '\'''\'''
Step #0: + exit 1
```
related to [this](https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e/java_image.sh#L33) line

nlopezgi:
@bajacondor we still have one test failing with message:
```
INFO: Build completed successfully, 1 total action
+ ./bazel-bin///testdata/workdir_with_tar_base
./testing/e2e/bazel_run_docker.sh: line 95: ./bazel-bin///testdata/workdir_with_tar_base: No such file or directory
```
corresponding to an error in [this](https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e/bazel_run_docker.sh#L95) line

bajacondor:
Hi @nlopezgi, Thanks for all your help with this.  I'm curious why it's on hold to merge?

nlopezgi:
> Hi @nlopezgi, Thanks for all your help with this. I'm curious why it's on hold to merge?

Because some tests have failed

bajacondor:
> Because some tests have failed

Ah, I'm unable to see the cloudbuild test results.  Is there anything I can do?

nlopezgi:
> > Because some tests have failed
> 
> Ah, I'm unable to see the cloudbuild test results. Is there anything I can do?

Did you fix the one I commented above?
`  testing/e2e/bazel_run_docker.sh: line 95: ./bazel-bin///testdata/workdir_with_tar_base`

bajacondor:
> Did you fix the one I commented above?

Ah, no. I  misunderstood that function and the loop.  I think I fixed it with this commit.

bajacondor:
Looks like one more failing on Cloud Build. Is there a way to get access to these test results for my google user?

nlopezgi:
Thanks a lot for all the work on this PR!

smparkes:
So I've been exploring `docker_flags` because when run in CI/CD, we use a remote docker.

We need host, tls, and the various PKI files.

The values are provided through env variables but getting those into bazel is tricky.

Critically, pretty much all the ways of getting the values into bazel are going to bust the cache if they change, which they generally will for us on every run. This is a show stopper.

I can see a couple of ways forward and was wondering what people think.

They both revolve around putting the necessary info in `$HOME/.docker`. Unfortunately, these values aren't part of `config.json`. If you use `--tls`, docker will look here for the PKI files but it doesn't use that directory for `host`, `tls` or `tlsverify`.

So I'm thinking, outside of bazel we could copy the env variables into that directory and then tweak the `rules_docker` scripts to use those values if those files exist. It's painful and has problems with shared home directories if people were trying to talk to multiple remote dockers but that isn't immediately a blocker for us. It's moderately straightforward.

Alternatively, rather than tweaking the `rules_docker` scripts, we can create a wrapper around the docker client that use the values to set flags/env variables. The painful part there is having to set an absolute path in the toolchain that must be the same on every host we ever run this on.

Just wondering if I'm missing anything or what people thought in general.

Also, FWIW, I think this line in the document might be a little misleading/confusing:

> When left unspecified (ie not set explicitly or set by the docker toolchain), docker will use the directory specified via the DOCKER_CONFIG environment variable. If DOCKER_CONFIG isn't set, docker falls back to $HOME/.docker.

Without an `--action_env` setting, I don't think `DOCKER_CONFIG` will be propagated from the shell invoking `bazel` to the docker client so won't have any impact (it'll still look in `$HOME/.docker`.)

Jonpez2:
Hello!  I've been trying to use docker_flags with container_test as well - is there a particular reason why the flags aren't piped through from the toolchain to the docker call in container_test?  That seems like the right thing to do, no?

Thank you!

Issue #300, opened 2018-01-31
support docker v2.2 manifest lists for multi-arch images


ixdy:
Consider this a general brainstorming issue, since I'm not really sure how exactly this would work, and I think most bazel rules are missing necessary support for this anyway.

But as motivating example: the `go_binary` rule from bazelbuild/rules_go lets you set `goos` and `goarch` attributes, so you can theoretically create multiple copies of a `go_binary` with different architecture settings and build all of them in a single `bazel build`.

It then isn't a far stretch to imaging wrapping these into appropriate base images and then bundling them into a single manifest list to push to your docker registry of choice.

Existing (non-Bazel) work around manifest lists includes https://github.com/estesp/manifest-tool.

mattmoor:
Having the capacity to produce fat images via Bazel is something I have long considered a goal of these rules, but for a long time Bazel has lacked the capacity for even basic cross compilation.

If I can simultaneously build two `go_binary` targets for distinct platforms, that's fantastic progress, but I'm not sure the Go rules alone supporting this would be enough.

I'd love a consistent story for producing these fat images across languages, so that we could write something like:
```python
fat_image(
   name = "fat",
   image = "//path/to:image",  # This could be anything
   platforms = [
      ...
   ],
)
```

... and under the hood, we could synthesize a target that stitches together the image for each platform, the dependency on which is expressed by something like:
```python
  "//path/to:image#--options"
```
*I recall seeing something like that in a Bazel config doc once upon a time (re: generalized fat binary production), but have no idea where it went.*

It is notable that the underlying library fully supports fat images, so the primary reason this doesn't exist is the lack of a good story for expressing it in Bazel.

@davidstanke FYI

@dlorenc @r2d4 FYI, since we'd probably want decent distroless image coverage (but not required)

chrislovecnm:
👍

ixdy:
per https://github.com/bazelbuild/bazel/issues/5575#issuecomment-458351171, this might be implementable now.

gregestren:
I'd be happy to chat feasibility with an appropriate rule owner expert.

ixdy:
cc @nlopezgi @xingao267

nlopezgi:
@ixdy thanks for creating this issue,
[...]

PR #1236, opened 2019-10-21
Add new adopter

nrvnrvn:
For the benefit of https://github.com/bazelbuild/rules_docker/issues/300#issuecomment-520451066


nrvnrvn:
@smukherj1 done!

Issue #1247, opened 2019-10-25
py_image/py3_image entrypoint conflicts with a custom Python container toolchain


mmikitka:
I am using Python's official `python:3.7.3-slim` image from Dockerhub. Here is my pull command:
```
[container_pull(
    name = name,
    registry = "index.docker.io",
    repository = "library/python",
    digest = "sha256:64d8fdeff90572068f5fb5e7fa0d67e28fa4629f54682a67156c6d6fbf3125b9",
) for name in [
    "py3_image_base",
    "py3_debug_image_base",
]]
```

As a result, I have a custom Python toolchain for the container execution target:
```
# Modified from @rules_docker//toolchains:BUILD
py_runtime(
    name = "container_py2_runtime",
    interpreter_path = "/usr/bin/python",
    python_version = "PY2",
)

# Path in the python:slim-3.7 image
py_runtime(
    name = "container_py3_runtime",
    interpreter_path = "/usr/local/bin/python3",
    python_version = "PY3",
)

py_runtime_pair(
    name = "container_py_runtime_pair",
    py2_runtime = ":container_py2_runtime",
    py3_runtime = ":container_py3_runtime",
)

toolchain(
    name = "container_py_toolchain",
    exec_compatible_with = [
        "@io_bazel_rules_docker//platforms:run_in_container",
    ],
    toolchain = ":container_py_runtime_pair",
    toolchain_type = "@bazel_tools//tools/python:toolchain_type",
)
```

I then register the toolchain and execution platforms as such (copied and altered from rules_docker/python3/image.bzl):

```
register_toolchains(
    "//toolchain:container_py_toolchain",
)

register_execution_platforms(
    "@local_config_platform//:host",
    "@io_bazel_rules_docker//platforms:local_container_platform",
)
```

My py3_image rule looks like this
```
DEPS = [
    requirement("certifi"),
    requirement("chardet"),
    requirement("idna"),
    requirement("prometheus_client"),
    requirement("requests"),
    requirement("urllib3"),
]

py3_image(
    name = "image",
    srcs = ["probe.py"],
    layers = DEPS + [
        ":probe",
    ],
    main = "probe.py",
)
```

I then build the image:
```
bazel run --toolchain_resolution_debug //insights/monitor/probe/pipeline:image
```
and I get this
```
[...]

PR #1248, opened 2019-10-25
Remove python entrypoint declarations 

mmikitka:
See issue https://github.com/bazelbuild/rules_docker/issues/1247

### DIFF
--- a/python/image.bzl
+++ b/python/image.bzl
@@ -116,7 +116,6 @@ def py_image(name, base = None, deps = [], layers = [], **kwargs):
     app_layer(
         name = name,
         base = base,
-        entrypoint = ["/usr/bin/python"],
         binary = binary_name,
         visibility = visibility,
         tags = tags,
--- a/python/image.bzl
+++ b/python/image.bzl
@@ -116,6 +116,7 @@ def py_image(name, base = None, deps = [], layers = [], **kwargs):
     app_layer(
         name = name,
         base = base,
+        entrypoint = ["python"],
         binary = binary_name,
         visibility = visibility,
         tags = tags,
--- a/python3/image.bzl
+++ b/python3/image.bzl
@@ -111,7 +111,6 @@ def py3_image(name, base = None, deps = [], layers = [], **kwargs):
     app_layer(
         name = name,
         base = base,
-        entrypoint = ["/usr/bin/python"],
         binary = binary_name,
         visibility = visibility,
         tags = tags,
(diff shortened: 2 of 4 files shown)


nlopezgi:
Test py_image_complex_layer is failing, can you double check it please

mmikitka:
I'm seeing the following failure, and the cause is not obvious to me:
```
FAILURE: Expected 'Calling from main module: through py_image_complex_library: Six version: 1.11.0' not found in 'Loaded image ID: sha256:164cea6a3df2588b416bf8393f9a03897e9446eb49ee0f4e7af7ee104c2eebbe

**Comment by mmikitka** [comment]
Note that the py_binary is in the image entrypoint, and the error message `exec user process caused "no such file or directory"` is caused by the absence of `/usr/bin/env` on the system, as referenced in the shebang.

Relevant output of `docker inspect bazel/testdata:py_image_complex`:
```
           "Cmd": null,
            "WorkingDir": "/app/testdata/py_image_complex.binary.runfiles/io_bazel_rules_docker",
            "Entrypoint": [
                "/app/testdata/py_image_complex.binary"
            ],
```

The top of `/app/testdata/py_image_complex.binary` contains:
```
#!/usr/bin/env python
```

And, /usr/bin/env is not in the py_image_base layer.

Nonetheless, this is somewhat beside the point, as we don't want to assume that the Python interpreter resolved by `/usr/bin/env` is the same as the interpreter specified by the toolchain.

Essentially, we need the toolchain-specific Python interpreter to be the first command in the entrypoint.

mmikitka:
Since we don't want language-specific logic (in this case Python) in the app_layer implementation, one solution is to remove the absolute path to Python, and rely on PATH resolution when calling python in the entrypoint.

mmikitka:
@nlopezgi Can we require users to explicitly declare the entrypoint in the event that `which` is not available (e.g., distroless) or the python path cannot be resolved?

Either way, an assumption must be made of the underlying base image i.e., is `which` available or is `/usr/bin/python` available.

However, if you can propose a solution that does not require language-specific code in app_layer, I would gladly implement it.

nlopezgi:
> @nlopezgi Can we require users to explicitly declare the entrypoint in the event that `which` is not available (e.g., distroless) or the python path cannot be resolved?
> 

no, that would break to many existing use cases

> Either way, an assumption must be made of the underlying base image i.e., is `which` available or is `/usr/bin/python` available.
> 

For now, we will assume the latter, as that is the way it's currently implemented.

> However, if you can propose a solution that does not require language-specific code in app_layer, I would gladly implement it.

I can't think of any. Would adding the symlink to /usr/bin/python in the base container be an acceptable workaround for you?

mmikitka:
@nlopezgi I'll resort to explicitly setting the entrypoint or using a symlink as you suggested. Thanks for the conversation, and I think that we can close this issue

Issue #987, opened 2019-07-13
Must migrate rule.output to OutputGroupInfo


or-shachar:
Just checked bazel 28 with the flag `--incompatible_no_rule_outputs_param` and failed on `rules_docker` 

```
ERROR: /private/var/tmp/_bazel_ors/e1cd358f405ae1d1a5186e8a38e284f5/external/io_bazel_rules_docker/container/bundle.bzl:86:21: parameter 'outputs' is deprecated and will be removed soon. It may be temporarily re-enabled by setting --incompatible_no_rule_outputs_param=false
```

I might try to solve and contribute but I wanted to have an open issue about it here:

See details here:
https://github.com/bazelbuild/bazel/issues/7977

Monnoroch:
This is not fixed in `push.bzl`.

PR #1263, opened 2019-11-07
Migrate push.bzl to attr.output

or-shachar:
This is to complete #987 

Thanks @Monnoroch for pointing this out

### DIFF
--- a/container/push.bzl
+++ b/container/push.bzl
@@ -87,10 +87,10 @@ def _impl(ctx):
 
     if ctx.attr.skip_unchanged_digest:
         pusher_args += ["-skip-unchanged-digest"]
-    digester_args += ["--dst", str(ctx.outputs.digest.path), "--format", str(ctx.attr.format)]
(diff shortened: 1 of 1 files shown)


container/push.bzl
```diff
@@ -131,9 +131,9 @@ def _impl(ctx):
         ),
     ]
 
-# Pushes a container image to a registry.
-container_push = rule(
+_container_push = rule(
     attrs = dicts.add({
+        "digest_output": attr.output(),
```
nlopezgi: add doc + add `mandatory = True`

container/push.bzl
```diff
@@ -131,9 +131,9 @@ def _impl(ctx):
         ),
```
nlopezgi: Does this now not also need to also return an
`OutputGroupInfo(digest = depset([digest_output])`
as indicated in https://github.com/bazelbuild/bazel/issues/7977

nlopezgi:
Hi @or-shachar , will you be finishing work on this PR or should we close?

Issue #1265, opened 2019-11-08
On MacOS, the py_image_base tar files that I get has the wrong permissions 


kovkev:
I use bazel-buildfarm to build the images and send them to my local machine.

I made sure `/tmp/some/` has recursively `chmod -R agu+rwx /tmp/some/` all permissions.

However, when bazel tries to get the image from bazel-buildfarm, I think it does `touch 002.tar.gz.nogz` for each file and does not set the right permissions. 

```
Users-MacBook-Pro:__main__ kevin$ ls -lah /tmp/some/execroot/__main__/bazel-out/darwin-fastbuild/bin/external/py_image_base/
total 0
drwxrwxrwx  3 kevin  wheel    96B Nov  8 08:50 .
drwxrwxrwx  3 kevin  wheel    96B Nov  8 08:50 ..
drwxrwxrwx  3 kevin  wheel    96B Nov  8 08:56 image
Users-MacBook-Pro:__main__ kevin$ ls -lah /tmp/some/execroot/__main__/bazel-out/darwin-fastbuild/bin/external/py_image_base/image
total 0
drwxrwxrwx  3 kevin  wheel    96B Nov  8 08:56 .
drwxrwxrwx  3 kevin  wheel    96B Nov  8 08:50 ..
-r-xr-xr-x  1 kevin  wheel     0B Nov  8 08:56 002.tar.gz.nogz
```

I dug a little too much in that direction the last couple of days so I'll take a break on this issue.

nlopezgi:
Hi @kovkev ,
we dont really have a way to repro buildfarm issues, can you try to repro this with a local build so we can help? Otherwise I'm not sure we can provide much assistance.

kovkev:
Hi @nlopezgi , without using buildfarm, the issue does not happen.

Here is what I get with buildfarm:
```
Users-MacBook-Pro:tgkj kevin$ bazel build --verbose_failures --verbose_explanations --explain=/tmp/explain //analysis/thinking:jupyter   --extra_execution_platforms=//toolchains:my_platform,@local_config_platform//:host,@io_bazel_rules_docker//platforms:local_container_platform --logging=6 --experimental_repository_resolved_file=/tmp/a --nolegacy_important_outputs --experimental_remote_grpc_log=/tmp/grpc_log --execution_log_json_file=/tmp/spawns -s
INFO: Writing tracer profile to '/private/tmp/some/command.profile.gz'
[...]

Issue #764, opened 2019-03-28
Need a way to add tags to container layers targets.


atchouprakov:
We are adding `tags=["no-cache"]` to containers to avoid upload these to bazel remote cache. But we couldn't find a way to propagate tag to the layers targets.

nlopezgi:
Hi @atchouprakov,
Thanks for opening this issue. Its possible that we are not properly propagating the tags attribute in all the rules in this repo (probably rules in layer.bzl - https://github.com/bazelbuild/rules_docker/blob/master/container/layer.bzl is not propagating it). It should be a matter of passing them through properly (Example PR: https://github.com/bazelbuild/rules_docker/pull/534). I'd be happy to review a PR fixing this!

sha1n:
@ishikhman I tried using `--experimental_allow_tags_propagation` after tagging some `container_image` and `container_pull` rules with `no-cache` and I do see that it affects the `JoinLayers` actions, but it definitely doesn't affect the generic `GUNZIP` actions that I believe are coming from SkyLib and are not specific to `rules_docker`. 

Is this expected?

I tested with Bazel 1.0.0, 1.0.1 and 1.1.0.

CC @nlopezgi

nlopezgi:
Hi @sha1n ,
No, I don't think this is expected, could you figure out if the actions are indeed in skyline and if so open an issue there? If it turns out otherwise we can see if there is action needed here. Let's reopen until we have more info.

sha1n:
@nlopezgi I'm pretty sure the `GUNZIP` and `GZIP` action mnemonics I saw are the ones implemented here:
https://github.com/bazelbuild/rules_docker/blob/a98bc3b9cb9c530e45d2c74a8da44ab9efdbb0f3/skylib/zip.bzl#L24 
and here:
https://github.com/bazelbuild/rules_docker/blob/a98bc3b9cb9c530e45d2c74a8da44ab9efdbb0f3/skylib/zip.bzl#L37

That file is a part of this repo, so I believe it's not really a part of the bigger skylib. I saw these downloading both in the terminal and I also ran bazel with `--profile` and opened it in chrome tracing to make sure that it is indeed downloading/uploading. 

Let me know if you want me to collect more info.

nlopezgi:
[...]

PR #1272, opened 2019-11-12
define gzip in the docker toolchain not as a local tool

nlopezgi:
local_tool is problematic as it does not work with remote execution.
Gzip, like other tools we use in these rules, should be part of the docker toolchain. This PR removes code related to using local_tool (moves it to a test, as use case is still valid) and moves gzip to the docker_toolchain. Also added tests in java_image to use a java based gzip target.

### DIFF
--- a/skylib/zip.bzl
+++ b/skylib/zip.bzl
@@ -25,7 +25,10 @@ def gzip(ctx, artifact):
     """
     out = ctx.actions.declare_file(artifact.basename + ".gz")
     toolchain_info = ctx.toolchains["@io_bazel_rules_docker//toolchains/docker:toolchain_type"].info
-    if toolchain_info.gzip_path == "":
+    tools = []
+    if toolchain_info.gzip_target:
+        tools = toolchain_info.gzip_target.executable
+    elif toolchain_info.gzip_path == "":
         fail("gzip could not be found. Make sure it is in the path or set it " +
              "explicitly in the docker_toolchain_configure")
     ctx.actions.run_shell(
@@ -34,6 +37,7 @@ def gzip(ctx, artifact):
         outputs = [out],
         use_default_shell_env = True,
         mnemonic = "GZIP",
+        tools = tools,
     )
     return out
 
@@ -49,7 +53,10 @@ def gunzip(ctx, artifact):
     """
     out = ctx.actions.declare_file(artifact.basename + ".nogz")
     toolchain_info = ctx.toolchains["@io_bazel_rules_docker//toolchains/docker:toolchain_type"].info
-    if toolchain_info.gzip_path == "":
+    tools = []
+    if toolchain_info.gzip_target:
+        tools = toolchain_info.gzip_target.executable
+    elif toolchain_info.gzip_path == "":
         fail("gzip could not be found. Make sure it is in the path or set it " +
              "explicitly in the docker_toolchain_configure")
     ctx.actions.run_shell(
@@ -58,5 +65,6 @@ def gunzip(ctx, artifact):
         outputs = [out],
         use_default_shell_env = True,
         mnemonic = "GUNZIP",
+        tools = tools,
     )
     return out
--- a/skylib/zip.bzl
+++ b/skylib/zip.bzl
@@ -25,19 +25,18 @@ def gzip(ctx, artifact):
     """
     out = ctx.actions.declare_file(artifact.basename + ".gz")
     toolchain_info = ctx.toolchains["@io_bazel_rules_docker//toolchains/docker:toolchain_type"].info
-    inputs = []
     input_manifests = []
     tools = []
     gzip_path = toolchain_info.gzip_path
     if toolchain_info.gzip_target:
-        gzip_path = toolchain_info.gzip_target.executable.path
+        gzip_path = toolchain_info.gzip_target.files.to_list()[0].path
         tools, _, input_manifests = ctx.resolve_command(tools = [toolchain_info.gzip_target])
     elif toolchain_info.gzip_path == "":
         fail("gzip could not be found. Make sure it is in the path or set it " +
              "explicitly in the docker_toolchain_configure")
 
     ctx.actions.run_shell(
-        command = "%s -n < %s > %s" % (toolchain_info.gzip_path, artifact.path, out.path),
+        command = "%s -n < %s > %s" % (gzip_path, artifact.path, out.path),
         input_manifests = input_manifests,
         inputs = [artifact],
         outputs = [out],
@@ -59,14 +58,17 @@ def gunzip(ctx, artifact):
     """
     out = ctx.actions.declare_file(artifact.basename + ".nogz")
     toolchain_info = ctx.toolchains["@io_bazel_rules_docker//toolchains/docker:toolchain_type"].info
+    input_manifests = []
     tools = []
+    gzip_path = toolchain_info.gzip_path
     if toolchain_info.gzip_target:
-        tools = toolchain_info.gzip_target.executable
+        gzip_path = toolchain_info.gzip_target.files.to_list()[0].path
+        tools, _, input_manifests = ctx.resolve_command(tools = [toolchain_info.gzip_target])
     elif toolchain_info.gzip_path == "":
         fail("gzip could not be found. Make sure it is in the path or set it " +
              "explicitly in the docker_toolchain_configure")
     ctx.actions.run_shell(
-        command = "%s -d < %s > %s" % (toolchain_info.gzip_path, artifact.path, out.path),
+        command = "%s -d < %s > %s" % (gzip_path, artifact.path, out.path),
         inputs = [artifact],
         outputs = [out],
         use_default_shell_env = True,
--- a/skylib/zip.bzl
+++ b/skylib/zip.bzl
@@ -14,36 +14,49 @@
 """Functions for producing the gzip of an artifact."""
 
 def gzip(ctx, artifact):
-    """Create an action to compute the gzipped artifact."""
+    """Create an action to compute the gzipped artifact.
+
+    Args:
+       ctx: The context
+       artifact: The artifact to zip
+
+    Returns:
+       the gzipped artifact.
+    """
     out = ctx.actions.declare_file(artifact.basename + ".gz")
+    toolchain_info = ctx.toolchains["@io_bazel_rules_docker//toolchains/docker:toolchain_type"].info
+    if toolchain_info.gzip_path == "":
+        fail("gzip could not be found. Make sure it is in the path or set it " +
+             "explicitly in the docker_toolchain_configure")
     ctx.actions.run_shell(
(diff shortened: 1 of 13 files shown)


nlopezgi:
this still needs some tests, but creating to test if this fixes both:
https://github.com/bazelbuild/rules_docker/issues/1265
https://github.com/bazelbuild/rules_docker/issues/764

darl:
It didn't seem to work.
`which gzip` executed on host machine and resolved to `/usr/bin/gzip`
which is not available on linux machine.

I think a better approach is to implement a simple `gzip` alternative with `go` or `python`.
It should be more portable and reproducible.

nlopezgi:
> It didn't seem to work.
> `which gzip` executed on host machine and resolved to `/usr/bin/gzip`
> which is not available on linux machine.


which gzip is the default behavior that works for local execution when gzip is properly installed and in the path. To try this for remote execution you will need to set up a custom 
`docker_toolchain_configure` rule in your WORKSPACE and pass an explicit path for your remote gzip.
See https://github.com/bazelbuild/rules_docker#setup (though obviously not updated to include the new attrs yet)

> 
> I think a better approach is to implement a simple `gzip` alternative with `go` or `python`.
> It should be more portable and reproducible.

That approach would require much more maintenance on our part, which we are not willing to take on. Please try using a custom `docker_toolchain_configure` first.

darl:
Ok. I successfully built my project with `docker_toolchain_configure` and explicit `gzip` path and remote execution enabled.

But I don't like the idea of having absolute paths in WORKSPACE file.
Also `select` is not supported in workspace files so I cannot specify platform-dependent path.

Can we change `gzip_path` parameter to `attr.label` which defaults to old `@gzip//:gzip`?
So I can provide my `gzip` binary.

nlopezgi:
> Ok. I successfully built my project with `docker_toolchain_configure` and explicit `gzip` path and remote execution enabled.
> 
> But I don't like the idea of having absolute paths in WORKSPACE file.

Yeah, its unfortunate, but the info likely has to go somewhere, unless you want to provide your own buildable gzip target (which sounds like you do).

> Also `select` is not supported in workspace files so I cannot specify platform-dependent path.
> 

The solution here for multi platform builds would be to define multiple toolchains and select them via command line flags. But that's even more involved, so let's try to get you moving with a target based gzip as you suggest.

> Can we change `gzip_path` parameter to `attr.label` which defaults to old `@gzip//:gzip`?
> So I can provide my `gzip` binary.
> 

Done, you should be able now to define gzip as an executable target. As I mentioned, I dont really have a good way to test this though. Can you give it a whirl and try to set gzip_target instead of gzip_path in the `docker_toolchain_configure` and let me know if that works?
If it fails, its due to this line in zip.bzl (likely):
```
    if toolchain_info.gzip_target:
        tools = toolchain_info.gzip_target.executable
```
Which could also be:
```
    if toolchain_info.gzip_target:
        tools = toolchain_info.gzip_target.files.to_list()[0]
```
Could you try this second alternative if you get a failure and get back to me?

> p.s. actually I can provide relative path and it works.
> 
> ```
> docker_toolchain_configure(
>   name = "docker_config",
>   gzip_path = "gzip"
> )
> ```
What you will sacrifice if you use relative paths, though, is that builds will no longer be hermetic, and if you are using, e.g., a remote cache, you could poison it very easily.

darl:
yep.
```
	File "/private/var/tmp/_bazel_darl/e51d20ff8cabb9ee5d304a36295c1184/external/vertis_java8/image/BUILD", line 4
		container_import(name = 'image')
	File "/private/var/tmp/_bazel_darl/e51d20ff8cabb9ee5d304a36295c1184/external/io_bazel_rules_docker/container/import.bzl", line 77, in _container_import_impl
		_layer_pair(ctx, layer)
	File "/private/var/tmp/_bazel_darl/e51d20ff8cabb9ee5d304a36295c1184/external/io_bazel_rules_docker/container/import.bzl", line 62, in _layer_pair
		_gunzip(ctx, layer)
	File "/private/var/tmp/_bazel_darl/e51d20ff8cabb9ee5d304a36295c1184/external/io_bazel_rules_docker/skylib/zip.bzl", line 58, in _gunzip
		toolchain_info.gzip_target.executable
object of type 'string' has no field 'executable'
```

toolchains/docker/toolchain.bzl
```diff
@@ -52,6 +58,14 @@ docker_toolchain = rule(
                   " DOCKER_CONFIG is not defined, the home directory will be" +
                   " used.",
         ),
+        "gzip_path": attr.string(
+            doc = "Path to the gzip binary. " +
+                  "Should only be set if gzip_target is unset.",
+        ),
+        "gzip_target": attr.string(
```
darl: shouldn't it be `attr.label`?

darl:
after some debugging got working patch:
https://github.com/darl/rules_docker/commit/d42f0e11015243ac2c6151f41a0272b623abc49f

my simple gzip target used in testing:
```java
package tools.gzip;
public class Gzip {
    public static void main(String[] args) throws IOException {
        boolean decompress = args.length > 0 && args[0].equals("-d");
        InputStream in = decompress ? new GZIPInputStream(System.in, 8 * 1024) : System.in;
        OutputStream out = decompress ? System.out : new GZIPOutputStream(System.out);
        byte[] buffer = new byte[8 * 1024 * 1024];
        int n;
        while (-1 != (n = in.read(buffer))) {
            out.write(buffer, 0, n);
        }
        out.close();
        in.close();
    }
}
```
```
java_binary(
    name = "gzip",
    srcs = ["Gzip.java"],
    deps = [],
    main_class = "tools.gzip.Gzip",
)
```

toolchains/docker/toolchain.bzl
nlopezgi: yeah, thanks, fixed

nlopezgi:
> after some debugging got working patch:
> [darl@d42f0e1](https://github.com/darl/rules_docker/commit/d42f0e11015243ac2c6151f41a0272b623abc49f)

Thanks, copied your patch with some minor changes, would you mind testing again on your side?

darl:
@nlopezgi 
It's not working with latest changes when using `java_binary` as executable.

I added some debug logging to `zip.bzl`
```
print("toolchain_info.gzip_target.files = " + str(toolchain_info.gzip_target.files.to_list()))
print("toolchain_info.gzip_target.files[0].path = " + str(toolchain_info.gzip_target.files.to_list()[0].path))
```

And it appears that first file is not always a binary:
```
DEBUG: /private/var/tmp/_bazel_darl/e51d20ff8cabb9ee5d304a36295c1184/external/io_bazel_rules_docker/skylib/zip.bzl:65:9: toolchain_info.gzip_target.files = [<generated file external/my_workspace/tools/gzip/gzip.jar>, <generated file external/my_workspace/tools/gzip/gzip>]
DEBUG: /private/var/tmp/_bazel_darl/e51d20ff8cabb9ee5d304a36295c1184/external/io_bazel_rules_docker/skylib/zip.bzl:66:9: toolchain_info.gzip_target.files[0].path = bazel-out/host/bin/external/my_workspace/tools/gzip/gzip.jar
```

skylib/zip.bzl
```diff
@@ -14,36 +14,65 @@
...
+    gzip_path = toolchain_info.gzip_path
+    if toolchain_info.gzip_target:
+        gzip_path = toolchain_info.gzip_target.files.to_list()[0].path
+        tools, _, input_manifests = ctx.resolve_command(tools = [toolchain_info.gzip_target])
+    elif toolchain_info.gzip_path == "":
+        fail("gzip could not be found. Make sure it is in the path or set it " +
+             "explicitly in the docker_toolchain_configure")
     ctx.actions.run_shell(
-        command = "%s -d < %s > %s" % (ctx.executable.gzip.path, artifact.path, out.path),
+        command = "%s -d < %s > %s" % (gzip_path, artifact.path, out.path),
```
darl:
```suggestion
        command = "%s -d < %s > %s" % (gzip_path, artifact.path, out.path),
        input_manifests = input_manifests,
```

skylib/zip.bzl
```diff
@@ -14,36 +14,65 @@
...
+    Returns:
+       the gunzipped artifact.
+    """
     out = ctx.actions.declare_file(artifact.basename + ".nogz")
+    toolchain_info = ctx.toolchains["@io_bazel_rules_docker//toolchains/docker:toolchain_type"].info
+    input_manifests = []
+    tools = []
+    gzip_path = toolchain_info.gzip_path
+    if toolchain_info.gzip_target:
+        gzip_path = toolchain_info.gzip_target.files.to_list()[0].path
```
darl: maybe `toolchain_info.gzip_target.files_to_run.executable.path`?

nlopezgi: thanks, that seems to work. I also folded your example as a test as part of this PR. While I cant test it on remote execution we can have a test that verifies gzip as tool works as intended. Thanks for your help!

skylib/zip.bzl
nlopezgi: fixed

toolchains/docker/BUILD.tpl
```diff
@@ -20,7 +20,8 @@ load("@io_bazel_rules_docker//toolchains/docker:toolchain.bzl", "docker_toolchai
 
 docker_toolchain(
     name = "toolchain",
-    tool_path = "%{DOCKER_TOOL}",
     client_config = "%{DOCKER_CONFIG}",
+    %{GZIP_ATTR}
```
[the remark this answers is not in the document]
nlopezgi: I'm replacing this for either a gzip_target or gzip_path, adding the comma in the template replacement

toolchains/docker/toolchain.bzl
```diff
@@ -114,6 +147,26 @@ toolchain_configure = repository_rule(
...
+            allow_files = True,
+            mandatory = False,
+            doc = "The bazel target for the gzip tool. " +
+                  "Can only be set if gzip_path is not set.",
+        ),
+        "xz_path": attr.string(
+            mandatory = False,
+            doc = "The full path to the xz binary. If not specified, it will" +
+                  "be searched for in the path. If not available, running commands" +
+                  "that xz will fail.",
```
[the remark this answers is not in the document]
nlopezgi: fixed

toolchains/docker/toolchain.bzl
```diff
@@ -114,6 +147,26 @@ toolchain_configure = repository_rule(
                   "be searched for in the path. If not available, running commands" +
                   "that require docker (e.g., incremental load) will fail.",
         ),
+        "gzip_path": attr.string(
+            mandatory = False,
+            doc = "The full path to the gzip binary. If not specified, it will" +
+                  "be searched for in the path. If not available, running commands" +
+                  "that gzip will fail.",
```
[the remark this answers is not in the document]
nlopezgi: fixed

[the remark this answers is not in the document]
nlopezgi: fixed

PR #1277, opened 2019-11-16
install_pkgs uses named volumes to work with DIND.

ralimi:
Previously, install_pkgs would mount files into
the target image directly (with '-v src:dst').
This unfortunately doesn't work with
docker-in-docker (DIND) setups where the mount
is done from the perspective of the host which
does not share a filesystem with the docker client.

To work with such setups, install_pkgs now creates
a named volume and copies files into the named
volume.  The named volume containing the install
script and TAR file is then mounted into the
target container.


nlopezgi:
This PR is producing multiple failures on buildkite presubmit. For example:
```
 ERROR: /workdir/tests/docker/package_managers/BUILD:135:1:
++ /usr/bin/docker create -v a6cd30043e5d12cb39f09df13eddcfe47e92070a8ec1551fa5bff1fa5dba2463:/tmp/pkginstall c0c84a906489f3f241b67702c57a146f0976e3e752600ff58361e81a17215527
Error response from daemon: No command specified
```
Please check the errors out and make sure you can build all these targets / run these tests locally with your changes.

ralimi:
Thanks for pointing that out. I've re-run the test under tests/docker/package_managers that showed that problem, and I've adjusted the code to fix that.

There are quite a few tests under this repository that simply fail for me, and they look to be for a number of reasons that are unrelated to this change (insufficient permissions for a project in GCR, join_layers gets killed because it runs out of RAM while loading the image).

ralimi:
Because of those failures, it's somewhat hard to interpret the results for a run of the full repository. Let me know if I'm missing something?

nlopezgi:
/gcbrun -  buildkite CI is having some infra problems. Running GCB tests anyway to check PR.

docker/package_managers/installer.sh.tpl
```diff
@@ -3,9 +3,8 @@
 # It expects to be volume-mounted inside a docker image, in /tmp along with the
 # installables.tar.
 set -ex
-pushd /tmp
+pushd /tmp/pkginstall
```
nlopezgi: Update docs above

docker/package_managers/run_install.sh.tpl
```diff
@@ -17,11 +17,41 @@ source %{util_script}
...
+# to the named volume.
+
+# Prepare directory structure. 'docker cp' will not create
+# intermediate paths.
+tmpdir=$(mktemp -d)
+trap "rm -rf $tmpdir" EXIT
+mkdir -p $(dirname $tmpdir/%{installables_tar})
+cp -L $(pwd)/%{installables_tar} $tmpdir/%{installables_tar}
+cp -L $(pwd)/%{installer_script} $tmpdir/installer.sh
+# Temporarily create a container sowe can mount the named volume
```
nlopezgi: so we

docker/package_managers/installer.sh.tpl
ralimi: Done.

docker/package_managers/run_install.sh.tpl
ralimi: Done.

ralimi:
No problem at all - thanks for the review!

Fixed up the docs.

PR #1299, opened 2019-11-26
move python modules out of bzl file

cristifalcas:
I moved only the pip packages in WORKSPACE.


cristifalcas:
@googlebot I signed it!

PR #1300, opened 2019-11-27
move python modules out of bzl file (#1299)

projectghostshell:


smukherj1:
Hi there! The target branch of this PR is set to `bazelbuild:client-0.0.9`. Is that intentional? You probably want `bazelbuild:master`.

PR #1302, opened 2019-11-29
Use 'portable' mtime as default for tars, like rules_pkg does

zoidbergwill:
`rules_pkg` defaults to the following:
- `mtime`: -1
- `portable_mtime`: True

Which sends `--mtime=portable` to rules_pkg's [`build_tar.py`](https://github.com/bazelbuild/rules_pkg/blob/master/pkg/build_tar.py) by default.

Potentially we could make these configurable like they are for `rules_pkg`'s `pkg_tar` rules.

I needed this for a program using an on-disk cache in `/tmp`, that compared `mtime` of the file it was caching and the file in the cache, but also using `0` as the value for a file missing in the cache.


zoidbergwill:
Maybe I should make a PR to move some of the logic from rules_pkg's pkg.bzl and build_tar.py to be in archive.py instead.

zoidbergwill:
Oh wait, this isn't backwards compatible. This'll cause a lot of cache misses for folks already using the rules.

Maybe I should just make them configurable instead.

smukherj1:
The tests are failing because as you indicated, the change is not backwards compatible. The image tarballs produced are now timestamped differently and thus tests that expected an image with a specific digest are now failing. A backwards compatible way to implement this might be to keep the previous default timestamp `0` but add an attribute to the `container_layer` and `container_image` rules to override this setting. So the steps to do this would be:
1. Modify the build_tar.py script to accept a flag to configure the default_mtime passed to TarFileWriter.
2. Add a new `default_mtime` attribute to `container_layer` [here](https://github.com/bazelbuild/rules_docker/blob/f13ac8133b5f1c3e3ea419ee07f131423174e589/container/layer.bzl#L224). This will automatically add this attribute to `container_image` as well since it inherits all layer attributes.
3. Pass in the value of the new `default_mtime` attribute to `build_layer` [here](https://github.com/bazelbuild/rules_docker/blob/f13ac8133b5f1c3e3ea419ee07f131423174e589/container/layer.bzl#L65) and the layer implementation function [here](https://github.com/bazelbuild/rules_docker/blob/f13ac8133b5f1c3e3ea419ee07f131423174e589/container/layer.bzl#L148).
4. Pass in the value of `default_mtime` from the container_image rule to the layer implementation function [here](https://github.com/bazelbuild/rules_docker/blob/f13ac8133b5f1c3e3ea419ee07f131423174e589/container/image.bzl#L363).

zoidbergwill:
Awesome, thanks for the quick feedback. @smukherj1

zoidbergwill:
Can I run the "Trigger" tests locally?

zoidbergwill:
Ohhhh, they need someone to `/gcbrun`

container/layer.bzl
```diff
@@ -103,6 +103,13 @@ def build_layer(
     args.add(directory, format = "--directory=%s")
     args.add(ctx.attr.mode, format = "--mode=%s")
 
+    if ctx.attr.mtime != -1:  # Note: Must match default in rule def.
```
smukherj1: Could use a local private variable at the top of this file to do this automatically. So at the top of this file, consider declaring
`_DEFAULT_MTIME = -1`
and use it here and in the rule attribute definition.

zoidbergwill: Great idea. Better than littering the code with duplicated magic numbers.

zoidbergwill: Duplicated it upstream to rules_pkg too https://github.com/bazelbuild/rules_pkg/pull/125

zoidbergwill:
Yay, most of the triggers passed. 🎉

Unfortunately I can't see the output. 🐼

smukherj1:
Hmm the test that's failing is
```
bazel --test_output=errors --spawn_strategy=standalone --verbose_failures //tests/contrib:set_cmd_repro_test 
```

the error is
```
rm: cannot remove '/workspace/output_base/execroot/io_bazel_rules_docker/bazel-out/k8-fastbuild/bin/tests/contrib/bazel_out1': Device or resource busy
```

not sure how your PR could have caused this. Does this reproduce for you locally?

zoidbergwill:
@smukherj1 I'm getting a different error when running it on my machine, for master as well as this branch.

Oh, because I'm on Mac.

Gonna try on a Linux machine.

PR #1318, opened 2019-12-13
Change to using python 3 in host configuration

Panfilwk:
I made this change in an internal fork of this repo that I help maintain for my job, but figured it might be useful upstream as well considering the approaching deprecation of Python 2. This _shouldn't_ cause any issues with containers being able to run Python 2 still as far as I can tell, it just changes tools to run in Python 3 and removes any instances of the `--host_force_python=PY2"` option.

If for some reason this shouldn't be merged right now (or at all) please let me know.

### DIFF
--- a/container/BUILD
+++ b/container/BUILD
@@ -31,7 +31,7 @@ py_binary(
     name = "build_tar",
     srcs = ["build_tar.py"],
     legacy_create_init = True,
-    python_version = "PY2",
+    python_version = "PY3",
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [":build_tar_lib"],
@@ -128,7 +128,7 @@ py_test(
     size = "medium",
     srcs = ["image_test.py"],
     data = TEST_DATA,
-    python_version = "PY2",
+    python_version = "PY3",
     deps = ["@containerregistry"],
 )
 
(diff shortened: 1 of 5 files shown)


nlopezgi:
hi @Panfilwk ,
Thanks for creating this PR.
It looks like some tests are failing (//container:image_test ). Some of our tests still have deps on py containerregistry and probably need to be run with PY2. For most other targets, though, I think this is a good change to get in. Can you verify all tests pass for you locally and let me know to re run tests once you have resolved the issues.

nlopezgi:
thanks @Panfilwk , can you resolve the conflicts in tools/BUILD please?

Panfilwk:
I'll take a look at the conflicts.

Also, I tried (very briefly) to upgrade the`//container:image_test` test to Python 3, but I think the better solution might be to rewrite it in golang using the newer go-containerregistry code. I may take a stab at that if you also think that would be valuable, but in a separate PR.

nlopezgi:
thanks @Panfilwk ,
I do think it would be valuable to migrate the tests to use go and be done with any py2 deps we have. Would be super happy to review an upcoming PR for that. Thanks again. Will kick of gcb tests and merge once they pass.

nlopezgi:
Oh, and the CLA check got stuck, can you confirm verbally here in comments that you "agree with the CLA" and I'll be able to override it manually.
Thanks!

Panfilwk:
I agree with the CLA.

(I think it probably got stuck because I've signed the CLA through my work email, but my personal email is the one primarily associated with my GitHub account)

PR #1324, opened 2019-12-17
Update rules_nodejs to 0.42.3

gregmagolan:


alexeagle:
oh awesome thank you!
@nlopezgi do you have an idea when you'll cut a rules_docker release? Since this is in, it lets us remove the old API for our 1.0 release which we're hoping to cut before the holidays. But I don't want rules_nodejs users to trip over their nodejs_image rules if there's no rules_docker release that works with our 1.0.

gregmagolan:
Thanks @nlopezgi !

nlopezgi:
Release 0.13.0 is out

PR #1353, opened 2020-01-06
Add robot_email attr to container push bzl rules

tabletenniser:
The robot_email attr is used at Google internally and github users should not be using it.

### DIFF
--- a/container/push.bzl
+++ b/container/push.bzl
@@ -187,6 +187,11 @@ container_push = rule(
             default = Label("//container:push-tag.sh.tpl"),
             allow_single_file = True,
         ),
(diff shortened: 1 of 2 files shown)


container/push.bzl
```diff
@@ -187,6 +187,11 @@ container_push = rule(
             default = Label("//container:push-tag.sh.tpl"),
             allow_single_file = True,
         ),
+        "robot_email": attr.string(
+            default = "",
```
nlopezgi: Move to line 159

tabletenniser: Created a new commit to move it, both here and contrib/push-all.bzl.

PR #1357, opened 2020-01-08
update_deps in Go

zoidbergwill:
This removes one of two uses of the Python containerregistry

I wanted to use a copy of `update_deps.py` in a repo of ours that uses rules_docker, and didn't want anything that needs Python 2, so I thought moving this to Go was the right approach.

Having a public binary for this would also be super useful for encouraging folks to use digests instead of tags.


tools/BUILD
```diff
@@ -13,12 +13,7 @@
 # limitations under the License.
 package(default_visibility = ["//visibility:private"])
 
-py_binary(
+alias(
```
zoidbergwill: This is just to have backwards compatibility with `update_deps.sh`, since the target is private, we could just move `update_deps.sh` to reference the new target at `//container/go/cmd/update_deps`

container/go/cmd/update_deps/BUILD
```diff
@@ -0,0 +1,39 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ###
+# Synthesize a .bzl file containing the digests for a given repository.
+
+load("@bazel_gazelle//:def.bzl", "gazelle")
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+# gazelle:prefix github.com/bazelbuild/rules_docker
+gazelle(name = "gazelle")
```
smukherj1: This should not be necessary. We already have this in the top level BUILD file [here](https://github.com/bazelbuild/rules_docker/blob/master/BUILD).

container/go/cmd/update_deps/update_deps.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	if *output == "" {
+		log.Fatalln("Required option -output was not specified.")
+	}
+	options := []crane.Option{}
+
+	latest := *repository + ":latest"
+	latestDigest, err := crane.Digest(latest, options...)
+	if err != nil {
+		log.Fatalf("computing digest for %s: %v", latest, err)
```
smukherj1: computing -> Computing

container/go/cmd/update_deps/update_deps.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+
+	debug := *repository + ":debug"
+	debugDigest, err := crane.Digest(debug, options...)
+	if err != nil {
+		if !strings.HasPrefix(err.Error(), "MANIFEST_UNKNOWN") {
+			log.Fatalf("computing digest for %s: %v", debug, err)
```
smukherj1: computing -> Computing

container/go/cmd/update_deps/update_deps.go
```diff
@@ -0,0 +1,117 @@
...
+		DebugTag:  debugDigest,
+		LatestTag: latestDigest,
+		Debug:     debug,
+		Latest:    latest,
+		Date:      date,
+	}
+
+	f, err := os.Create(*output)
+	if err != nil {
+		log.Fatalf("failed to open file %s: %s", *output, err)
```
smukherj1: failed -> Failed

container/go/cmd/update_deps/update_deps.go
```diff
@@ -0,0 +1,117 @@
...
+	}
+	defer f.Close()
+
+	err = t.Execute(f, r)
+	if err != nil {
+		log.Fatalf("executing template:", err)
```
smukherj1: executing -> Executing

tools/BUILD
smukherj1: Couldn't we just change `update_deps.sh` to `bazel run //container/go/cmd/update_deps` instead?

zoidbergwill:
@smukherj1 Thanks for all the feedback, addressing it now.

container/go/cmd/update_deps/BUILD
zoidbergwill: Probably, don't really know what it's here for. I copied the puller/BUILD file originally I think.

tools/BUILD
zoidbergwill: That does seem much cleaner

zoidbergwill:
I think I've addressed all feedback

container/go/cmd/update_deps/BUILD
```diff
@@ -11,14 +11,26 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-package(default_visibility = ["//visibility:private"])
+# ###
+# Synthesize a .bzl file containing the digests for a given repository.
 
-py_binary(
+load("@bazel_gazelle//:def.bzl", "gazelle")
```
nlopezgi: buildifier is complaining this is unused

zoidbergwill: Whoops, good catch. Thanks! Forgot to run buildifier before committing changes

zoidbergwill:
Should be good now.

PR #1367, opened 2020-01-13
Path join issues in contrib/dockerfile_build.bzl

bajacondor:
I've been getting errors that can be fixed with `--incompatible_string_join_requires_strings=false` from  [line 79](https://github.com/bazelbuild/rules_docker/blob/master/contrib/dockerfile_build.bzl#L79) of `dockerfile_build.bzl` when the docker build fails.  This obfuscates the real error.  The problem comes from trying to join the `dockerfile_path` as a string, because its a `path` type.


nlopezgi:
@bajacondor we're still getting issues for your CLA, can you confirm here verbally you approve it and I will manually override the check? (just comment "I approve the CLA")

bajacondor:
I approve the CLA personally and as an employee of Nordstrom

PR #1371, opened 2020-01-14
Not all rules have deps, don't assume they exist

NathanHowell:
The app layer filters assume all rules have a `deps` attribute. This causes build failures if `deps` is missing. It seems like this should be optional or configurable. This PR makes it optional.

If you're okay with this approach I'll add some tests.


nlopezgi:
> The app layer filters assume all rules have a `deps` attribute. This causes build failures if `deps` is missing. It seems like this should be optional or configurable. This PR makes it optional.
> 
> If you're okay with this approach I'll add some tests.

Can you explain a bit more your use case for this? Which rule are you using along with the xx_image rules that does not have deps (can you post here your BUILD file for the rule)? Why does it not have deps?

NathanHowell:
@nlopezgi  here is a [small repro case](https://gist.github.com/NathanHowell/59373178c881c2d3d985ac307e561359)

the examples in https://docs.bazel.build/versions/master/skylark/aspects.html#simple-example check for existence of an attribute so I figured this practice is recommended when writing aspects. if this is incorrect that's fine, but it seems like an arbitrary requirement. The built image contains what I expect and runs.

let me know if you have other recommendations to fix this. if i need to add an empty `deps` attribute as a workaround I'd like to know why this is important- there is a lot of tribal knowledge not captured in the Bazel docs. thanks!

NathanHowell:
@nlopezgi have you had a chance to look at this yet?

Issue #1421, opened 2020-02-14
Incorrect architecture label for container image


murilofv:
I have an issue when using rules_docker in a different architecture (in my case ppc64le) where all the container images generated get the label as amd64 architecture despite the architecture I am running in.

### Steps to recreate
Need a system from a different architecture than amd64.

The following is an example of a `WORKSPACE` file:
```
load(
    "@bazel_tools//tools/build_defs/repo:http.bzl",
    "http_archive",
    "http_file",
)

http_archive(
    name = "io_bazel_rules_docker",
    sha256 = "dc97fccceacd4c6be14e800b2a00693d5e8d07f69ee187babfd04a80a9f8e250",
    strip_prefix = "rules_docker-0.14.1",
    urls = ["https://github.com/bazelbuild/rules_docker/releases/download/v0.14.1/rules_docker-v0.14.1.tar.gz"],
)

load(
    "@io_bazel_rules_docker//repositories:repositories.bzl",
    container_repositories = "repositories",
)
container_repositories()

load("@io_bazel_rules_docker//repositories:deps.bzl", container_deps = "deps")

container_deps()

load(
    "@io_bazel_rules_docker//container:container.bzl",
    "container_image",
    "container_pull",
)

http_file(
    name = "go_puller_linux_ppc64le",
    executable = True,
    sha256 = "540f4d7b2a3d627d7c3190f11c4fab5f8aad48bd42a9dffb037786e26270b6bd",
    urls = [
        "https://storage.googleapis.com/builddeps/540f4d7b2a3d627d7c3190f11c4fab5f8aad48bd42a9dffb037786e26270b6bd",
    ],
)

container_pull(
    name = "fedora_ppc64le",
    digest = "sha256:de1e23d807365800621d10efe27fa7eea865ee0bf9f1beca316f919972312695",
    puller_linux = "@go_puller_linux_ppc64le//file:downloaded",
    registry = "index.docker.io",
    repository = "library/fedora",
)
```
Note: I am using a custom rebuilt `puller` binary given it is currently only built for amd64

The following is an example of the `BUILD.bazel` file:
```
load(
    "@io_bazel_rules_docker//container:container.bzl",
    "container_image",
    "container_push",
)

container_image(
    name = "fedora30_ppc64le",
    base = "@fedora_ppc64le//image",
    visibility = ["//visibility:public"],
)

container_push(
    name = "push-fedora30_ppc64le",
    format = "Docker",
[...]

PR #1427, opened 2020-02-18
Expose architecture in container_image

murilofv:
container_image builds all images labeled as amd64 despite the
possibility of it being intended for another architecture, without a way
to overwrite this value. Aside from being confusing to a user, this could
also be an issue when using said images as part of a manifest list, given
the wrong label won't pull the proper image to the system.

This patch exposes an architecture attribute that can be used by an user
in order to properly label the container image. The default is still amd64
in case it is not set.

It also adds a sh_test called architecture_test to verify that when a
labeled is set it was actually be properly modified in the config of the
target image.

Signed-off-by: Murilo Fossa Vicentini <muvic@linux.ibm.com>

Fixes https://github.com/bazelbuild/rules_docker/issues/1421


murilofv:
@smukherj1 is this what you had in mind regarding for your comment in https://github.com/bazelbuild/rules_docker/issues/1421#issuecomment-587500437 ?

murilofv:
@googlebot I signed it!

murilofv:
> Thanks! The PR looks good! Could you please add a test? I suggest adding a `container_image` target [here](https://github.com/bazelbuild/rules_docker/blob/master/tests/container/BUILD) and then maybe write a `sh_test` that depends on the `<target>.json` and validates that the architecture string is present in the config.

Sure, I can do that, do you want me to run it as as part of any automated test currently in place like `.bazelci/presubmit.yml` or `testing/e2e.sh` or you want me to just add it so folks can run it manually if needed?

smukherj1:
> > Thanks! The PR looks good! Could you please add a test? I suggest adding a `container_image` target [here](https://github.com/bazelbuild/rules_docker/blob/master/tests/container/BUILD) and then maybe write a `sh_test` that depends on the `<target>.json` and validates that the architecture string is present in the config.
> 
> Sure, I can do that, do you want me to run it as as part of any automated test currently in place like `.bazelci/presubmit.yml` or `testing/e2e.sh` or you want me to just add it so folks can run it manually if needed?

Would prefer it runs automatically in the Bazel CI presubmit which it should (if added as an sh_test target) unless explicitly excluded in the `.bazelci/presubmit.yaml` file.

murilofv:
> Would prefer it runs automatically in the Bazel CI presubmit which it should (if added as an sh_test target) unless explicitly excluded in the `.bazelci/presubmit.yaml` file.

I might be missing something, was it expected to run automatically indeed? I am looking at the build results in https://buildkite.com/bazel/rules-docker-docker/builds/3872 and I don't see the test listed there. I assume it doesn't actually run automatically and I need to explicitly make it a part of test_targets in `.bazelci/presubmit.yaml`, is this the case?

smukherj1:
> > Would prefer it runs automatically in the Bazel CI presubmit which it should (if added as an sh_test target) unless explicitly excluded in the `.bazelci/presubmit.yaml` file.
> 
> I might be missing something, was it expected to run automatically indeed? I am looking at the build results in https://buildkite.com/bazel/rules-docker-docker/builds/3872 and I don't see the test listed there. I assume it doesn't actually run automatically and I need to explicitly make it a part of test_targets in `.bazelci/presubmit.yaml`, is this the case?

Apologies. I was assuming we used `bazel test ...` to run the test and only listed `-<taget>` for tests we wanted to exclude. I see this was changed by this [commit](https://github.com/bazelbuild/rules_docker/commit/0237f78e90155e5d0095ed1ec6780cea1b618964). So yes, please go ahead and add your new test to the presubmit.yaml for all the test environments.

murilofv:
> > > Would prefer it runs automatically in the Bazel CI presubmit which it should (if added as an sh_test target) unless explicitly excluded in the `.bazelci/presubmit.yaml` file.
> > 
> > 
> > I might be missing something, was it expected to run automatically indeed? I am looking at the build results in https://buildkite.com/bazel/rules-docker-docker/builds/3872 and I don't see the test listed there. I assume it doesn't actually run automatically and I need to explicitly make it a part of test_targets in `.bazelci/presubmit.yaml`, is this the case?
> 
> Apologies. I was assuming we used `bazel test ...` to run the test and only listed `-<taget>` for tests we wanted to exclude. I see this was changed by this [commit](https://github.com/bazelbuild/rules_docker/commit/0237f78e90155e5d0095ed1ec6780cea1b618964). So yes, please go ahead and add your new test to the presubmit.yaml for all the test environments.

No problem, I added there an it shows up now, thank you. Let me know if you have any additional feedback / concerns regarding the patch.

murilofv:
Just out of curiosity, now that I noticed, did something happen that the commit message body was not added during the merge? This doesn't look like something expected, but I unfortunately deleted the branch / fork after the merge happened and looks like the references were lost, so haven't been able to look into what could possibly have happened.

PR #1428, opened 2020-02-19
Allow setting compression method and options.

### LINKED ISSUE #1425: Allow setting compression level for gzip
When building large images for a development workflow (as opposed to a release), compressing the layers with gzip can be quite expensive. For example, in a case I just tested, compressing an image took ~50s to produce a ~250MB file. By changing `zip.bzl` to invoke `gzip -1`, this becomes ~20s to produce a ~290MB file.

What do you think about adding a parameter to `container_image` to allow setting the compression level, eg `zip_fast=True` or `gzip_flags="-1"`? Alternatively, it could select automatically based the compilation mode, using `-1` for fastbuild and the default for optimized builds. I'd be happy to send a PR.

drigz:
This makes it easier for users to specify fast compression to reduce
build times. One way to use it is to depend on compilation mode:

```
container_image(
    name = "my-image",
    compression_options = select({
        "@io_bazel_rules_docker//:optimized": [],
        "//conditions:default": ["--fast"],
    })
)
```

The only supported method is "gzip", but in future, zstd could be an
alternative compression method:
https://github.com/opencontainers/image-spec/blob/master/layer.md

zstd(1) also accepts the `--fast` flag: https://www.mankier.com/1/zstd

Fixes #1425.


tests/container/BUILD
```diff
@@ -576,6 +583,16 @@ container_push(
     tags = ["manual"],
 )
 
+container_push(
```
smukherj1: This container push won't actually run in the tests because you need to `bazel run` a container_push target to actually push the image. I suggest creating a `container_test` target instead and look for an expected file. This was we ensure the layer can be loaded correctly.

drigz: The reason I used `container_push` is that other rules (eg `container_test`) use the uncompressed tarball. I verified this by changing the gzip command in `zip.bzl` and running `bazel test -s tests/container:structure_test`, which shows that gzip is not invoked when running `container_test`. Even more surprisingly, running `bazel build tests/container/compression_gzip_fast-layer.tar.gz` fails with `target not declared`, but `bazel build tests/container:new_push_compression_gzip_fast` results in creation of `compression_gzip_fast-layer.tar.gz`.

I've realized that this test is quite useless without checking that the flags affect the image size. I'm thinking of adding a testing that asserts `$(wc -c compression_gzip_normal-layer.tar.gz) -lt $(wc -c compression_gzip_fast-layer.tar.gz)`: where should I add this? Will I need to create new yaml/sh files under testing/e2e/?

smukherj1: Another option is to build two versions of the same image using different compression levels and load the image in the python test [here](https://github.com/bazelbuild/rules_docker/blob/master/container/image_test.py). In this test, you could implement the equivalent check to `$(wc -c compression_gzip_normal-layer.tar.gz) -lt $(wc -c compression_gzip_fast-layer.tar.gz)`.

drigz: Thanks for the tip! PTAL

drigz:
The last two are preexisting: is it OK to resolve the missing docstring by making zip_layer private? Or are users depending on it?

smukherj1:
> The last two are preexisting: is it OK to resolve the missing docstring by making zip_layer private? Or are users depending on it?

I'm not 100% sure but I think the reason the pre-existing issue is being flagged because the buildifier lint only runs on files affected by a PR. I think it would be better to just add a simple docstring like `Compresses the given layer` or something similar instead of trying to conclusively determine if `rules_docker` users depend on it.

drigz:
> I think it would be better to just add a simple docstring like `Compresses the given layer` or something similar instead of trying to conclusively determine if `rules_docker` users depend on it.

Done.

PR #1430, opened 2020-02-21
Plumb os.version through container image config files.

### LINKED ISSUE #1429: Expose a way to specify os.version for images
Similar to https://github.com/bazelbuild/rules_docker/pull/1427 but for os.version.

I'd love for `os.version` information to get propagated down into the config file, like `architecture` and `os`. This would simplify a lot of things re: constructing windows images. There is an existing test [here](https://github.com/bazelbuild/rules_docker/blob/b7abf82341aeacb320b1b81804a046930fc4d421/tests/container/windowsservercore.1803.config.json#L36) but that field gets dropped currently when using rules_docker to import and push.

Related: https://github.com/google/go-containerregistry/pull/676

jonjohnsonjr:
Fixes #1429

Just tested this manually and it seems to work!

I'm mostly just interested in getting os.version to plumb through bazel from the intermediate form config file, which is fixed by just bumping the dependency on go-containerregistry to pick up https://github.com/google/go-containerregistry/pull/676, but it seems like a reasonable thing to also expose in the rules.

I can remove the rules changes and just update the dependency if this feels unnecessary, but exposing this just feels more "complete", y'know? Let me know what you think.

### DIFF
--- a/container/image.bzl
+++ b/container/image.bzl
@@ -107,7 +107,8 @@ def _add_create_image_config_args(
         base_config,
         base_manifest,
         architecture,
-        operating_system):
+        operating_system,
+        os_version):
     """
     Add args for the create_image_config Go binary.
     """
@@ -166,6 +167,9 @@ def _add_create_image_config_args(
     if operating_system:
         args.add("-operatingSystem", operating_system)
 
+    if os_version:
+        args.add("-osVersion", os_version)
+
     if ctx.attr.stamp:
         stamp_inputs = [ctx.info_file, ctx.version_file]
         args.add_all(stamp_inputs, before_each = "-stampInfoFile")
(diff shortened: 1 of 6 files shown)


container/image.bzl
```diff
@@ -551,6 +561,8 @@ _attrs = dicts.add(_layer.attrs, {
     # We need these flags to distinguish them.
     "null_cmd": attr.bool(default = False),
     "null_entrypoint": attr.bool(default = False),
+    "operating_system": attr.string(default = "linux"),
```
smukherj1: This was already being inherited from [here](https://github.com/bazelbuild/rules_docker/blob/834fa017067752b5d7bd908bccbddffbd6687f0b/container/layer.bzl#L250) because of the `_layer.attrs` above.

jonjohnsonjr: Done!

Issue #1384, opened 2020-01-20
`container_run_and_commit` does not support incremental loading.


uri-canva:
`container_run_and_commit` doesn't support incremental loading as output or input. I couldn't find out if it's because it was never implemented, or if it's a fundamental limitation due to having to interact with `docker save`. Even if it was due to `docker save`, it doesn't explain why it doesn't support incremental loading as input, as incremental loading does load the image into docker, so it should be possible to run commands in it?

Can this be clarified in the readme? https://github.com/bazelbuild/rules_docker/blob/b48dddfdad7764c6164701067c6e3736b8e707d7/docker/util/README.md

Related: https://github.com/bazelbuild/rules_docker/issues/1281

nlopezgi:
Hi @uri-canva thanks for creating this issue. I think this is just never been implemented for this rule. We would be happy to accept contributions to either implement this feature for container_run_and_commit or document better the limitation better in the readme.

uri-canva:
I have put up a PR to support incremental loading as input: https://github.com/bazelbuild/rules_docker/pull/1439

Supporting intermediate format as output seems to be more complicated since `docker save` does not support saving specific layers: https://github.com/moby/moby/issues/3870, https://github.com/moby/moby/issues/8039. The feature was implemented twice already https://github.com/moby/moby/pull/5067, https://github.com/moby/moby/pull/9304, but has been rejected both times.

The only API that is layer aware is the registry API: https://docs.docker.com/registry/spec/api/. It is possible to run either https://github.com/docker/distribution configured to use filesystem storage, then symlink / copy the existing layers where it expects them to be, or a simple http server that implements enough of the spec to handle image pushes, that pretends to have the previous layers.

Unfortunately this might require configuring the docker daemon to list localhost as an insecure registry, and hosting a local registry on mac or indows seems to get broken now and again: https://github.com/docker/for-mac/issues/3611 https://github.com/docker/for-win/issues/624 (it's not working for me on mac right now, I tested it by running the docker daemon in a VM).

This all smells pretty bad, I'm considering avoiding `container_run_and_commit` and instead build up images that require running programs inside the container using a combination of `container_run_and_extract` and `container_image` passing the extracted files, but that's not really nice either.

uri-canva:
Can this be reopened? I've updated my PR again: https://github.com/bazelbuild/rules_docker/pull/1439

amir-f:
[...]

PR #1439, opened 2020-02-28
Incremental load support for commit and extract.

uri-canva:
`container_run_and_commit` and `container_run_and_extract` don't support the intermediate format because they were moved from https://github.com/GoogleContainerTools/base-images-docker.

This modifies them to support that.

See https://github.com/bazelbuild/rules_docker/issues/1384.


container/layer_tools.bzl
```diff
@@ -255,10 +286,10 @@ def incremental_load(
...
                 tag_reference,
-                _get_runfile_path(ctx, image["config_digest"]),
+                image["config_digest"].path if action_run else _get_runfile_path(ctx, image["config_digest"]),
             ),
         ]
-        if run:
+        if run or commit or extract:
             # Args are embedded into the image, so omitted here.
             run_statements += [
                 "\"${DOCKER}\" ${DOCKER_FLAGS} run %s %s" % (run_flags, tag_reference),
```
uri-canva: This causes a small difference in behaviour: before the resulting command was `docker run ... $commands`, now it's `docker run ...` and the commands are taken from the image.

uri-canva:
As I'm trying to use this patch in our repo I'm noticing there's a couple of incompatibilities between the way these and the other rules are used that is making the code very finnicky, I will smooth out the difference in a second commit, but the tradeoff for that is that the interface will be very incompatible. It might be worth having that as a separate PR.

uri-canva:
Note this isn't ready to merge because it doesn't update all the other rules in the repo to use the new interface, but I'd like some initial feedback on the approach before continuing with that.

uri-canva:
Yes, that makes sense: use incremental load to just load the image, then run the separate script to do the extracting or committing.

I tried it as my first approach but I couldn't get it to work: templating it so the the two ended up into one script only required a lot of changes, and making it two different scripts was hard to make correct because it required expressing the side effect of having the image loaded as a side effect of the first script, rather than having an output produced as usual.

Let me try again now that I have a working approach, might be able to do something. I didn't know `//container` was supposed to not use `docker`, is that documented somewhere? Maybe there's other things I don't know which I should know.

uri-canva:
Done, now with fewer changes in `container/`.

uri-canva:
@smukherj1 What you do think of this approach?

smukherj1:
This approach looks good! Could you please look into the CI failures (example [here](https://buildkite.com/bazel/rules-docker-docker/builds/3934)

I'm going to launch the GCB tests as well...

HackAttack:
Any update on this? I have a quite heavy image (50 GB of intermediate tarballs) built using the “build image with source layer, run_and_extract” pattern, and if I understand this issue correctly, this PR would eliminate those.

uri-canva:
No this only cuts it in half: it removes the need for the tarballs on the read side, but it still produces tarballs on the write side. Sorry we've been using this patch for a while now and it's working fine, but I haven't had time to clean it up, I'll come back to it.

uri-canva:
Also see my comment in https://github.com/bazelbuild/rules_docker/issues/1384#issuecomment-592789568: it's unlikely we'll ever be able to remove the need for the tarball on the write side since docker does not expose any APIs to output layers, so we've been changing our usage of `docker_rules` to minimise use of `container_run_and_*` rules. It is possible to get very far with `container_image` rules and https://github.com/GoogleContainerTools/distroless/tree/master/package_manager.

uri-canva:
Updated the PR, can you have another look? Note this changes the interface of the rules to be more similar to the one of `container_image`, since they use the same functions underneath. See the changes in the test fixtures for examples.

lugehorsam:
I tried out this PR because I am definitely hoping for some better performance from `run_and_extract`. It looks like the `container_run_and_extract` API has changed -- `no such attribute 'commands'` If so could you update the documentation in the README inside the `util/` folder?

uri-canva:
Good catch, let me do that.

uri-canva:
There's a failure in the tests:

```
ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/debian9/BUILD:46:1: @debian9//:rootfs: no such attribute 'commands' in 'container_run_and_extract' rule
```

This comes from this: https://github.com/GoogleContainerTools/base-images-docker/blob/3eed1bbda3e500f72b36745c9d74385d82ca1b19/debian9/BUILD#L48

Looks like that repo uses this repo and this repo uses that repo, not sure what the process is for releasing backwards incompatible changes: should there be an overlap in versions keeping them backwards compatible? Should I introduce new layer enabled rules and leave the existing ones as is so that dependent repos can switch over bit by bit?

smukherj1:
> There's a failure in the tests:
> 
> ```
> ERROR: /var/lib/buildkite-agent/.cache/bazel/_bazel_buildkite-agent/ec321eb2cc2d0f8f91b676b6d4c66c29/external/debian9/BUILD:46:1: @debian9//:rootfs: no such attribute 'commands' in 'container_run_and_extract' rule
> ```
> 
> This comes from this: https://github.com/GoogleContainerTools/base-images-docker/blob/3eed1bbda3e500f72b36745c9d74385d82ca1b19/debian9/BUILD#L48
> 
> Looks like that repo uses this repo and this repo uses that repo, not sure what the process is for releasing backwards incompatible changes: should there be an overlap in versions keeping them backwards compatible? Should I introduce new layer enabled rules and leave the existing ones as is so that dependent repos can switch over bit by bit?

So the issue here is, `base_images_docker` wouldn't be broken by this change because they [pin](https://github.com/GoogleContainerTools/base-images-docker/blob/0931e30e0e7d18d94bf2ead8405846ff0240d639/WORKSPACE#L34) the version of rules_docker. However, since `base_images_docker` uses the `container_run_*` rules, we would like to ensure they are able to keep using us at `HEAD` which is what the failing test is ensuring. This is to ensure `base_images_docker` can upgrade its rules_docker version at any time if a need arises.

To make this change, I propose:
1. Branch the incremental loader to both the legacy & newer version. Add the ability to use both in the `container_run_*` rules but default to legacy.
2. Update the rules_docker pin in `base_images_docker` and switch them to use the new API of the `container_run_*` rules.
3. Switch the default in `rules_docker` and remove the legacy incremental loader.
4. Clean up `base_images_docker` if necessary.
3.

uri-canva:
Sounds good, since the attributes of the legacy rules and the new rules are so disjoint, I will make the rule accept both sets of attributes, and switch between legacy and new behaviour based on which ones are specified. That should be fully backwards and forwards compatible.

docker/util/run.bzl
```diff
@@ -49,74 +50,50 @@ def _extract_impl(
...
-            "%{image_id_extractor_path}": ctx.executable._extract_image_id.path,
-            "%{image_tar}": image.path,
-            "%{output}": output_file.path,
-        },
-        is_executable = True,
+    image_result = _image.implementation(
+        ctx,
+        name,
+        base = base,
+        cmd = cmd,
```
chickenandpork, line 65: The benefit here is that this allows the build to define an entry point other than `sh -c`.  For example, it allows to simulate a check such as:

`docker run amazon/aws-cli --version`

(The key with this example is that `amazon/aws-cli` has an entry point of `aws` rather than `sh`, so this is running a command `aws --version` )

uri-canva:
Superseded by https://github.com/bazelbuild/rules_docker/pull/1586.

uri-canva:
Looking more carefully #1586 is orthogonal to this, as it still requires building the tar for the image you load. In that sense this PR implements incremental loading for commit and extract rather than incremental writing.

I have updated the PR to be backwards compatible by wrapping both the old and new rule in a macro. I couldn't get it to work with a single rule since `docker_run_flags` exists in both the old rule and `container_image`, and it has different types.

uri-canva:
@jamiees2 I have not updated commit layer to support incremental load, I might do it after this one is merged, or you can do it if you think it makes sense for the use case.

docker/util/run.bzl
uri-canva: This is now backwards compatible so you can keep using `commands`.

jamiees2:
Awesome! I'd love it if we could also add the same support for this in `container_run_and_commit_layer`, but I'm fine with doing that after this is merged to not balloon this PR further. Thanks for this, it seems like it'll solve the other half of our problems :)

uri-canva:
Will fix the failures.

uri-canva:
@edaily Can you set up the CLA for yourself to pass the check?

uri-canva:
`//tests/docker/package_managers:install_pkgs_reproducibility_test` fails but I don't see how the changes in this PR affects it.

uri-canva:
pinging code owners for reviews and help with the test failure ^:
@smukherj1 @alexeagle @pcj @gravypod

uhthomas:
I noticed this in other areas, I don't really understand what's happening. Either the tests are okay and something else is broken, or the tests are broken and only seem to pass because of some cache problem. I suspect the latter as the errors produced make sense, but unrelated to other changes.

@linzhp Is there an issue or PR for said e2e improvements?

gravypod:
So, this is a fun one. We have a circular dep here. The package install tests are attempting to see if we can reproducible install a package [from apt into a container](https://github.com/bazelbuild/rules_docker/blob/master/tests/docker/package_managers/BUILD) using `docker run`. To do this we depend on a package called [ubuntu1604](https://github.com/bazelbuild/rules_docker/blob/094a012750a9b12f3c92634d34cb16f3314ea2c2/WORKSPACE#L241-L246) which is [downloaded from here](https://codeload.github.com/GoogleContainerTools/base-images-docker/tar.gz/01267e68a505f32188553a2706bd7096bd9ea6f5). This package has a bazel workspace in it. It has this BUILD file under `//ubuntu1604`:

```
...
container_image(
    name = "ubuntu1604_vanilla",
    env = UBUNTU_ENV,
    tars = ["@ubuntu1604_tar//file"],
)
...
```

This change modifies the way container_image is implemented so it seems like we have some circular dep thing here. Not sure as to the cause but this is what I'd investigate. 

This is done in this repo: https://github.com/GoogleContainerTools/base-images-docker

Maybe this is the culprit: https://github.com/GoogleContainerTools/base-images-docker/blob/master/WORKSPACE#L33-L38

uri-canva:
TODO: the new code has a trap in `output_logfile` so it cannot be concatenated to `incremental_load.sh` since that has a trap too without handling multiple traps.

uri-canva:
Turns out the test failure might not have been related to the PR at all:
https://github.com/bazelbuild/rules_docker/pull/2068#issuecomment-1107835922=
https://github.com/bazelbuild/rules_docker/pull/2068#issuecomment-1107840086=

but now I can't run the tests locally because of the transitions not working on macOS.

uri-canva:
And we're back, this time `//tests/docker/util:test_extracted_file` is failing: https://buildkite.com/bazel/rules-docker-docker/builds/6193#4331b881-4955-47be-a9cf-f362e2f0f799/228-309.
I tried to run it in a container, with the docker socket mounted, but it couldn't resolve the toolchain through the transitions, can someone help me figure out what's wrong?

laurynaslubys:
> 



> And we're back, this time `//tests/docker/util:test_extracted_file` is failing: https://buildkite.com/bazel/rules-docker-docker/builds/6193#4331b881-4955-47be-a9cf-f362e2f0f799/228-309. I tried to run it in a container, with the docker socket mounted, but it couldn't resolve the toolchain through the transitions, can someone help me figure out what's wrong?

Which version of bazel are you running? This repo seems to not work with 5.x. You can try 4.0.0 as that's the one used in CI.

uri-canva:
Great catch, that was exactly the issue. I was able to run the test locally and it passes, so maybe it's another test caching / flakiness issue on CI?

uri-canva:
> TODO: the new code has a trap in output_logfile so it cannot be concatenated to incremental_load.sh since that has a trap too without handling multiple traps.

Done by removing those traps, the only thing that is leftover in each case is a file created with `mktemp` containing the log output so hopefully not a big deal since it should be removed by the tmp dir cleaner eventually anyway.

uri-canva:
This is the error from the build:
```
bazel-out/k8-fastbuild-ST-6390b2712f3f/bin/tests/docker/util/test_container_extract.build: line 338: fg: no job control
bazel-out/k8-fastbuild-ST-6390b2712f3f/bin/tests/docker/util/test_container_extract.build: line 343: fg: no job control
```
not sure where it's coming from since I couldn't find any related `fg` in the repo, any of the dependencies or bazel itself.

uri-canva:
ping @linzhp @uhthomas

amir-f:
was this merged ? `container_run_and_commit` is really slow on large containers

uhthomas:
Sorry for the lack of activity. This seems like a useful change which I'm sure we'd be happy to add.

PR #1452, opened 2020-03-04
Switch go_deps to use http mode.

uri-canva:
Using `go_repository` in repo mode doesn't have reproducible outputs, which can end up causing unnecessary rebuilds, and using it in module mode interacts with gazelle in the workspace using the rules.

### DIFF
--- a/repositories/go_repositories.bzl
+++ b/repositories/go_repositories.bzl
@@ -37,32 +37,46 @@ def go_deps():
     if "com_github_google_go_containerregistry" not in excludes:
         go_repository(
             name = "com_github_google_go_containerregistry",
-            commit = "221517453cf931400e6607315045445644122692",
             importpath = "github.com/google/go-containerregistry",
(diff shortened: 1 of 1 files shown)


uri-canva:
This PR is now conflicting because the commits were changed, but I'd like to confirm this is a good idea before I rebase it.

alex1545:
Hi @uri-canva and thank you for your PR.
These `go_repository` targets are currently being updated automatically by the Renovate tool (see [PR](https://github.com/bazelbuild/rules_docker/pull/1459) for an example). Can you please confirm that using the module mode will not stop/confuse the automation when these targets become out of date?

uri-canva:
No, unfortunately looking at https://github.com/renovatebot/renovate/blob/53ca91301f6a36c0c87e7d6b06b1bd8c9ec1b283/lib/manager/bazel/extract.ts#L197 and the test cases `go_repository` on renovate only supports commit and tag.

I have filed an issue: https://github.com/renovatebot/renovate/issues/6207

uri-canva:
Superseded by https://github.com/bazelbuild/rules_docker/pull/1619.

PR #1466, opened 2020-03-26
Add option to preserve input tar file mtimes in container_image

dannysullivan:
This is the second step towards resolving static file browser cache issues in application that are deployed using the container_image bazel rule. The first step is [here](https://github.com/bazelbuild/bazel/pull/11028); this PR should **not** be merged until that one has been merged.

This PR allows users to specify that they'd like their container_image to preserve file mtimes from ther input tar file. It is an additive, opt-in change and should not affect any existing functionality.

### DIFF
--- a/container/layer.bzl
+++ b/container/layer.bzl
@@ -284,6 +284,7 @@ _layer_attrs = dicts.add({
         mandatory = False,
     ),
     "portable_mtime": attr.bool(default = False),
+    "preserve_tar_mtimes": attr.bool(default = False),
     "symlinks": attr.string_dict(),
     "tars": attr.label_list(allow_files = tar_filetype),
 }, _hash_tools, _layer_tools)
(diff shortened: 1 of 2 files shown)


container/layer.bzl
```diff
@@ -284,6 +284,7 @@ _layer_attrs = dicts.add({
         mandatory = False,
     ),
     "portable_mtime": attr.bool(default = False),
+    "preserve_tar_mtimes": attr.bool(default = False),
```
smukherj1: I'm not seeing this flag passed on to the build_tar.py binary called [here](https://github.com/bazelbuild/rules_docker/blob/8a924f69ebdc1514a230bbf8b08c0edefbb4af9f/container/layer.bzl#L145). Are you sure this is working as intended? Might make sense to add a test on a container_image rule for this as well. The `container_test` rule allows you to check file timestamps I think

dannysullivan: You're right, thanks for catching that - I missed the lines that copy over that argument while I was copying my changes into my forked git repo. I'll look into adding those tests now.

PR #1467, opened 2020-03-27
Fix puller timeout argument

alexandrvb:


alexandrvb:
@googlebot I signed it!

smukherj1:
Note that a follow up PR like #1365 will be required once this PR is merged

alexandrvb:
@smukherj1 do we really need it? I didn't change any Go code.

PR #1477, opened 2020-04-14
Inject docker headers for container pushes

ngyna:
We use [HTTP headers](https://github.com/docker/cli/blob/master/cli/config/configfile/file.go#L28) in the `~/.docker/config.json` file to inject metadata into our image registry. 

This responsibility seems to be passed on to the image pusher client in `go-containerregistry` [here](https://github.com/google/go-containerregistry/blob/master/cmd/crane/cmd/root.go#L51-L60). We ported this implementation over to `cmd/pusher`.

A new struct was made in order to not add in an additional direct import to `github.com/docker/cli/cli/config`.

### DIFF
--- a/container/go/cmd/pusher/pusher.go
+++ b/container/go/cmd/pusher/pusher.go
@@ -171,23 +171,30 @@ func push(dst string, img v1.Image) error {
 		}
 	}
 
-	file, err := os.Open(path.Join(os.Getenv("DOCKER_CONFIG"), "config.json"))
-	if err != nil {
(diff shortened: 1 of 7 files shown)


ngyna:
@googlebot I signed it!

ImJasonH:
I don't think there's much that go-containerregistry really does beyond calling [`config.Load`](https://godoc.org/github.com/docker/cli/cli/config#Load), which you could also do yourself. Why is a direct dep on `cli/config` unacceptable?

PR #1483, opened 2020-04-27
Use rules_pkg repository instead of deprecated pkg files

dannysullivan:
This repository is currently using a deprecated version of Bazel's package management code. This PR updates all references to `@bazel_tools//tools/build_defs/pkg` so that they instead use `@rules_pkg//`, which is maintained.

Note: this PR does not replace `rules_docker`'s custom version of `build_tar.py` with the one contained in `rules_pkg`. It should not result in any changed behavior at all.

### DIFF
--- a/README.md
+++ b/README.md
@@ -691,7 +691,7 @@ rules, you can override the default `base="..."` attribute.  Consider this
 modified sample from the `distroless` repository:
 
 ```python
-load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
+load("@rules_pkg//:pkg.bzl", "pkg_tar")
 
 # Create a passwd file with a root and nonroot user and uid.
 passwd_entry(
--- a/container/BUILD
+++ b/container/BUILD
@@ -44,7 +44,7 @@ py_library(
     visibility = ["//visibility:public"],
     deps = [
         "@bazel_tools//third_party/py/gflags",
-        "@rules_pkg//pkg:archive",
+        "@rules_pkg//:archive",
     ],
 )
 
--- a/container/BUILD
+++ b/container/BUILD
@@ -44,7 +44,7 @@ py_library(
     visibility = ["//visibility:public"],
     deps = [
         "@bazel_tools//third_party/py/gflags",
-        "@bazel_tools//tools/build_defs/pkg:archive",
+        "@rules_pkg//pkg:archive",
     ],
 )
 
--- a/container/build_tar.py
+++ b/container/build_tar.py
@@ -25,7 +25,7 @@ import re
 import tarfile
 import tempfile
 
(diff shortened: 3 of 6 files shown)


dannysullivan:
Closing temporarily pending more testing.

container/image_test.py
```diff
@@ -91,7 +91,7 @@ def test_files_in_layer_with_file_base(self):
 
   def test_tar_base(self):
     with TestImage('tar_base') as img:
-      self.assertDigest(img, '611270a900f1d95f896af3d8c38ff04f75d75f972c02eab0403aa12847881489')
+      self.assertDigest(img, '9fdc5d975efcc73690089a2eedaa28557ce7a792baa1fcd5e56949fb637c38f2')
```
smukherj1, line 94: What's the diff in the new image that results in the digest mismatch?

container/image_test.py
```diff
@@ -216,48 +216,6 @@ def test_with_rfc_3339_creation_time(self):
       cfg = json.loads(img.config_file())
       self.assertEqual('1989-05-03T12:58:12.345Z', cfg.get('created', ''))
 
-  # This test is flaky. If it fails, do a bazel clean --expunge_async and try again
-  def test_with_stamped_creation_time(self):
```
smukherj1, line 220: Why were these tests deleted?

dannysullivan: The primary reason for making this PR is that the TarFileWriter in rules_pkg now preserves the mtimes on files in input tars by default (instead of overwriting them to different values based on the presence of `--stamp`), so the default behavior described in these tests no longer applies, if I'm understanding them correctly. @smukherj1 has some extra context on this.

container/image_test.py
dannysullivan: Per my other comment, the version of TarFileWriter in rules_pkg defaults to preserving input tar file mtimes. I have verified that these digests have the original values if I turn off that functionality, so that seems to be the only change. If you'd prefer, I'm open to splitting this into two PRs: one that introduces rules_pkg as a dependency with the new functionality turned off, and then another that turns the mtime-preserving functionality on.

container/image_test.py
smukherj1: These tests validate the image creation time populated in the image config. See [here](https://github.com/bazelbuild/rules_docker/blob/634ff5b5678e0db01baef8fadc9bdb2493193aee/container/image.bzl#L235). This should be unaffected by this change because the image config is created by the Go binary [here](https://github.com/bazelbuild/rules_docker/blob/634ff5b5678e0db01baef8fadc9bdb2493193aee/container/go/cmd/create_image_config/create_image_config.go), not the layer tarball generator that your change affects. Can you look into why the creation time of the image is changing as well?

PR #1503, opened 2020-05-18
fix(nodejs): remove node binary from layers

thesayyn:
Platform-specific nodejs binaries already come with nodejs_binary so we don't need to put them into another layer again. Also, these binaries do not work with --platforms thus when we build an image from OSX while targeting Linux, we get an image that contains both nodejs_darwin_amd64 and nodejs_linux_amd64 binaries.

### DIFF
--- a/nodejs/image.bzl
+++ b/nodejs/image.bzl
@@ -136,14 +136,6 @@ def nodejs_image(
             **kwargs
         )
 
-    nodejs_layers = [
-        # Put the Node binary into its own layers.
-        "@nodejs//:node",
-        "@nodejs//:node_files",
-        "@nodejs//:bin/node_repo_args.sh",
-    ]
-
-    all_layers = nodejs_layers + layers
 
     visibility = kwargs.get("visibility", None)
     tags = kwargs.get("tags", None)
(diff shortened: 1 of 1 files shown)


alex1545:
Hi @thesayyn and thank you for your PR.
Can you please run buildifier on the modified file as well as fix the failing `//container:image_test` test target that fails [here](https://github.com/bazelbuild/rules_docker/blob/master/container/image_test.py#L774).

Toxicable:
@thesayyn Those node layers are meant to point to the platform specific ones, they're only aliases. If they don't work correctly then I think it's an issue with rules_nodejs not the integration here.

What happens if you move nodejs up to the top layer is that every time you make a code change will increase the size of that layer by 60MB~. So your incremental images will now always be 62MB~ instead of the current behaviour where it's like 1-2MB

thesayyn:
The problem is not about the image size but sure it is a problem on slow networks. Also, in some environments such as circleci there is no layer caching which makes this a problem.

I am not sure if this is a problem of rules_nodejs though.

Toxicable:
The size issue isn't just for slow networks, there's many other places where this would cause a regression:
- The disk storage size on repositories, most notibly private repositories.
- The update times of deployed applications
- The rebuild size of systems that are incremental (There's many other CI systems then circleci)
- Cache hits on a shared remote cache
- Storage size on workstations

There's probably more that I cant think of, but generally we just don't want larger and less incremental layers, I'd very much take that as a regression

PR #1506, opened 2020-05-20
[Alt1] Exposing external base image properties (as tags)

or-shachar:
Greetings!

In my company we build docker containers using bazel but we publish to docker registry using external process that digests bazel outputs and pushes it to the remote.

For that we need to know exactly what layers construct the image we want to publish and what was the root external base image (given by registry,repository and digest)

Problem is - using `container_pull` generates a `container_import` target that doesn't have this data at all.

There are (at least) two ways `container_import` can expose this properties:
1. By adding tags with predefined prefix
2. By adding more attributes to `container_import` and exposing it in the provider.

In this PR - I chose option (1).

This PR makes sure that the generated `container_import` would have useful tags. like so:
```python
container_import(
    name = "image",
    config = "config.json",
    layers = glob(["*.tar.gz"]),
    tags = [
        "dockerRegistry=gcr.io",
        "dockerRepository=distroless/java",
        "dockerDigest=sha256:9d4092ba5e1c9dc4d1490fdead1dd7ea5c64e635b729fee11a6af55f51b337f8"
        ],
)
```

Example of how I can use this data in order to expose the layers metadata [can be found here](https://github.com/or-shachar/export_base_image_example). There I use aspect (similar to like [bazel-common's pom_file](https://github.com/google/bazel-common/blob/f1115e0f777f08c3cdb115526c4e663005bec69b/tools/maven/pom_file.bzl#L177)) - to traverse until the root base image and fetch from it the base image properties using the tags.

This might be good for merging even, but I'd love to hear your thoughts? 
Also - do you test such features somehow? Am I supposed to add some failing tests?

### DIFF
--- a/container/pull.bzl
+++ b/container/pull.bzl
@@ -17,6 +17,11 @@ This wraps the rulesdocker.go.cmd.puller.puller executable in a
 Bazel rule for downloading base images without a Docker client to
 construct new images.
 """
+pull_tags_prefix = struct(
+    registry = "dockerRegistry=",
+    repository = "dockerRepository=",
+    digest = "dockerDigest=",
+)
 
 _container_pull_attrs = {
     "architecture": attr.string(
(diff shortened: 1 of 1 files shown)


smukherj1:
Hi @or-shachar,

Thanks for the PR. I see the value in what you're proposing. I'm curious what lead you to choose tags over updating the provider? To me, the provider approach seems more type safe.

or-shachar:
closing in favor of #1507

PR #1508, opened 2020-05-21
Add boiler plate to begin migrating image_test to Go

smukherj1:


zoidbergwill:
Would it be useful for someone else to pick this up?

PR #1513, opened 2020-05-26
Fix sha256 for bazel-release.pub.gpg

meteorcloudy:
The Bazel GPG key was renewed to extend it's validity period.


meteorcloudy:
This will fix recent build error on Bazel CI:
https://buildkite.com/bazel/rules-docker-docker/builds/4156

PR #1509, opened 2020-05-24
Updated go-containerregistry version

or-shachar:
We want to support this import fix: https://github.com/google/go-containerregistry/issues/717

PR #1517, opened 2020-05-30
update puller and loader binaries to pich up #1509

or-shachar:


or-shachar:
Thanks @alex1545  - 
Any chance we can create a new release of rules_docker based on it?

alex1545:
Hi @or-shachar,
Release 0.14.3 is now available.

PR #1526, opened 2020-06-05
Populate list of instrumented files for coverage in container_image_

tilmangiese:


smukherj1:
Travis CI passed [link](https://travis-ci.org/github/bazelbuild/rules_docker/builds/695072752) but the webhook was dropped. Going to ignore the check here.

Issue #1533, opened 2020-06-15
Depend on gflags from WORKSPACE rather than @bazel_tools


aiuto:
https://github.com/bazelbuild/rules_docker/blob/f7e6d6f28ac9f9e92d8b875426233031c5c5257c/container/BUILD#L46
depends on   "@bazel_tools//third_party/py/gflags",

Bazel has removed gflags because it is not used any more.

aiuto:
Or... just replace the single use of gflags with the python flag library.  It is only container/build_tar

meteorcloudy:
@aiuto Can you help file bugs for other projects failing in downstream due to this issue to upgrade their rules_docker version?
https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1531

aiuto:
Maybe over the weekend. 
It is not sustainable for us to monitor and warn on everything. Don't the downstream projects themselves have an obligation to monitor this CI we run for the?
We should certainly be enforcing that for the rule sets. For projects (like robotics) that is less clear.

meteorcloudy:
@aiuto 
> Don't the downstream projects themselves have an obligation to monitor this CI we run for the?

I don't think all third party projects should monitor our Downstream Pipeline, because it could be broken by Bazel bugs. It make more sense for them to monitor our "Bazel@Releasse + Incompatible flags" pipeline, but this change is not guarded by a flag. Even though, the current situation is the author of every incompatible change should drive the downstream projects migration, like filing bugs.

meteorcloudy:
@aiuto Can you make sure there is a paragraph in the next Bazel release note that mentions the removal of gflags and the necessary to upgrade rules_docker to fix the build error?

meteorcloudy:
@smukherj1 When is the next release of rules_docker happening? Should the projects depending on rules_docker wait for the new release or just ping rules_docker to the latest commit?

aiuto:
Sure. I created a milestone for the next major release. That will be the place to look.

smukherj1:
> @smukherj1 When is the next release of rules_docker happening? Should the projects depending on rules_docker wait for the new release or just ping rules_docker to the latest commit?

We don't really have a release schedule. We only create a release if specifically requested. If pinning to the latest commit is possible, downstream projects are free to do so.

PR #1535, opened 2020-06-16
Revert to using gflags but use it from a pips package instead of Bazel Tools

smukherj1:
Next version of Bazel is expected to stop providing gflags via Bazel tools according to #1533

### DIFF
--- a/container/BUILD
+++ b/container/BUILD
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 load("@bazel_skylib//:bzl_library.bzl", "bzl_library")
+load("@pip_deps//:requirements.bzl", "requirement")
 
 package(default_visibility = ["//visibility:public"])
 
@@ -43,7 +44,7 @@ py_library(
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [
-        "@io_abseil_py//absl/flags",
+        requirement("python-gflags"),
         "@rules_pkg//:archive",
     ],
 )
--- a/container/BUILD
+++ b/container/BUILD
@@ -45,6 +45,7 @@ py_library(
     visibility = ["//visibility:public"],
     deps = [
         requirement("python-gflags"),
+        requirement("six"),
         "@rules_pkg//:archive",
     ],
 )
--- a/container/BUILD
+++ b/container/BUILD
@@ -43,7 +43,7 @@ py_library(
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
     deps = [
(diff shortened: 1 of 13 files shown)


mmorearty:
@smukherj1 should `container_repositories()` be updated to load `pip_deps`?

The way things are now, after this change, my own project gets errors like this:

```
ERROR: .../BUILD:45:12: every rule of type container_image_ implicitly depends upon the target '@io_bazel_rules_docker//container:build_tar', but this target could not be found because of: error loading package '@io_bazel_rules_docker//container': Unable to find package for @pip_deps//:requirements.bzl: The repository '@pip_deps' could not be resolved.
```

smukherj1:
> @smukherj1 should `container_repositories()` be updated to load `pip_deps`?
> 
> The way things are now, after this change, my own project gets errors like this:
> 
> ```
> ERROR: .../BUILD:45:12: every rule of type container_image_ implicitly depends upon the target '@io_bazel_rules_docker//container:build_tar', but this target could not be found because of: error loading package '@io_bazel_rules_docker//container': Unable to find package for @pip_deps//:requirements.bzl: The repository '@pip_deps' could not be resolved.
> ```

@mmorearty Due to a quirk of how Bazel works, `container_repositories()` can't be updated to run `pip_deps` because `load` statements in Bazel can only be defined at the top level and the `pip_deps()` isn't defined until the `container_repositories()` macro executes. Instead, consider updating your WORKSPACE to do the following additional step from rules_docker's [setup](https://github.com/bazelbuild/rules_docker#setup)

```
load("@io_bazel_rules_docker//repositories:deps.bzl", container_deps = "deps")

container_deps()
```

mmorearty:
Ah, thanks! I didn't notice that `container_repositories()` and `container_deps()` were separate things, and sure enough my repo was missing the call to `container_deps()`.

whilp:
See related report in https://github.com/bazelbuild/rules_docker/pull/1535

alexeagle:
@smukherj1 Hey, I think this should have qualified as a breaking change and not rolled out in a patch release. This broke rules_nodejs examples as well, and I don't think these users should need to change their code (WORKSPACE) to accomodate some internal implementation detail released as a bugfix. Should have been a 0.15.
Do you have tests in rules_docker than told you this was a breaking change? I've been adding such integration tests (like recently to rules_python) and would be happy to help.
Also rather than using pip, how about just distributing the couple of needed .whl files vendored into your distro, so you don't introduce a tangle of WORKSPACE dependencies? Ideally IMO a core ruleset like docker should have ~0 dependencies.

PR #1540, opened 2020-06-24
fix launcher args not being passed correctly

griffin:
the launcher arguments were not being passed correctly to the `create_image_config` binary, it seems to be an issue with the cli flag parser. 

i was passing arguments like so, when i noticed that the arguments were not propagated to the entrypoint:
```
    container_image(
        name = name,
        base = go_image_name,
        launcher = "//rules:dlv",
        launcher_args = [
            "exec",
            "--continue",
            "--headless",
            "--listen=:2345",
            "--accept-multiclient",
        ],
    )
```

the fix seems to line up with how the `list<string>` type is passed elsewhere to this program.

additionally, there seems to be an e2e test to ensure that this functionality works, but it doesn't seem to be enabled for CI (please correct me if i'm wrong). i'd be happy to enable it.

### DIFF
--- a/container/image.bzl
+++ b/container/image.bzl
@@ -179,7 +179,7 @@ def _add_create_image_config_args(
         fail("launcher_args does nothing when launcher is not specified.", attr = "launcher_args")
     if ctx.attr.launcher:
         args.add("-entrypointPrefix", ctx.file.launcher.basename, format = "/%s")
-        args.add_all(ctx.attr.launcher_args)
+        args.add_all(ctx.attr.launcher_args, before_each="-entrypointPrefix")
 
 def _format_legacy_label(t):
     return ("--labels=%s=%s" % (t[0], t[1]))


griffin:
@googlebot I fixed it.

griffin:
i did some more digging and it looks like the test was passing due to an overlap in the flag name. both the [create_image_config](https://github.com/bazelbuild/rules_docker/blob/master/container/go/cmd/create_image_config/create_image_config.go#L61) and the [test binary](https://github.com/bazelbuild/rules_docker/blob/master/testdata/launcher_main.go#L27) used `env` to specify a variable. this caused the test to still pass, but for the wrong reason. 

this flag was being picked up by create_image_config and set without error in the image manifest. i've confirmed this using `docker image inspect` to see the env variable set and the entrypoint missing the provided arguments

i've updated the test to use a flag that does not overlap, and it now passes locally for me.

griffin:
@smukherj1 is there anything outstanding that i need to do?

smukherj1:
Oops need to merge master into the branch for this to auto merge. I'll do that now. Sorry!

smukherj1 (approved):

griffin:
no problem! thanks again!

meyerhot95:
@griffin is `"//rules:dlv"` your own custom delve binary rule?

PR #1577, opened 2020-08-02
Add container_layer_prune to remove files from container_layer. Fixes #7.

### LINKED ISSUE #7: Feature request: support file deletion on docker_build
From @justinsb:

> Could we / should we add the ability to delete files from a docker layer as part of docker_build?
> 
> The primary use case I am thinking of is that we install a deb package, but then we remove e.g. the man pages or the locale files, so as to keep the image size small.

Original Issue: https://github.com/bazelbuild/bazel/issues/1335

huan086:


huan086:
@googlebot I signed it!

smukherj1:
Thanks for the PR! Hoping to review this tomorrow

huan086:
Added documentation today. Hope I didn't miss anything else.

container/prune_tar.py
```diff
@@ -0,0 +1,97 @@
...
+  def __enter__(self):
+    self.tarfile = archive.TarFileWriter(self.output)
+    return self
+
+  def __exit__(self, t, v, traceback):
+    self.tarfile.close()
+
+  def name_filter(self, path):
+    """Callback for archive.TarFileWriter.add_tar that returns false
+    if the file or directory in remove_paths.
```
smukherj1, line 54: Suggest changing this comment to describe the scenario when the function returns false, so something like:

returns false when the given path should be removed from the layer.

container/prune_tar.py
```diff
@@ -0,0 +1,97 @@
...
+    """
+    if path.startswith(self.root_directory + '/'):
+      path = path[len(self.root_directory):]
+
+    for remove_path in self.remove_paths:
+      if remove_path.endswith('/'):
+        if path.startswith(remove_path):
+          return False
+      elif path == remove_path or path.startswith(remove_path + '/'):
+        return False
```
smukherj1, line 67: Wouldn't just the following cover all cases?

```
if path.startswith(remove_path):
   return False
```

container/prune_tar.py
```diff
@@ -0,0 +1,97 @@
...
+    including its contents, is removed.
+    """
+    self.tarfile.add_tar(self.tar, name_filter=self.name_filter)
+
+
+def main(unused_argv):
+  if FLAGS.manifest:
+    with open(FLAGS.manifest, 'r') as f:
+      manifest = json.load(f)
+      remove_paths = manifest.get('remove_paths', [])
```
smukherj1, line 87: I'm assuming many users might also want to use glob here. It might be better to support that instead of the slightly non-obvious semantic currently supported, i.e. if the path ends in "/" it deletes the directory contents but preserves the directory itself.

Supporting glob should be trivial with https://docs.python.org/3/library/fnmatch.html#module-fnmatch I think?

huan086: Great idea. Let me find a way to support glob. fnmatch doesn't seem to be the right way because

```
Note that the filename separator ('/' on Unix) is not special to this module. See module glob for pathname expansion (glob uses filter() to match pathname segments). Similarly, filenames starting with a period are not special for this module, and are matched by the * and ? patterns.
```

container/prune_tar.py
huan086: Assuming there is the file and folder `/etc/someapp` and `etc/someapp.d` and I want to remove `etc/someapp` only, the simplification would remove too much.

Let me find a way to implement glob so that I don't need this custom logic

huan086:
Ready to be reviewed

huan086:
@smukherj1 I noticed that the distroless java image has a layer with tar structure I don't understand and is not handled in this code. The first few entries of the tarball is

```sh
drwxr-xr-x 0/0               0 1970-01-01 07:30 ./
tar: Removing leading `/' from member names
drwxr-xr-x 0/0               0 1970-01-01 07:30 /usr/
drwxr-xr-x 0/0               0 1970-01-01 07:30 /usr/bin/
lrwxr-xr-x 0/0               0 1970-01-01 07:30 /usr/bin/java -> /usr/lib/jvm/java-11-openjdk-amd64/bin/java
drwxr-xr-x 0/0               0 1970-01-01 07:30 ./var/
drwxr-xr-x 0/0               0 1970-01-01 07:30 ./var/lib/
```

How can I handle this strange entry using the `archive.py`?

`tar: Removing leading `/' from member names`

huan086:
Removed dependency on archive.py. Passes through any tarinfo that doesn't start with `./` for Linux or `Files/` for Windows.

Please review.

huan086:
I realized I've written a generic tar manipulation rule that doesn't need to be specific to docker. And I have other tar manipulation ideas in mind. Maybe I'll start another repo

Issue #1404, opened 2020-02-01
Clashing of Python and Docker workspace rules


mvukov:
Hi, I am working with Bazel 1.2.1 and 2.0.

In my WORKSPACE file I have:
```
http_archive(
    name = "io_bazel_rules_docker",
    sha256 = "dc97fccceacd4c6be14e800b2a00693d5e8d07f69ee187babfd04a80a9f8e250",
    strip_prefix = "rules_docker-0.14.1",
    urls = ["https://github.com/bazelbuild/rules_docker/releases/download/v0.14.1/rules_docker-v0.14.1.tar.gz"],
)

http_archive(
    name = "rules_python",
    url = "https://github.com/bazelbuild/rules_python/archive/94677401bc56ed5d756f50b441a6a5c7f735a6d4.tar.gz",
    strip_prefix = "rules_python-94677401bc56ed5d756f50b441a6a5c7f735a6d4",
    sha256 = "acbd018f11355ead06b250b352e59824fbb9e77f4874d250d230138231182c1c",
)

# ...

# Python setup

load("@rules_python//python:repositories.bzl", "py_repositories")
load("@rules_python//python:pip.bzl", "pip_repositories")
load("@rules_python//python:pip.bzl", "pip3_import")
py_repositories()
pip_repositories()
pip3_import(
   name = "pip_deps",
   requirements = "//:requirements.txt",
)
load("@pip_deps//:requirements.bzl", "pip_install")
pip_install()

# Docker setup
load(
    "@io_bazel_rules_docker//repositories:repositories.bzl",
    container_repositories = "repositories",
)
container_repositories()

load(
    "@io_bazel_rules_docker//cc:image.bzl",
    _cc_image_repos = "repositories",
)

_cc_image_repos()

load("@io_bazel_rules_docker//repositories:deps.bzl", container_deps = "deps")
container_deps()

# ...

```

This setup works. At the moment I only build Docker images with cc binaries.

If I reorder Python setup and Docker setup, Bazel picks up some other requirements.txt file --  I guess from the Docker setup, that one is rather small, includes PyYAML etc. All in all, I cannot run some of my Python targets.

I will try to alias Python command imports, but wanted to let you know of the issue.

mmikitka:
I am encountering the same problem, and the issue is that you have a `@pip_deps` repository target in your WORKSPACE that conflicts with the same `@pip_deps` repository target in rules_docker.

[...]

PR #1582, opened 2020-08-07
Replace dependency on gflags with argparse

robbertvanginkel:
Since https://github.com/bazelbuild/rules_docker/pull/1535, rules_docker depends on a global `@pip_deps` external repo and pip to install the python gflags library. This easily conflicts with other python rules that claim that name (https://github.com/bazelbuild/rules_docker/issues/1404) and forces a dependency on pypi, making rules_docker harder to depend upon. I'd prefer not to have to setup a pypi mirror just to be able to reliably use rules_docker.

This seems pretty painful for importing a flag library, the comment in https://github.com/bazelbuild/rules_docker/issues/1533#issuecomment-644153801 suggests
> Or... just replace the single use of gflags with the python flag library. It is only container/build_tar

Which this PR attempts to implement.

### DIFF
--- a/container/BUILD
+++ b/container/BUILD
@@ -46,10 +46,6 @@ py_library(
     ],
     srcs_version = "PY2AND3",
     visibility = ["//visibility:public"],
-    deps = [
-        requirement("python-gflags"),
-        requirement("six"),
-    ],
 )
 
 TEST_TARGETS = [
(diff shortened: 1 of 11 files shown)


robbertvanginkel:
Had a small mistranslation for replacing the multistring, should be fixed!

robbertvanginkel:
Thanks!

or-shachar:
This is great :) Solved a regression for Wix when [upgrading to bazel 3.5.0](https://github.com/bazelbuild/bazel/issues/11885)
Can we create a matching release?

PR #1596, opened 2020-08-26
Update shas

cgdolan:
The digest tests under //tests/container on darwin (OpenJDK8) have been failing for the scheduled builds on master since Friday: https://buildkite.com/bazel/rules-docker-docker/builds?branch=master

### DIFF
--- a/container/archive.py
+++ b/container/archive.py
@@ -140,7 +140,12 @@ class TarFileWriter(object):
       # Instead, we manually re-implement gzopen from tarfile.py and set mtime.
       self.fileobj = gzip.GzipFile(
           filename=name, mode='w', compresslevel=9, mtime=self.default_mtime)
-    self.tar = tarfile.open(name=name, mode=mode, fileobj=self.fileobj)
+    self.tar = tarfile.open(
+      name=name,
+      mode=mode,
+      fileobj=self.fileobj,
+      # https://github.com/bazelbuild/rules_docker/issues/1602
+      format=tarfile.PAX_FORMAT)
     self.members = set([])
     self.directories = set([])
 
--- a/container/image_test.py
+++ b/container/image_test.py
@@ -77,19 +77,19 @@ class ImageTest(unittest.TestCase):
 
   def test_files_base(self):
     with TestImage('files_base') as img:
-      self.assertDigest(img, 'b2042de8d0d7f2cd89328f22ba4f9e4884d1ea7385e3b1ec8ae686c1cf8377de')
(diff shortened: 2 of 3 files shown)


cgdolan:
Guess I should have seen that coming - tests now pass on darwin (OpenJDK8) and fail on all others.

cgdolan:
Reopening. From #1602 its likely this change is good and the test failures on non-Darwin platforms are due to a yet-to-be diagnosed caching issue.

cgdolan:
Once CI caching is disabled, the digest tests on non-Darwin platforms should pass.

/assign @smukherj1

cgdolan:
Sure thing, updated .bazelci as requested.

smukherj1:
> Sure thing, updated .bazelci as requested.

Sigh, still getting cache hits. Ok last idea, also add `--noremote_accept_cached` to the test_flags sections?

smukherj1:
Could you try adding a line [here](https://github.com/bazelbuild/rules_docker/blob/v0.14.4/container/build_tar.py#L30) that does `tarfile.DEFAULT_FORMAT = tarfile.PAX_FORMAT` and revert the flag changes to the build kite config? This *might* get the Linux and Mac CI digests to agree again

smukherj1:
(╯°□°)╯︵ ┻━┻

smukherj1:
Ok I think a better way to specify the tar format might be to edit the line [here](https://github.com/bazelbuild/rules_docker/blob/2ae5f7f0c43b7efd32a4d7f927bb701f060972af/container/archive.py#L143) to

```
self.tar = tarfile.open(name=name, mode=mode, fileobj=self.fileobj, format=tarfile.PAX_FORMAT)
```

as documented [here](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile)

I expect global import variable freezing meant the DEFAULT_FORMAT was already frozen for the archive package before we overrode it manually. Can't be messing around with python import machinery too much 😅

cgdolan:
> Ok I think a better way to specify the tar format might be to edit the line [here](https://github.com/bazelbuild/rules_docker/blob/2ae5f7f0c43b7efd32a4d7f927bb701f060972af/container/archive.py#L143) to
> 
> ```
> self.tar = tarfile.open(name=name, mode=mode, fileobj=self.fileobj, format=tarfile.PAX_FORMAT)
> ```
> 
> as documented [here](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile)
> 
> I expect global import variable freezing meant the DEFAULT_FORMAT was already frozen for the archive package before we overrode it manually. Can't be messing around with python import machinery too much



> (╯°□°)╯︵ ┻━┻

We'll get there! One step at a time.

PR #1600, opened 2020-08-26
Simplify reading the digest file obtained when pulling a Docker image

faube:
repository_ctx.read exists specifically to read the content of a file.
`cat` is not available on all platforms (notably Windows).

### DIFF
--- a/container/pull.bzl
+++ b/container/pull.bzl
@@ -176,10 +176,7 @@ def _impl(repository_ctx):
     }
     updated_attrs["name"] = repository_ctx.name
 
-    digest_result = repository_ctx.execute(["cat", repository_ctx.path("image/digest")])
-    if digest_result.return_code:
-        fail("Failed to read digest: %s" % digest_result.stderr)
-    updated_attrs["digest"] = digest_result.stdout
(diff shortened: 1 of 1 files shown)


smukherj1:
Sigh, looks like `//tests/container:alpine_custom_attr_digest_test` is broken at HEAD. Will need to take a look

smukherj1:
I was looking into the failures myself but I noticed @alex1545 is also working on it in #1597. I'll sync with @alex1545 on the debugging status.

I was able to reproduce the failure locally on a Linux machine for the target that's failing on Mac which is confusing because in the CI it *passes* on Linux but fails on Mac. Is the CI caching a stale result?

smukherj1:
Skipping the check for this PR because I see the puller works as expected and the pulled image has an unchanged digest. However, more involved changes to rules_docker are blocked until we can root cause why remote caching isn't working for the failing tests :(

Basically, something's changed that causing the Mac CI (which runs Bazel with local execution) to produce the test image with a changed digest but the Linux CI (which runs Bazel with remote execution) to get a cache hit. I'm able to reproduce the changed digest on Linux on my machine so I suspect somewhere in rules_docker we're failing to specify an input to an action resulting in an erroneous remote cache hit.

PR #1609, opened 2020-08-31
Fix the root directory when building for Linux on Windows.

faube:
Otherwise, the .tar file gets a weird structure that is not well interpreted in Linux; e.g. files get names like ".\a\b\c\file" (that is a filename, not a path).

### DIFF
--- a/container/layer.bzl
+++ b/container/layer.bzl
@@ -125,6 +125,9 @@ def build_layer(
     if (operating_system == "windows"):
         args.add("--root_directory=Files")
         empty_root_dirs = ["Files", "Hives"]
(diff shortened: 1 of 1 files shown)


faube:
Please let me know if there is a better way to know if the Bazel host is Windows; I could not find a consistent approach elsewhere in rules_docker.

smukherj1 (approved):

faube:
> New changes are detected. LGTM label has been removed.

Oops, I may have clicked the wrong button. I was seeing a message saying this branch was out of date with the master.

What is the next step for this Pull Request?

smukherj1 (approved):

smukherj1 (approved):

PR #1611, opened 2020-09-01
Remove --privileged flag in package_manager rules

samschlegel:
This shouldn't be needed, is a security risk, and conflicts with userns remapping.

I'd like to test compatibility with userns remapping, but it requires restarting the docker daemon with a different configuration and it's unclear to me the best way to do that


samschlegel:
re: testing userns_remap compatibility, it seems like I could update travis to run with userns-remap. Not sure if we want to always run with that, or run the e2e twice: with it off, and with it on

PR #1612, opened 2020-09-02
For gzip operations, use a new Go helper by default instead of which(gzip)

faube:
This is mostly for consistency across platforms, and to remove an external dependency.


faube:
Forgot to address tests.

PR #1613, opened 2020-09-02
For gzip operations, use a new Go helper by default instead of which(gzip)

faube:
This is mostly for consistency across platforms, and to remove an external dependency.

### DIFF
--- a/container/go/cmd/zipper/BUILD
+++ b/container/go/cmd/zipper/BUILD
@@ -1,0 +1,34 @@
+# Copyright 2020 The Bazel Authors. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ###
+# Build file for zipper binary.
+
+load("@bazel_gazelle//:def.bzl", "gazelle")
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+# gazelle:prefix github.com/bazelbuild/rules_docker
+gazelle(name = "gazelle")
+
+go_binary(
+    name = "zipper",
+    embed = [":go_default_library"],
+    visibility = ["//visibility:public"],
+)
+
+go_library(
+    name = "go_default_library",
+    srcs = ["zipper.go"],
+    importpath = "github.com/bazelbuild/rules_docker",
+    visibility = ["//visibility:private"],
+)
--- a/container/go/cmd/zipper/BUILD
+++ b/container/go/cmd/zipper/BUILD
@@ -14,11 +14,7 @@
 # ###
 # Build file for zipper binary.
 
-load("@bazel_gazelle//:def.bzl", "gazelle")
 load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
-
-# gazelle:prefix github.com/bazelbuild/rules_docker
-gazelle(name = "gazelle")
 
 go_binary(
     name = "zipper",
--- a/container/go/cmd/zipper/zipper.go
+++ b/container/go/cmd/zipper/zipper.go
@@ -1,0 +1,81 @@
+// Copyright 2020 The Bazel Authors. All rights reserved.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+////////////////////////////////////////////////
+// This package performs gzip operations.
+
+package main
+
+import (
+	"bytes"
+	"compress/gzip"
+	"flag"
+	"io/ioutil"
+	"log"
+	"os"
+)
+
+var (
+	src        = flag.String("src", "", "The source location of the file to zip/unzip.")
+	dst        = flag.String("dst", "", "The destination location of the file, after zip/unzip.")
+	decompress = flag.Bool("decompress", false, "If true, perform gunzip. If false, gzip.")
+	fast       = flag.Bool("fast", false, "If true, use the fastest compression level.")
+)
+
+func main() {
+	flag.Parse()
+
+	if *src == "" {
+		log.Fatalln("Required option -src was not specified.")
+	}
+
+	if *dst == "" {
+		log.Fatalln("Required option -dst was not specified.")
+	}
+
+	dat, err := ioutil.ReadFile(*src)
+	if err != nil {
+		log.Fatalf("Unable to read input file: %v", err)
+	}
+	var buf bytes.Buffer
+	if *decompress {
+		zr, err := gzip.NewReader(bytes.NewReader(dat))
+		if err != nil {
+			log.Fatalf("Unable to read: %v", err)
+		}
+		if _, err := buf.ReadFrom(zr); err != nil {
+			log.Fatalf("Unable to decompress: %v", err)
+		}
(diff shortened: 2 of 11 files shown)


skylib/zip.bzl
```diff
@@ -60,22 +82,38 @@ def gunzip(ctx, artifact):
...
-        input_manifests = input_manifests,
-        inputs = [artifact],
-        outputs = [out],
-        use_default_shell_env = True,
-        mnemonic = "GUNZIP",
-        tools = tools,
-    )
+    gzip_path = _gzip_path(toolchain_info)
+
+    if gzip_path == "":
```
smukherj1, line 87: Given this if/else condition has a few subtleties, I suggest refactoring it into a function and re-using it for both gzip & gunzip and pass in parameters that change (I think there should only be 2-3 such parameters?)

Regarding the condition itself, IIUC what we want to do is the following in order of precedence:
1. If neither gzip_path nor gzip_target is specified, use internal tool
2. else if gzip_path is specified, use the gzip tool at the specified absolute path
3. else if gzip_target is specified, build the specified target and use it assuming it accepts the same flags as gzip

There's a possibility `if gzip_path == ""` at the beginning might break this because, if gzip_target is specified, gzip_path might be blank (or maybe it's None?). If gzip_path defaults to None instead of "" when gzip_target is set, your logic still works but given this is very subtle, I suggest commenting exactly which scenario we're hitting above the `if` and `else` conditions and the values we expect for the gzip_* variables.

tests/container/BUILD
```diff
@@ -238,13 +238,13 @@ container_push(
 # BEGIN_DO_NOT_IMPORT
 file_test(
     name = "test_digest_output1",
-    content = "sha256:18e27a73e3faabc819f903fc8e603de2bfd49813e231c2e8ce98b533577450e2",
+    content = "sha256:4ec78686f1edd4f30cca6222f17ace5a2b9f8db9708727c3a140c954af0a534d",
```
smukherj1, line 241: Hmm curious why the digest changed given we're still using gzip in the Go tool. Are we specifying a different compression level than before? In any case, I expect you'll need to update a bunch of digests here (https://github.com/bazelbuild/rules_docker/blob/master/container/image_test.py) as well in case we can't get the digests to match.

container/go/cmd/zipper/zipper.go
```diff
@@ -0,0 +1,81 @@
...
+
+	if *dst == "" {
+		log.Fatalln("Required option -dst was not specified.")
+	}
+
+	dat, err := ioutil.ReadFile(*src)
+	if err != nil {
+		log.Fatalf("Unable to read input file: %v", err)
+	}
+	var buf bytes.Buffer
```
smukherj1, line 50: Suggest using io.Copy (https://pkg.go.dev/io?tab=doc#Copy) to directly copy from a buffer pointing to the source file to a buffer pointing to the destination file. This avoids loading the entire src into memory.

When decompressing, the src would be a gzip.NewReader(dat), dst would be os.Create (https://pkg.go.dev/os?tab=doc#Create). When compressing, src would be dat and dst would be gzip.NewWriterLevel wrapping a os.Create.

Finally use io.Copy (https://pkg.go.dev/io?tab=doc#Copy) to copy from the source to destination buffer. If you declare io.Reader (https://pkg.go.dev/io?tab=doc#Reader) and io.Writer interface variables outside the if conditions, you can have one io.Copy at the end that handles both cases.

tests/container/BUILD
faube: I tried digging into this a little. None of the compression levels in compress/gzip produce a digest that matches the gzip tool. There may be subtle differences in how gzip and compress/gzip package the header or something?

//container/image_test.py passes; I suspect it's because we never verify the digest of gzipped files, only manifests.

container/go/cmd/zipper/BUILD
```diff
@@ -0,0 +1,34 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+# ###
+# Build file for zipper binary.
+
+load("@bazel_gazelle//:def.bzl", "gazelle")
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library")
+
+# gazelle:prefix github.com/bazelbuild/rules_docker
+gazelle(name = "gazelle")
```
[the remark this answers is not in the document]
faube: Done.

container/go/cmd/zipper/zipper.go
faube: Done.

skylib/zip.bzl
faube: Done. Note that the gzip_path == "" was done _after_ resolving gzip_path from the gzip_path/gzip_target attributes.
I replaced it with `if not gzip_path` to avoid the None issue.
Also added some comments.

skylib/zip.bzl
```diff
@@ -25,27 +85,7 @@ def gzip(ctx, artifact, options = None):
...
-    ctx.actions.run_shell(
-        command = "%s -n %s < %s > %s" % (gzip_path, opt_str, artifact.path, out.path),
-        input_manifests = input_manifests,
-        inputs = [artifact],
-        outputs = [out],
-        use_default_shell_env = True,
-        mnemonic = "GZIP",
-        tools = tools,
-    )
+    _gzip(ctx, artifact, out, False, options, "GZIP")
```
smukherj1, line 88: nit: here and below use named parameters to improve readability. Example:

```
_gzip(ctx=ctx,
         artifact=artifact,
         ...)
```

tests/container/BUILD
smukherj1: > I tried digging into this a little. None of the compression levels in compress/gzip produce a digest that matches the gzip tool. There may be subtle differences in how gzip and compress/gzip package the header or something?

SGTM. Thanks for looking into it.

> //container/image_test.py passes; I suspect it's because we never verify the digest of gzipped files, only manifests.

This is weird. I believe manifest transitively includes the digests the compressed layers. I don't remember if the manifest contents directly list the layer digests or the manifest which includes the digest of the image config which in turn includes the digests of the compressed layers. Note that some images only include uncompressed layers. However, I think most of our images include compressed layers. I'm a little worried that the current implementation isn't actually calling the Go tool to compress the layers for the images used by this test which may be undesirable for your use case :)

smukherj1 (commented):
One nit and one suggested exploration to ensure your change does what you wanted it to do. LGTM otherwise. Thanks for working on this!

skylib/zip.bzl
faube: Done.

faube:
[1613.zip](https://github.com/bazelbuild/rules_docker/files/5171194/1613.zip)
I looked at test_files_base and I am attaching the resulting files_base.tar
None of the layers in the file are .gz
In my experience, the intermediate layers that appear on disk (e.g. files_base-layer.tar.gz) are gz but the final image file that gets built in a :imagename.tar rule (as inspected in image_test) has no compression, so I'm under the impression that this is all WAI.

faube:
Similarly, if I 'docker save' a Docker image to a tarball, none of the internal layers are gz.

tests/container/BUILD
faube: Please see comment below: https://github.com/bazelbuild/rules_docker/pull/1613#issuecomment-686773106

smukherj1:
> [1613.zip](https://github.com/bazelbuild/rules_docker/files/5171194/1613.zip)
> I looked at test_files_base and I am attaching the resulting files_base.tar
> None of the layers in the file are .gz
> In my experience, the intermediate layers that appear on disk (e.g. files_base-layer.tar.gz) are gz but the final image file that gets built in a :imagename.tar rule (as inspected in image_test) has no compression, so I'm under the impression that this is all WAI.

TL; DR I'll approve this to make progress but I suggest keeping this point in mind if you notice the Go compressor isn't being called.

The tarball writer is special (See [here](https://github.com/google/go-containerregistry/blob/v0.1.2/pkg/legacy/tarball/write.go#L268)). The tarball *always* writes out uncompressed layers even if the [layer media type](https://github.com/google/go-containerregistry/blob/v0.1.2/pkg/v1/manifest.go#L29) indicated the layer is compressed. Thus, given that you noticed the compression by the Go tool results in a different digest, I would expect the digest of compressed layers to have changed as well resulting in image manifests being updated. You would need to see the image manifest (not the one in the tarball, the tarball [manifest](https://github.com/google/go-containerregistry/blob/master/pkg/v1/tarball/image.go#L116) has a different format.

smukherj1 (approved):

smukherj1:
@faube there's a buildifier formatting issue that's making buildkite fail:
```
buildifier: found 1 lint issue in your WORKSPACE, BUILD and *.bzl files
skylib/zip.bzl:31: function-docstring: Argument "mnemonic" is not documented.
```

smukherj1 (approved):

PR #1618, opened 2020-09-07
Add repository_file attribute to the container_pull rule

nlewo:
This attribute allows to set the repository value from a file as it is
already possible for the tag thanks to the tag_file attribute.

### DIFF
--- a/container/push.bzl
+++ b/container/push.bzl
@@ -45,6 +45,12 @@ def _impl(ctx):
     # Parse and get destination registry to be pushed to
     registry = ctx.expand_make_variables("registry", ctx.attr.registry, {})
     repository = ctx.expand_make_variables("repository", ctx.attr.repository, {})
+
+    # If a repository file is provided, override <repository> with tag value
+    if ctx.file.repository_file:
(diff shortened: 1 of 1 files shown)


nlewo:
Note we always need to provide the `repository` attribute because it is mandatory. Since for most use cases the `repository` attribute should be sufficient, i think it would be better to preserve this behavior.
WDYT?

smukherj1:
Seems reasonable to have `repository_file` given `tag_file` already exists. Feel free to update the [pusher](https://github.com/bazelbuild/rules_docker/blob/master/container/go/cmd/pusher/pusher.go) binary to implement this.

nlewo:
@smukherj1 What exactly should i implement in the pushed binary? 
The `repository_file` (and `tag_file`) value is substituted by bash (in the generated bash script): the pusher takes as argument a string such as `--dst=repository:$(cat ${RUNFILES}/tag)`, where `$(cat ${RUNFILES}/tag)` is evaluated by bash.

smukherj1:
Apologies, I missed the `cat`.

smukherj1 (approved):

nlewo:
Thanks ;)

PR #1619, opened 2020-09-11
Support airgapped environment

smartkiwi:
`go_repository` git `commit` doesn't support `http_load` caching mechanism thus prevents `rules_docker` to be used in air gapped environment.

Replacing it with alternative `urls` address makes go_repository to use `http_load` and utilize [caching mechanism](https://docs.bazel.build/versions/master/guide.html#running-bazel-in-an-airgapped-environment) if needed.
An additional piece of information will be needed in future for updating dependencies' versions - sha256.
The way to get sha256 on mac - to download the file from `https://api.github.com/repos/<org>/<repo>/tarball/<commit>` and running `shasum -a 256 <file>`

### DIFF
--- a/repositories/go_repositories.bzl
+++ b/repositories/go_repositories.bzl
@@ -37,32 +37,46 @@ def go_deps():
     if "com_github_google_go_containerregistry" not in excludes:
         go_repository(
             name = "com_github_google_go_containerregistry",
-            commit = "e5f4efd48dbff3ab3165a944d6777f8db28f0ccb",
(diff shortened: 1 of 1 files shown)


smukherj1 (approved):

gravypod:
@smartkiwi, is this something that could be accomplished using a [Remote Asset API](https://github.com/buchgr/bazel-remote#experimental-remote-asset-api-support)? I haven't ever worked in an airgapped environment so I don't really know much about the difficulties encountered here. 

This change is probably ok to make anyway. Something else to look forward to for your use case is the [External Dep Overhaul](https://github.com/bazelbuild/proposals#under-review).

Issue #1562, opened 2020-07-16
Passing flag --network=host to docker_flags of docker_toolchain_configure


whs-dot-hk:
In jenkins of kubernetes in order to run rules like `install_pkgs` the flag `--network=host` is need to pass to `docker build` but there is no way to supply that flag

```groovy
podTemplate(yaml: """
apiVersion: v1
kind: Pod
spec:
  serviceAccountName: ecr
  containers:
  - name: bazel
    image: gcr.io/cloud-marketplace-containers/google/bazel@sha256:bea7cec14f05aea5a8a8ead0ddbf9a8ef66b4ad2a0f997c8d1d0217804044b8a
    command:
    - cat
    tty: true
    volumeMounts:
    - name: dockersock
      mountPath: /var/run/docker.sock
  volumes:
  - name: dockersock
    hostPath:
      path: /var/run/docker.sock
"""
) {
    node(POD_LABEL) {
      checkout scm

      container("bazel") {
        sh "bazel run //:go_image"
      }
    }
}
```
```starlark
download_pkgs(
    name = "cacerts_pkg",
    image_tar = "@debian_image//image",
    packages = [
        "ca-certificates",
    ],
)

install_pkgs(
    name = "cacerts_pkg_image",
    image_tar = "@debian_image//image",
    installables_tar = ":cacerts_pkg.tar",
    installation_cleanup_commands = "rm -rf /var/lib/apt/lists/*",
    output_image_name = "cacerts_pkg_image",
)

container_image(
    name = "cacerts_wrapper",
    base = ":cacerts_pkg_image.tar",
)

go_image(
    name = "go_image",
    base = ":cacerts_wrapper",
    embed = [":go_default_library"],
    visibility = ["//visibility:public"],
)
```

austince:
Also running into this issue when building in GitLab CI, where the docker daemon is running as a service. Are you seeing an error message similar to this?
```
+ /usr/local/bin/docker load -i bazel-out/k8-fastbuild/bin/external/node_lts_slim/image/image.tar
Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?
```

whs-dot-hk:
Hi, @austince I didn't came across the error because I supplied the docker socket through `hostPath`.

Can you config the docker though `docker_toolchain_configure`?

austince:
[...]

PR #1638, opened 2020-10-01
Add docker_run_flags to download_pkgs

whs-dot-hk:
This pr add `docker_run_flags` to `dowload_pkgs`. An extra run flags `--network=host` is need in some case, to download the packages.

#1562 is related.


carolguo-dd:
Hi, When can we merge this? I think I also need this feature. I ran into this error on gitlab ci 
```
docker: Error response from daemon: privileged mode is incompatible with user namespaces.  You must run the container in the host namespace when running privileged mode.
```
which seems to be caused by the `--privileged` flag https://github.com/austince/rules_docker/blob/master/docker/package_managers/run_download.sh.tpl#L31 and I might need to use `--userns=host` flag to fix it

pcj:
This appears to be a reasonable low-risk change to merge.  @smukherj1 Any thoughts?

smukherj1:
> This appears to be a reasonable low-risk change to merge. @smukherj1 Any thoughts?

Change itself lgtm. Might be worth thinking about a test to verify if the arguments were actually passed to the final generated docker command.

whs-dot-hk:
Hi, @smukherj1, the regex test checked that `--network=host` is pass to the final docker command?

lukokr-aarch64:
@smukherj1 any chance we could reopen this PR?

Issue #1623, opened 2020-09-15
container_run_and_commit does not support `repository`


fosrias:
Because this rule outputs an image, it should support setting the `repository` like other rules, should it not?

PR #1641, opened 2020-10-02
add repository attribute to container_run_and_commit rule

fosrias:
The PR addresses https://github.com/bazelbuild/rules_docker/issues/1623

I would like to add tests, but in reviewing other tests w/r to the `repository` property on `container_image`, it was not clear if/how it is tested. Please point me to a representative test and I will add one. Thx.


smukherj1 (commented):
I agree customizing the repository name in this rule makes sense. As for test, a possibility is to add an e2e test that builds an image with a custom repository name, load it directly into docker and validate the stdout from the docker load mentioned the custom repository name. To do you, you'll need to add an image test target and then build & load it probably [here](https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e/bazel_run_docker.sh) and then actually call the new function you add [here](https://github.com/bazelbuild/rules_docker/blob/master/testing/e2e/test_bazel_run_docker.yaml). The YAML file is the Google Cloud Build config.

davidxia:
I would love this functionality. Is there still appetite for this PR? Is anyone working on it?

davidxia:
How are the e2e tests run? I don't see where they run in the buildkite build here https://github.com/bazelbuild/rules_docker/pull/1889.

PR #1653, opened 2020-10-18
Fixed load() call for pip_deps in WORKSPACE described in README

MattTannahill:
After setting up my WORKSPACE as described in README.md, I received the error:
```
Error: file '@io_bazel_rules_docker//repositories:pip_repositories.bzl' does not contain symbol 'io_bazel_rules_docker_pip_deps'
```

This was fixed by changing the requested symbol to `pip_deps`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,9 +167,9 @@ load("@io_bazel_rules_docker//repositories:deps.bzl", container_deps = "deps")
 
 container_deps()
 
-load("@io_bazel_rules_docker//repositories:pip_repositories.bzl", "io_bazel_rules_docker_pip_deps")
+load("@io_bazel_rules_docker//repositories:pip_repositories.bzl", container_pip_deps = "pip_deps")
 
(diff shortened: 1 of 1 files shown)


MattTannahill:
@googlebot I signed it!

MattTannahill:
After looking into this a bit more closely, it looks like the name of the symbol was changed since release 0.14.4.  The correction is valid for release used in the docs, but will need to be updated again in the next release unless the symbol is changed back to `pip_deps`.

In version: 0.14.4:
https://github.com/bazelbuild/rules_docker/blob/faaa10a72fa9abde070e2a20d6046e9f9b849e9a/repositories/pip_repositories.bzl#L24

In master:
https://github.com/bazelbuild/rules_docker/blob/cc45596d140b3b8651eb7b51b561f1bf72d1eea9/repositories/pip_repositories.bzl#L24

Issue #525, opened 2018-09-21
Failed while running push-all


hanglethao:
Hi

I am using `push_all` to push a `container_bundle` which contains more than 50 images.
And I got the error below.

```
line 134: wait: pid 2438 is not a child of this shell
```

I think it relates to this `wait`.
https://github.com/bazelbuild/rules_docker/blob/master/contrib/push-all.sh.tpl#L37

smukherj1:
Hello hanglethao,

Thanks for reporting this issue. We need more information about the platform where you observed this issue. Specifically the following:-

1. Operating System and Version (eg. Ubuntu 16.04 64-bit or Debian 9 64 bit, etc)
2. Bazel version (eg. 0.15.1)
3. Version of your execution shell (eg. bash 4.3.12)

Please also provide a test case to reproduce this issue. You could add a test case under //rules_docker/tests in a forked repo or simply a tarball of an individual bazel project that shows the issue.

Also, if you have any ideas on why the issue might be happening, don't hesitate to submit a pull request with the proposed fix. The error above indicates that the process the script is being asked to wait for wasn't launched from the same shell.

hanglethao:
Hi @smukherj1 

> Operating System and Version

ubuntu:16.04

> Bazel version

0.16.1

> rules_docker version

0.5.1

> Version of your execution shell

version 4.3.48(1)-release (x86_64-pc-linux-gnu)

I just created 50 binaries and 50 go_images. Then I used the following `docker_push` to push all the images and got that error.


1. BUILD.bazel
``` 
load(":image.bzl", "all_images")
load("@io_bazel_rules_docker//container:container.bzl", "container_bundle")

container_bundle(
    name = "bundle_to_push",
    images = all_images(),
    stamp = True,
)

load("@io_bazel_rules_docker//contrib:push-all.bzl", "docker_push")

docker_push(
    name = "push_all_images",
    bundle = ":bundle_to_push",
)

```

2. image.bazel
```
def all_images():
    images = {}
    for i in range (1, 50):
        cmd = "example-%s" % i
        images["$(REGISTRY)/$(PROJECT_ID)/%s:{STABLE_GIT_COMMIT}" %cmd ] = "//cmd/%s:image" % cmd
    return images
```

nlopezgi:
Thanks for providing the repro details, unfortunately we do not have tests that use a large number of images at the moment, so this might be a regression or a long standing issue (or even a flawed design that prevents this from working with a large number of images). 

[...]

PR #1658, opened 2020-10-21
Add option to push images sequentially in push-all

ImJasonH:
There seems to be a bug with pushing many images (50+) related to the
async+wait behavior previously used, and this new option will give users
who have lots of images to push a safer option.

Fixes https://github.com/bazelbuild/rules_docker/issues/525

### DIFF
--- a/contrib/push-all.bzl
+++ b/contrib/push-all.bzl
@@ -68,10 +68,15 @@ def _impl(ctx):
     ctx.actions.expand_template(
         template = ctx.file._all_tpl,
         substitutions = {
-            "%{push_statements}": "\n".join([
+            "%{async_push_statements}": "\n".join([
                 "async \"%s\"" % _get_runfile_path(ctx, command)
                 for command in scripts
             ]),
+            "%{push_statements}": "\n".join([
+                "\"%s\"" % _get_runfile_path(ctx, command)
+                for command in scripts
+            ]),
+            "%{sequential}": "true" if ctx.attr.sequential else "",
         },
         output = ctx.outputs.executable,
         is_executable = True,
@@ -99,6 +104,10 @@ container_push = rule(
                 "Docker",
             ],
             doc = "The form to push: Docker or OCI.",
+        ),
+        "sequential": attr.bool(
+            default = False,
+            doc = "If true, push images sequentially.",
         ),
         "_all_tpl": attr.label(
             default = Label("//contrib:push-all.sh.tpl"),
--- a/contrib/push-all.sh.tpl
+++ b/contrib/push-all.sh.tpl
@@ -34,11 +34,17 @@ function async() {
     PIDS+=($!)
 }
 
-%{push_statements}
-
-# Wait for all of the subprocesses, failing the script if any of them failed.
-if [ "${#PIDS[@]}" != 0 ]; then
+SEQUENTIAL="%{sequential}"
+if [ -z "${SEQUENTIAL}" ]; then
+  %{async_push_statements}
+  # Wait for all of the subprocesses, failing the script if any of them failed.
+  if [ "${#PIDS[@]}" != 0 ]; then
     for pid in ${PIDS[@]}; do
-        wait ${pid}
+      wait ${pid}
     done
+  fi
+else
+  %{push_statements}
 fi
+
+


ImJasonH:
cc @nlopezgi @smukherj1

eytankidron:
I am worried that this change might negatively affect existing use cases (i.e. make them slower).
I'd like to suggest one of two alternate approaches:
(1) A combination of sequential and parallel: Group the images into groups of, say, 50, and push the images within a group in parallel. The groups will be pushed sequentially.
(2) Add an optional *sequential* boolean flag to container_push and have it default to false.

ImJasonH:
> I am worried that this change might negatively affect existing use cases (i.e. make them slower).
> I'd like to suggest one of two alternate approaches:
> (1) A combination of sequential and parallel: Group the images into groups of, say, 50, and push the images within a group in parallel. The groups will be pushed sequentially.
> (2) Add an optional _sequential_ boolean flag to container_push and have it default to false.

That's a reasonable concern. In my experience, pushing 178 relatively small images sequentially, from source uploaded to and built in GCB to GCR, took only a minute or two longer than pushing in parallel.

I talked with @jonjohnsonjr and he mentioned it could actually be somewhat _faster_ to push sequentially, since blob existence checks could get more hits. This is especially true if the images tend to share layers, which in my case they do, and even moreso if the layers are large (mine generally aren't).

Weigh this against the not-entirely-well-known issue that manifests when "too many" (How many? Not empirically known) images are pushed, caused by the `async` function and `wait $PID`. Ideally we'd not rely on bash for async fan-out/fan-in behavior, since it's not well suited for that. Rewriting parallel pushing efficiently in Go seems useful.

eytankidron:
You might be right about sequential pushes being faster in some cases. That sounds reasonable, but honestly, I don't know, and I'd rather not make this change which might affect existing users in unforeseeable ways. 
Just to be on the safe side, we can go with the sequential (or parallel) flag for now.
I see that you started adding something like that in https://github.com/bazelbuild/rules_docker/commit/5e7547efeaa37019f72498ad569c8165c8eec3db but I don't see it in this PR. Did you revert that commit?

ImJasonH:
> You might be right about sequential pushes being faster in some cases. That sounds reasonable, but honestly, I don't know, and I'd rather not make this change which might affect existing users in unforeseeable ways.
> Just to be on the safe side, we can go with the sequential (or parallel) flag for now.
> I see that you started adding something like that in [5e7547e](https://github.com/bazelbuild/rules_docker/commit/5e7547efeaa37019f72498ad569c8165c8eec3db) but I don't see it in this PR. Did you revert that commit?

I did revert the commit, since I wanted to plead my case for sequential-by-default first.

Existing users are _already_ affected in unforeseeable ways -- at some point they can add too many images, and push-all will fail in mysterious and hard to debug ways. This change removes that unspecified hard-to-debug behavior at the cost of some latency.

Some users might be negatively affected by that increase in latency, but push-all never specified parallel pushes to users, and since it's turning out to be unsafe, I think it's safe to do from a user-surprise perspective.

I'm separately working on some changes in go-containerregistry to efficiently (and safely!) push multiple images in parallel, and if that lands I expect it could be very useful in this scenario.

eytankidron:
I'm honestly not sure how most users use push-all. Do they mostly use it to push <50 images and may be negatively affect by slowness? Maybe, maybe not.
Having said that, you do have a point that being broken in a way that's hard to debug is a scenario we want to avoid.

So how about adding a sequential/parallel flag and have it default to sequential. Then if anyone notices a hit in performance, they can always switch back to parallel.

Additionally, once your changes to container registry land, we can switch to using that.

ImJasonH:
The main reason I don't want to add a flag, whatever its default, is that we fully expect to remove it in the future, and that seems like a hairy breaking change to manage. Meanwhile I just want to successfully push my 100+ images using Bazel 😄 

If the behavior of push-all remains "unspecified", we have flexibility to say that, sure, it _was_ in parallel, but now it's sequential for safety. If a user is unsatsified with the additional latency after this change (which we're not even sure they'll notice), they can use an older version of rules_docker until some future change supports safe parallel pushing without unspecified random failures.

eytankidron:
I'm okay with paying the price of the eventual deprecation of the sequential/parallel flag. I don't think it needs to be a breaking change. The way I see it, once the efficient and correct parallel push is rolled out, we just declare this flag as deprecated and allow users to set whatever value for this flag, while documenting that we are now ignoring it.

I think this is a reasonable price to pay for the ability to have users be able to choose the best solution that works for them until this better push lands. I am saying this with the full understanding that we have no way of knowing (nor do we plan to try and figure out) what the best solution is for users or what the users think the best solution for them is. I just don't want to be in a situation where a customer wants to go back to the the old way of pushing but for some other reason can't use the older version.

ImJasonH:
> I'm okay with paying the price of the eventual deprecation of the sequential/parallel flag. I don't think it needs to be a breaking change. The way I see it, once the efficient and correct parallel push is rolled out, we just declare this flag as deprecated and allow users to set whatever value for this flag, while documenting that we are now ignoring it.
> 
> I think this is a reasonable price to pay for the ability to have users be able to choose the best solution that works for them until this better push lands. I am saying this with the full understanding that we have no way of knowing (nor do we plan to try and figure out) what the best solution is for users or what the users think the best solution for them is. I just don't want to be in a situation where a customer wants to go back to the the old way of pushing but for some other reason can't use the older version.

If you're comfortable managing the removal of the flag that's fine with me. I'll ping this PR when I have the `sequential=True` flag landed (the default will remain `False` to maintain current unsafe parallel push behavior)

FYI I just landed [`remote.MultiWrite` in go-containerregistry](https://godoc.org/github.com/google/go-containerregistry/pkg/v1/remote#MultiWrite) which a tool in this repo could leverage to make fast and safe parallel uploads a reality.

ImJasonH:
Confirmed that this PR works to push distroless images sequentially, and that removing `sequential = False` successfully pushed images, then resulted in the error we saw before.

https://github.com/GoogleContainerTools/distroless/pull/624

eytankidron (approved):
Thanks for doing this Jason.

eytankidron:
Looks like buidifier is failing because of lexicographical order.
> contrib/push-all.bzl:72: unsorted-dict-items: Dictionary items are out of their lexicographical order.
> contrib/push-all.bzl:74: unsorted-dict-items: Dictionary items are out of their lexicographical order.

eytankidron (approved):

wyattanderson:
> FYI I just landed [`remote.MultiWrite` in go-containerregistry](https://godoc.org/github.com/google/go-containerregistry/pkg/v1/remote#MultiWrite) which a tool in this repo could leverage to make fast and safe parallel uploads a reality.

Is there interest in migrating push behavior into Go code from Bash? We're hitting some rate limits with AWS when pushing hundreds of images in parallel, and we may be able to remedy that by pushing sequentially, but if there's interest/alignment in moving the parallelism into a Go binary, I could explore the concept in a pull request.

PR #1660, opened 2020-10-22
Use bash from the environment instead of absolute path

kalbasit:
Not all operating systems ship with bash located at `/bin/bash`. Change the interpreter to `/usr/bin/env bash` to request bash from the environment.

For instance, I use [NixOS](https://nixos.org) as my daily driver and this operating system does not follow [FHS](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard); In fact it does not have a `/bin` and the only thing in `/usr/bin` is `env`.

Related to https://github.com/bazelbuild/rules_go/pull/2295

### DIFF
--- a/contrib/cmp_images.sh.tpl
+++ b/contrib/cmp_images.sh.tpl
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # Copyright 2017 The Bazel Authors. All rights reserved.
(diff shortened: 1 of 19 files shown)


smukherj1 (approved):

smukherj1:
CI is failing due to a known infra issue. Will need to retry tomorrow

PR #1674, opened 2020-11-09
Fix select with condition on darwin.

comius:
Bazel is changing how @bazel_tools/conditions:darwin is implemented.
With old implementation based on flags, it was ok to have darwin and
darwin_x86_64 in a select. This was based on flag --cpu=darwin and
--cpu=dawin_x86_64.

New implementation is based on constraints, where "darwin" means OS (and
any CPU), while "darwin_x86_64" mean only specific CPU. As such they
cannot be used in the same select, because the selection would be
ambiguous.

### DIFF
--- a/contrib/BUILD
+++ b/contrib/BUILD
@@ -63,7 +63,6 @@ alias(
     name = "structure_test_executable",
     actual = select({
         "@bazel_tools//src/conditions:darwin": "@structure_test_darwin//file",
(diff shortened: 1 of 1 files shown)


DaveGay (approved):

comius:
Is there anyway to push this merge forward? It seems there is a problem with travis CI.

PR #1679, opened 2020-11-19
container_test pass tags to container_bundle

kalbasit:
Calling `container_test` with tags is creating a `container_bundle` that is not tagged. This change passes the tags to the container_bundle. The change is inspired by the passing of `restricted_to` and `compatible_with`.

Question: Why aren't we passing `**kwargs` instead of calling out specific keys?


kalbasit:
fixed by #1730.

PR #1685, opened 2020-11-23
Introduce output_group for pusher executable

### LINKED ISSUE #1683: container_push doesn't allow kwargs for all attrs
Three undocumented attributes on container_push
https://github.com/bazelbuild/rules_docker/blob/c85e7b6cd5f812081bbfe44fb657ce89ffebfcc3/container/push.bzl#L227-L238
in particular `extension` can be passed by users, but result in an error after macro expansion
`Error in rule: rule(...) got multiple values for parameter 'extension'`
because they should be pop'ed from the `**kwargs` if present (even better, use these selects as the default if the attribute isn't provided)

I'll send a PR for this.

### LINKED ISSUE #1684: container_push: impossible to reference the pusher executable
I want to pass the pusher program produced as the executable output of a `container_push` rule as the value of a `allow_single_file` attribute in my custom release rule.

Currently the default (non-windows) executable path is the same as the target:
`//my:cont_push` is both the target name and the executable name. However the default outputs also include the `cont_push.digest` file so the only valid label resolves to two files.

Proposal: add an `OutputGroupInfo` to the container_push like `exe = ["exe"]` so I can `filegroup(output_group="exe")` to label just the pusher executable.

Will send a PR for this.

alexeagle:
This allows the pusher to be referenced as a label without including all the default outputs including the .digest file

Fixes #1684

Also use the user-provided value for the (undocumented) "extension" parameter if one is set.

Fixes #1683

### DIFF
--- a/container/push.bzl
+++ b/container/push.bzl
@@ -132,6 +132,9 @@ def _impl(ctx):
             executable = exe,
             runfiles = runfiles,
         ),
+        OutputGroupInfo(
+            exe = [exe],
+        ),
         PushInfo(
             registry = registry,
             repository = repository,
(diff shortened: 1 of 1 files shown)


alexeagle:
ping @smukherj1 :)

smukherj1:
> ping @smukherj1 :)

Apologies for the massive delay. We'll get to this within a week

eytankidron:
Dumb question: when you call
`extension = kwargs.pop("extension", ...)`
that pops "extension" out of kwargs. Is it what you want to do given that you then pass **kwargs as the last arg to _container_push?
Would `kwargs.get` be better?

alexeagle:
@eytankidron yes it needs `pop` so that we don't pass the `extension` attribute twice.

ping again on getting this in?

eytankidron (approved):

PR #1686, opened 2020-11-25
Use stamping in container image env parameter values

escodebar:
...since the documentation says it works!


escodebar:
I committed with my work email. Please let me know if you prefer me to use the one I use for GitHub.

escodebar:
Oh... I am pretty sure, this PR is not necessary. When I saw that the parameter `stamp` is deprecated, I didn't notice that it concerns the `container_push` rule and `stamp` is still a valid parameter for `container_image`.

So, this PR can be closed and deleted.

PR #1700, opened 2020-12-17
go_register_toolchains: set default version

rgs1:
The support for a default version in go_register_toolchains is gone, so
we need to pass an explicit version.

Signed-off-by: Raul Gutierrez Segales <rgs@pinterest.com>


rgs1:
@googlebot I signed it!

rgs1:
Not really needed, see workaround here:

https://github.com/bazelbuild/rules_go/issues/2761

PR #1713, opened 2021-01-15
python3: Respect tags field

### LINKED ISSUE #523: Common attribute: tags=["manual"] doesn't work. Rule is always ran with //...
I would like to run container_image and container_push as manual rules and not part of the //... build. In the common attributes page of bazel: https://docs.bazel.build/versions/master/be/common-definitions.html#common-attributes. It states that tags are a common attribute to control the build. 

Would it be possible to add this to the container_* rules? Or is there a work around to have container_* rules not run with the generic //... build?

kibumh:
'tags' fields are not applied to layer rules.
Fixes #523 for python3.

### DIFF
--- a/python3/image.bzl
+++ b/python3/image.bzl
@@ -103,12 +103,12 @@ def py3_image(name, base = None, deps = [], layers = [], **kwargs):
     # TODO(mattmoor): Consider making the directory into which the app
     # is placed configurable.
     base = base or DEFAULT_BASE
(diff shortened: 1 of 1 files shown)


combacsa:
Seems like #1836 is already merged which addressed same issue?

kibumh:
@combacsa, thanks for letting me know.
Happy to close this PR.

PR #1730, opened 2021-02-08
Propagate tags to the container image test bundle

### LINKED ISSUE #1729: container_test synthetic image does not inherit tags from the macro
A container_test rule "foo_test" will generate another target "foo_test.image" but currently tags like "tags = ['requires-cgo'] are not propagated to that image. 

Screen Shot 2021-02-08 at 12 31 58 PM

pcj:
Fixes #1729

Screen Shot 2021-02-08 at 12 36 08 PM


pcj:
Gentle Ping

alexeagle (approved):

PR #1732, opened 2021-02-11
Switch to non-api fetch URLs

illicitonion:
This is a no-op change.

These URLs have a few advantages:
* GitHub is less aggressive at rate-limiting these URLs, and these URLs
  are easier to set up pass-through proxies for in an enterprise
  setting.
* The URLs self-document which tag is being fetched, rather than
  requiring separate comments to track tags (but the sha256 ensures
  re-pointed tags will still halt builds). That said, there are
  sha1-equivalent URLs if we'd prefer to keep them.
* The URLs have file extensions, so don't need a separate `type`
  attribute


alexeagle (approved):
Thanks! Avoiding rate-limiting sounds great

PR #1739, opened 2021-02-23
Add Domino Data Lab to list of adopters

urianchang:
At [Domino Data Lab](https://www.dominodatalab.com/), we're using `rules_docker` to build Docker images in our monorepo. Hopefully our usage encourages others to adopt `rules_docker` as well.


zoidbergwill:
This seems like a pretty safe change to approve.

alexeagle (approved):
Thanks!

Issue #770, opened 2019-03-29
Ability to add healthcheck


dgarzon:
Hi, is there a way to add a healthcheck to a `container_image` or any of the language rules?

xingao267:
HI @dgarzon what kind of health check would you like?

dgarzon:
Hi @xingao267, we are running a web-server on a `scala_image` and we would like to check that hitting an endpoint yields `20X` when `curl`'d. Any ideas? Thanks in advanced!

xingao267:
Sorry, I'm not sure I understand it. Could you expand a little more?

dgarzon:
I would like to do this: https://docs.docker.com/engine/reference/builder/#healthcheck, it is a deterministic operation, so it should be totally compatible with Bazel.

alex1545:
Hi @dgarzon, in order for us to determine whether or not we will be able to add a healthcheck to any of the image rules, we need to know how the healthcheck is being set once the image is built from a Dockerfile with a `HEALTHCHECK` instruction. I.e., does it add/modify any files in any of the image layres? Or maybe it sets any special fields in the manifest.json file?
Are you able to help answer this question or maybe explore this if needed?

dgarzon:
Hi @alex1545 , after looking at https://github.com/docker/docker-ce/blob/8a582e89b407639018e067123eb17f61dbfd10ee/components/engine/daemon/health_test.go, I believe `HEALTHCHECK` only adds an entry to the `manifest.json` file with certain fields. Does this help?

alex1545:
Hi @dgarzon, this is definitely useful information, but certainly not enough. Can you investigate further with more details?
If `HEALTHCHECK` only adds an entry to the `manifest.json` file, then can you find out the exact entry being added and its format? Can you verify that modifying the `manifest.json` is all that's needed and possibly share an example of a `manifest.json` file with and without the healthcheck entry that can be used with the same image?

dgarzon:
@alex1545 I did some more research. Adding a `HEALTHCHECK` to a `Dockerfile`, does not add any entry to the `manifest.json`, and rather it adds to `config.json`. Here are some examples:

1. User defines `HEALTHCHECK NONE`:
```
{
    ...
    "config": {
        ...
        "Healthcheck": {
            "Test": [
                "NONE"
            ]
        }
    }
}
```

2. User defines `HEALTHCHECK CMD curl --fail http://localhost:5000 || exit 1`:
```
{
    ...
[...]

PR #1742, opened 2021-03-01
Add the ability to add healthcheck definition for containers

svagner:
Original issue: https://github.com/bazelbuild/rules_docker/issues/770

Adds follow options for container_image:
- healthcheck_test
- healthcheck_interval
- healthcheck_timeout
- healthcheck_start_period
- healthcheck_retries

### DIFF
--- a/README.md
+++ b/README.md
@@ -2201,9 +2201,12 @@ container_image(name, base, data_path, directory, files, legacy_repository_namin
         <code>List of strings, optional</code>
         <p><a href="https://docs.docker.com/engine/reference/builder/#healthcheck">
         The test to perform to check that the container is healthy.</a></p>
-        <p>An empty list means to inherit the default.</p>
+        <p>
+        An empty list means to inherit the default. If <code>healthcheck_test</code> wasn't defined
+        <code>HEALTHCHECK</code> will be inherited from an upper layers.
+        </p>
         <p>Possible definition options are:</p>
-        <p><code>[]</code> : inherit healthcheck</p>
+        <p><code>None</code>/<code>[]</code> : inherit healthcheck</p>
         <p><code>["NONE"]</code> : disable healthcheck</p>
         <p><code>["CMD", "test", "-f", "file"]</code> : exec arguments directly</p>
         <p><code>["CMD-SHELL", "curl -q localhost:8080"]</code> : run command with system's default shell</p>
--- a/README.md
+++ b/README.md
@@ -2195,6 +2195,82 @@ container_image(name, base, data_path, directory, files, legacy_repository_namin
         </p>
       </td>
     </tr>
+    <tr>
+      <td><code>healthcheck_test</code></td>
+      <td>
+        <code>List of strings, optional</code>
+        <p><a href="https://docs.docker.com/engine/reference/builder/#healthcheck">
+        The test to perform to check that the container is healthy.</a></p>
+        <p>An empty list means to inherit the default.</p>
+        <p>Possible definition options are:</p>
+        <p><code>[]</code> : inherit healthcheck</p>
+        <p><code>["NONE"]</code> : disable healthcheck</p>
+        <p><code>["CMD", "test", "-f", "file"]</code> : exec arguments directly</p>
+        <p><code>["CMD-SHELL", "curl -q localhost:8080"]</code> : run command with system's default shell</p>
+        <p>
+          <code>
+          healthcheck_test = ["CMD", "test", "-f", "file"],
+          </code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>healthcheck_interval</code></td>
+      <td>
+        <code>strings, optional</code>
+        <p>Interval is the time to wait between health checks.</p>
+        <p>
+          Should be in the duration format. Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h"
+        </p>
+        <p>
+          <code>
+          healthcheck_interval = "30s",
+          </code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>healthcheck_timeout</code></td>
+      <td>
+        <code>strings, optional</code>
+        <p>Health check timeout is the time to wait before considering the check to have hung.</p>
+        <p>
+          Should be in the duration format. Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h"
+        </p>
+        <p>
+          <code>
+          healthcheck_timeout = "30s",
+          </code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>healthcheck_start_period</code></td>
+      <td>
+        <code>strings, optional</code>
+        <p>The health check start period for the container to initialize before the retries starts to count down.</p>
+        <p>
+          Should be in the duration format. Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h"
+        </p>
+        <p>
+          <code>
+          healthcheck_start_period = "10s",
+          </code>
+        </p>
+      </td>
+    </tr>
+    <tr>
+      <td><code>healthcheck_retries</code></td>
+      <td>
+        <code>integer, optional</code>
+        <p>Health check retries is the number of consecutive failures needed to consider a container as unhealthy.</p>
+        <p>
+          <code>
+          healthcheck_retries = 2,
+          </code>
+        </p>
+      </td>
+    </tr>
   </tbody>
 </table>
 
--- a/container/go/pkg/compat/config.go
+++ b/container/go/pkg/compat/config.go
@@ -42,6 +42,29 @@ const (
 	defaultTimestamp = "1970-01-01T00:00:00Z"
 )
 
+var (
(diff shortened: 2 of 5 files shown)


svagner:
@googlebot I signed it!

svagner:
cc @xingao267 , @alex1545 as you were in the discussion in the original issue

container/go/pkg/compat/config.go
```diff
@@ -42,6 +42,29 @@ const (
 	defaultTimestamp = "1970-01-01T00:00:00Z"
 )
 
+var (
+	validHealcheckTypes = map[string]func([]string) error{
```
zoidbergwill, line 46: nit

```suggestion
	validHealthcheckTypes = map[string]func([]string) error{
```

container/go/pkg/compat/config.go
```diff
@@ -42,6 +42,29 @@ const (
...
+		"NONE": func(args []string) error {
+			if len(args) == 1 {
+				return nil
+			}
+			return errors.New("NONE doesn't accept any extra args")
```
zoidbergwill, line 51: I'm not super familiar with Go, but I think it's common practice to put the error case in the indented if statement and have the default be the positive best case path.

zoidbergwill:
Hiding whitespace changes definitely makes this diff easier to review: https://github.com/bazelbuild/rules_docker/pull/1742/files?w=1

container/go/pkg/compat/config.go
svagner: Fixed. Thanks!

container/go/pkg/compat/config.go
svagner: Yeah. I don't have a strong opinion on that so I've moved the error case under a conditional branch as you offered

README.md
```diff
@@ -2195,6 +2195,82 @@ container_image(name, base, data_path, directory, files, legacy_repository_namin
...
     </tr>
+    <tr>
+      <td><code>healthcheck_test</code></td>
+      <td>
+        <code>List of strings, optional</code>
+        <p><a href="https://docs.docker.com/engine/reference/builder/#healthcheck">
+        The test to perform to check that the container is healthy.</a></p>
+        <p>An empty list means to inherit the default.</p>
+        <p>Possible definition options are:</p>
+        <p><code>[]</code> : inherit healthcheck</p>
```
gravypod, line 2206: `None` also inherits the healthcheck. It may be nice to specify this for callers who write their own macros wrapping this.

gravypod:
I don't see a reason why this couldn't be merged. The only concern I have is that we're overloading `healthcheck_test` to do multiple types of tests (`CMD`, `CMD-SHELL`, `NONE`, etc). Ideally I'd gravitate towards separating these out but I'm not sure that would make sense in this case due to this being a docker-specific rule. 

I'm leaning towards thinking this is fine to include but I'd like the opinion of another maintainer just as a sanity check.

pcj:
I think it's cool that you put this together, but I'm a little wary of this as-is for a couple of reasons:

1. It's a bit telling that I've never heard of this feature before today.  As far as I can tell, it's not supported in kubernetes but was a feature added on by Docker Inc for swarm specifically.  I don't think this feature is going to get much mileage, and it comes at a time when we're trying to simplify this repo rather than add complexity.
2. Given a `Dockerfile` only has the `HEALTHCHECK` command, why does `rules_docker` now need 5 new options to say the same thing?  If there is no parsing library in `google/go-containerregistry` (upon which this repo depends) to parse that into the necessary config, then that library/function should probably there upstream before merging this.  I don't think we should need more than a single new `string` or `string_list` attribute `healthcheck` for this.

README.md
svagner: Added. Thanks!

svagner:
> I don't see a reason why this couldn't be merged. The only concern I have is that we're overloading `healthcheck_test` to do multiple types of tests (`CMD`, `CMD-SHELL`, `NONE`, etc). Ideally I'd gravitate towards separating these out but I'm not sure that would make sense in this case due to this being a docker-specific rule.
> 
> I'm leaning towards thinking this is fine to include but I'd like the opinion of another maintainer just as a sanity check.

Yeah. I also thought that probably it's better to have `healthcheck_cmd` and `healthcheck_cmd_shell` options. But it will lead to more complications:(

svagner:
> I think it's cool that you put this together, but I'm a little wary of this as-is for a couple of reasons:
> 
> 1. It's a bit telling that I've never heard of this feature before today.  As far as I can tell, it's not supported in kubernetes but was a feature added on by Docker Inc for swarm specifically.  I don't think this feature is going to get much mileage, and it comes at a time when we're trying to simplify this repo rather than add complexity.
> 2. Given a `Dockerfile` only has the `HEALTHCHECK` command, why does `rules_docker` now need 5 new options to say the same thing?  If there is no parsing library in `google/go-containerregistry` (upon which this repo depends) to parse that into the necessary config, then that library/function should probably there upstream before merging this.  I don't think we should need more than a single new `string` or `string_list` attribute `healthcheck` for this.

Thanks for the feedback. Let me explain my thoughts about it.

1. These rules are called `rules_docker` so I believe it means that we should support at least all the main features of docker. I agree that kube doesn't use docker's `HEALTHCHECK` since it has it's own` livenessProbe` but this plugin isn't dedicated to kube only is it?
2. I think having `healthcheck_test =" --interval = 5s --timeout = 10s --retries = 3 CMD curl -sS 127.0.0.1 "" will be a little bit confusing for users of these rules since interval / to / retries are optional parameters for health checking right? And unfortunately I don't follow your point about the parsing of this parameter in the `google / go-containerregistry`. Since it is a single directive there is not necessary to parse it. But I'm again worrying about moving complications of the definition for health checks from the rules code to users.

Let me know if my arguments make sense to you.

pcj:
I don't see how `--interval = 5s --timeout = 10s --retries = 3 CMD curl -sS 127.0.0.1` could possibly be confusing to any user wanting to use this, because that's the syntax Docker came up with and has clearly documented.

And it is definitely necessary to parse it, because the configuration for it is not a simple string.  It is https://github.com/moby/moby/blob/46cdcd206c56172b95ba5c77b827a722dab426c5/api/types/container/config.go#L16-L35:

```go
// HealthConfig holds configuration settings for the HEALTHCHECK feature.
type HealthConfig struct {
	// Test is the test to perform to check that the container is healthy.
	// An empty slice means to inherit the default.
	// The options are:
	// {} : inherit healthcheck
	// {"NONE"} : disable healthcheck
	// {"CMD", args...} : exec arguments directly
	// {"CMD-SHELL", command} : run command with system's default shell
	Test []string `json:",omitempty"`

	// Zero means to inherit. Durations are expressed as integer nanoseconds.
	Interval    time.Duration `json:",omitempty"` // Interval is the time to wait between checks.
	Timeout     time.Duration `json:",omitempty"` // Timeout is the time to wait before considering the check to have hung.
	StartPeriod time.Duration `json:",omitempty"` // The start period for the container to initialize before the retries starts to count down.

	// Retries is the number of consecutive failures needed to consider a container as unhealthy.
	// Zero means inherit.
	Retries int `json:",omitempty"`
}
```

Part of the parser is here: https://github.com/moby/buildkit/blob/master/frontend/dockerfile/parser/line_parsers.go#L343-L369

But rules_docker has no dependency on moby library, it has a dependency on go-containerregistry, which has the same copy of the config: https://github.com/moby/buildkit/blob/master/frontend/dockerfile/parser/line_parsers.go#L343-L369.  

Yes, as far as I know, there is no parsing logic in go-containerregistry to transform a string into that config.  The parsing logic should be upstreamed there, in the same place that has the config.

svagner:
Ah, you meant about an unmarshaling line to the struct.  Yeah. Makes sense.  I'm not sure `go-containerregistry` will approve this change though since it looks like they don't need such a directive there. I can try to prepare PR for `go-containerregistry` for it but are we ready to add a health check directive here? I really just don't want to waste reviewers of  `go-containerregistry` time and my time if we don't want this feature at all.

> Also, rules_docker should not strive to replicate every feature in docker or a Dockerfile. A noteable absence is the RUN feature.

pcj:
If we can get a parsing function upstreamed and simplify this PR to add only a single `healthcheck` attribute, I would support it.

medzin:
@svagner will you continue this PR?

alexeagle:
Hey @jonjohnsonjr what do you think about the proposed upstream change to go-containerregistry?

jonjohnsonjr:
> what do you think about the proposed upstream change to go-containerregistry?

That sounds reasonable to me, something like `func ParseHealthConfig(s string) (*HealthConfig, error)`?

I'm happy to review a PR.

alexeagle:
@dgocoder I'm not sure if @svagner is still around to finish this work (by upstreaming the Go parts) or if it needs a new contributor to step up to finish.

nadavwe:
hey @pjjw, any luck?
We would really love to see this merged.
currently we need to have a base image created in a different build to include the healthcheck, quite far from the application definition... :/

pjjw:
no, have not plumbed this upstream yet, but this diff works for us as-is.

PR #1749, opened 2021-03-08
Allows container_push to postpone image download to build phase.

igorgatis:
Usage:
```starlark
container_pull(
    name = "base",
    registry = "gcr.io",
    repository = "my-project/my-base",
    digest = "sha256:deadbeef",
    lazy_download = True, # Will postpone download to build phase.
)
```

### DIFF
--- a/container/pull.bzl
+++ b/container/pull.bzl
@@ -96,6 +96,10 @@ _container_pull_attrs = {
         doc = "(optional) The tag of the image, default to 'latest' " +
               "if this and 'digest' remain unspecified.",
     ),
+    "lazy_download": attr.bool(
+        default = False,
+        doc = "(optional) Indicates whether donwload is postponed to build phase.",
+    ),
 }
 
 def _impl(repository_ctx):
@@ -120,9 +124,6 @@ def _impl(repository_ctx):
             puller = repository_ctx.attr.puller_linux_s390x
 
     args = [
-        repository_ctx.path(puller),
-        "-directory",
-        repository_ctx.path("image"),
         "-os",
         repository_ctx.attr.os,
         "-os-version",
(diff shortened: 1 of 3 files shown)


zoidbergwill:
We have some large images we pull in in our WORKSPACE, and fetching and creating the intermediate forms even when they're not used is a huge performance cost, so this'd be amazing to see added.

pcj:
Hi @igorgatis this *seems* useful but can you expand on the use case?  How would I adopt my existing bazel workflow using this flag?

You used the `metadata-only` for the puller but `lazy_download` for the attribute name.  Was this intended?

Needs a test (the good news is that the test for this feature will likely be the fast :) ).

container/pull.bzl
```diff
@@ -96,6 +96,10 @@ _container_pull_attrs = {
         doc = "(optional) The tag of the image, default to 'latest' " +
               "if this and 'digest' remain unspecified.",
     ),
+    "lazy_download": attr.bool(
+        default = False,
+        doc = "(optional) Indicates whether donwload is postponed to build phase.",
```
pcj, line 101: Spelling

igorgatis:
> Unfortunately, using the network during an action is considered bad practice and IIRC network access is disabled in many execution contexts, particularly remote execution.

I believe everybody agrees with that. However, the "best practice" makes the scenario of multiple container_pull calls in WORKSPACE impractical.

This is why the field lazy_download flag is false by default. User needs to be aware of pros-and-cons. Parhaps, user might make use of this flag only for arely used images or HUGE images.

> Also, what are the security implications of this change? Is it possible for an attacker to replace content in an image by deferring download to the build phase? Is hash integrity preserved?

We can mitigage this issue passing expected SHA to genrule call. It should be pretty easy to implement. Should I proceeed with that?

sluongng:
https://github.com/bazelbuild/rules_python#fetch-pip-dependencies-lazily-experimental

This might be a relevant reference

pcj:
Many organizations use an HTTP_CACHE or artifactory to mitigate this.  Can you not use that?

UebelAndre:
Not every organization has the flexibility to stand something like that up and maintain it. There's still a large startup cost in bigger repos when having to download all images when one or none will be used. I think with sufficient documentation this would be a fantastic feature.

pcj:
Perhaps we should discuss at next maintainer meeting.  @igorgatis @sluongng @UebelAndre would you be willing to join a call in a few weeks to go over this?

sluongng:
> Perhaps we should discuss at next maintainer meeting. @igorgatis @sluongng @UebelAndre would you be willing to join a call in a few weeks to go over this?

For sure. Please send and invite to sluongng at gmail dot com.

mindaugasrukas:
@igorgatis could you fix Buildifier error:
```
##### :bazel: buildifier: found 1 format issue in your WORKSPACE, BUILD and *.bzl files
--
  | Please download <a href="https://github.com/bazelbuild/buildtools/releases/tag/0.22.0">buildifier 0.22.0</a> and run the following command in your workspace:<br/><pre><code>buildifier container/pull.bzl</code></pre>
  | ##### :bazel: buildifier: found 1 lint issue in your WORKSPACE, BUILD and *.bzl files
```

container/pull.bzl
```diff
@@ -210,13 +219,44 @@ def _impl(repository_ctx):
...
+            "$(RULEDIR)",
+        ] + args
+
+        download_target = """
+genrule(
+    name = "download",
+    message = "Fetching",
+    outs = ["{outs}"],
+    cmd = "{args}",
+    tools = ["{tool}"],
```
dmivankov, line 245:
```suggestion
    tools = ["{tool}"],
    tags = ["requires-network"],
```
to make sure network is allowed even in build sandbox

dmivankov:
Another interesting thing is compatibility with private registries, for example `docker-credential-gcloud` helper wants writable `$HOME/.config/gcloud/logs/` and probably read access for cached tokens. Maybe "no-sandbox" (and maybe also "local") tag is required to make it work for private registries with credential helpers.

PR #1750, opened 2021-03-09
Invite maintainers from community

smukherj1:
FYI @hicksjoseph 

Interest for this has already been expressed in #1743

PR #1753, opened 2021-03-12
Revert "Invite maintainers from community (#1750)"

alexeagle:
We got the new maintainers here now, and don't need to advertise this way.

We'll surely add more maintainers from the community over time, but can bootstrap that the normal way by promoting ppl who are active and helpful in the repo.

This reverts commit a4d580ee357b3e40e80becb311ea88214b70e0db
and moves the config file to .github/ folder to improve tidiness in the root

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+* @smukherj1 @alexeagle @pcj @gravypod
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,1 +1,0 @@
-* @smukherj1 @hicksjoseph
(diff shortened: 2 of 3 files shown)


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+* @smukherj1 @davegay @eytankidron @alexeagle @pcj @gravypod
```
smukherj1, line 1: I suggest removing @davegay and @eytankidron here as they're no longer actively involved with rules_docker

smukherj1 (approved):

.github/CODEOWNERS
alexeagle: cool thx

PR #1754, opened 2021-03-12
Add issue and PR templates

alexeagle:
This helps a lot to make initial user interactions at a higher quality, so maintainers don't spend as much time asking for clarifications and so on.

See https://docs.github.com/en/github/building-a-strong-community/configuring-issue-templates-for-your-repository


gravypod (approved):
We may want to do some edits in the future. I'm not a huge fan of the <pre>...</pre> tags being used when we can just use codeblocks. This is still a step in the right direction though and we can always work from here as a base.

PR #1757, opened 2021-03-12
Introduce github action to mark stale issues/PRs

### LINKED ISSUE #1752: Setup probot curation of stale PRs/issues
In rules_nodejs we have https://github.com/bazelbuild/rules_nodejs/blob/stable/.github/workflows/stale.yml
that sets users expectations: an issue or PR that hasn't had activity in a long time is never going to get looked at.

In order of preference:
1. have enough maintainers to satisfactorily answer/close all issues and review/merge all PRs
2. have a bot say "sorry"
3. leave them all open forever

Since 1 is not feasible, I think we have to settle for 2.

alexeagle:
If an issue or PR goes 180 days without comments, and has no assignee, then add a nice warning comment and add a "Can Close?" label.
If it then goes another 30 days without comments or assignee, then close it with a nice comment.

Note that after this is merged, we'll need to also enable this in the Actions tab of the repo settings.

Closes #1752

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,59 @@
+# See https://github.com/marketplace/actions/close-stale-issues
+
+name: Mark stale issues and pull requests
+
+on:
+  schedule:
+  # run at 01:45 UTC daily
+  - cron: "45 1 * * *"
+
+jobs:
+  stale:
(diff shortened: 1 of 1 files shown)


gravypod (approved):
Looks like something very similar is used on rules_nodejs. Great to see the cross pollination!

PR #1767, opened 2021-03-12
Remove google cloud build integration

alexeagle:
It's redundant with the testing on Travis


pcj (approved):
Separate discussion, but travis may not be a good long term solution.

PR #1770, opened 2021-03-13
update_deps: Correctly check if manifest pull failed

gravypod:
This script finds reproducable hashes for pulling containers from
distroless. Since the last time this was run it seems that they
have stopped pushing a :debug tag for some images (static). This
makes sense as static is just an image with some standard openssl
and user file deps that do not have debug versions.

The code that implements update_deps is written to recover from
this situation (seemingly anticipating it) but it looks like the
author copied a Fatalf which os.Exit(1)s after printing. I've
changed this ti a Printf() so that the existing recovery logic
can take over.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [x] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/cc/cc.bzl
+++ b/cc/cc.bzl
@@ -19,8 +19,8 @@
 # git repository.
 
 DIGESTS = {
-    # "gcr.io/distroless/cc:debug" circa 2020-10-30 14:07 -0700
-    "debug": "sha256:7c31f56955cc66134efb3e8282c68f7c92ae3dddcfd99ebf0e7e3dcd9b686e4f",
-    # "gcr.io/distroless/cc:latest" circa 2020-10-30 14:07 -0700
-    "latest": "sha256:c4014bdecaede16f767f8cfc496f968736bc08632bf54a37c004820e85cd8209",
+    # "gcr.io/distroless/cc:debug" circa 2021-03-14 11:14 -0400
+    "debug": "sha256:2b083aa36dbfe76eff429ee21596d53a6ca5ab69ec48aef375264c267e3da5a0",
+    # "gcr.io/distroless/cc:latest" circa 2021-03-14 11:14 -0400
+    "latest": "sha256:937aa8c7edd91618d8ed4510a0fd06a9dd97db3a786b23e953213cdcf88ccfde",
 }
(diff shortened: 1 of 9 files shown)


pcj:
I guess I'm not fully understanding this.  It appears to me that the original logic is correct, which goes something like this:

- get the digest via crane for the latest image.  If any error occurs, die (can't continue without the latest digest)
- try and get the digest for the debug image.  If any error other than "MANIFEST_UNKNOWN" (aka Not Found) occurs, die (might be a network error, or some condition we don't know).  But if we know the image is "Not Found", fallback to the latest image.

The bad part about the original code was doing stringly typed error checking.  It would have been nicer if the crane library exposed a `ErrDigestNotFound` error, and one could do an `errors.Is(err, crane.ErrDigestNotFound)` check here rather than trying to interpret the error message string.  Do you know if that exists?

So, what is the error message that we're seeing now?  It's possible that the error message that is printed has changed that is messing with the logic here.

At a minimum, the message that would be printed after this change is not meaningful.  Either don't print the message at all, or make it more meaningful, something like `warning: the debug image digest for %s was not found, falling back to non-digest image %s`.

pcj (changes requested):
See message above

gravypod:
Nice catch @pcj! Unfortunately it doesn't seem like checking `errors.Is(err, crane.ErrDigestNotFound)` is going to be possible. The go-containerregistry package implements it's [error handling](https://github.com/google/go-containerregistry/blob/v0.1.4/pkg/v1/remote/transport/error.go#L45) in a way in which all errors share a single struct and provide no way to distinguish between them it seems. 

I used reflect to find the type I was actually getting there and it does seem to be the transport.Error package's implementation so nothing inbetween the error.go and their user-facing APIs clean up that error struct to be more usable. 


I've updated the commit message to reflect on what's changed.

pcj:
Looks good. I'd err on the side of caution and expand the check to `strings.Contains(err.Error(), "MANIFEST_UNKNOWN: Failed to fetch \"debug\" from request"`.

pcj (approved):
Recommend make the stringly-typed error check more specific (see message above).  Otherwise, go ahead and merge when you are ready.

Issue #1216, opened 2019-10-14
Document minimum version of Bazel required


mariusgrigoriu:
The README should document what version(s) of Bazel  are required to run rules_docker.  For example v0.8.0 uses `default_mtime` which appears to have been introduced with Bazel 0.24, but there was no mention that a newer version was required.

nlopezgi:
We don't do any backward compatibility testing with older versions of Bazel (nor do we have the time atm to set these up). Support for older versions of Bazel is not something we are doing explicitly. We always recommend you use the latest version of Bazel; at this point, if we document anything, it would be that we support 1.0.0 and anything else backwards that works is just by accident. Keeping open to consider documenting this explicitly.

alexeagle:
I'd like to add that backwards compat testing with a new focus on bazel-integration-testing, maybe something we can pick up.

PR #1782, opened 2021-03-15
Add .bazelversion file

alexeagle:
This pins what version we use for our own local builds. Note that #1216 is the issue for testing against other bazel versions.


pcj:
Assuming you are relying on the existing tests to assert this is valid?  Did you do any local testing with this?

alexeagle:
replaced with #1873

PR #1786, opened 2021-03-17
Revert "Setup assign-size-label action"

gravypod:
Reverts bazelbuild/rules_docker#1762

Seems to be failing a bit.


pcj:
cc @alexeagle

alexeagle:
SGTM I don't think it's useful enough to even spend time diagnosing why the auth is flaky

alexeagle (approved):

Issue #1664, opened 2020-10-27
[puller] Feature request: support registry-mirrors global configuration


or-shachar:
Recently I had to change a lot of my `container_pull` definitions to prepare for this:

> As of November 1, 2020, Docker Hub [rate limits](https://www.docker.com/blog/scaling-docker-to-serve-millions-more-developers-network-egress/) apply to unauthenticated or authenticated pull requests on the Docker Free plan. To avoid disruptions and have greater control over your software supply chain, you can migrate your dependencies to Container Registry or Artifact Registry.

Basically I replaced:
```python
container_pull(
    name = "redis_latest",
    registry = "index.docker.io",
    repository = "library/redis",
    tag = "latest"
  )
```
with
```python
container_pull(
    name = "redis_latest",
    registry = "mirror.gcr.io",
    repository = "library/redis",
    tag = "latest"
  )
```

I found it a little awkward that I can't configure `registry-mirrors` globally by global bazelrc.
In non-bazel builds that use native docker client they just added this global config to their docker daemon:
```json
{
  "registry-mirrors": ["https://mirror.gcr.io"]
}
```

I know we love everything to be defined in our committed source code but in this case - I think there's a great value in allowing global definition.  

BTW: I see that `docker_toolchain_configure` has `client_config` property - but I'm pretty sure it's not used in case of the puller (am I wrong?). Maybe the best case would be to read that property directly from the client_config.

smukherj1:
Hi @or-shachar,

Generally you're right that injecting files whose contents aren't part of the source tree is an anti-pattern. However, if the docker client supports this, it makes sense for rules_docker to support this as well.

Note that client_config is a standard docker client configuration JSON and I would make rules_docker accept arbitrary fields from this file unless docker recognizes it as well. Put another way, if docker client doesn't expect `registry-mirrors` in the client config JSON, I wouldn't update rules_docker to expect it in that file. OTOH, if the client config JSON is where docker client reads the registry mirrors from, I'm open to having rules_docker support it as well.

I'm open to PR(s) that implements this. It's possible part of the implementation needs to be in go-containerregistry.

rmohr:
#1789 would solve our use-case. @or-shachar would it solve yours too?

sluongng:
Just FYI we have had a lot of discussion around the approach to solve this in Bazel Build Slack

https://bazelbuild.slack.com/archives/CA3NW13MH/p1618411736057600
[...]

PR #1789, opened 2021-03-23
Allow configuring a http proxy exclusively for the image puller

rmohr:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: related to #1664

In order to support proxy-mirrors for various container registries, one can configure http proxies for various container runtimes like docker or cri-o by simply exporting `HTTP_PROXY` and `HTTPS_PROXY`. This is for instance more powerful than explicitly defining mirror repositories for the docker runtime, like described in #1664 and allows the usage of mirrors like https://github.com/rpardini/docker-registry-proxy. We use this in kubevirt as a general precaution to protect our CI system from outages and disappearing images from various registries.

The puller right now already picks up `HTTP_PROXY` and `HTTPS_PROXY`, but this is now a all-or-nothing knob for bazel. One can't tell bazel to use the proxy only for image pulls and ignore it for the rest.

## What is the new behavior?

Add `CONTAINER_PULL_HTTP_PROXY` and `CONTAINER_PULL_HTTPS_PROXY` env variables which can be set in a bazelrc file via action envs, without interfering with the rest of bazel.

- [ ] Yes
- [x] No

### DIFF
--- a/README.md
+++ b/README.md
@@ -1318,6 +1318,11 @@ variable. The caching feature hasn't been thoroughly tested and may be thread
 unsafe. If you notice flakiness after enabling it, see the warning below on how
 to workaround it.
 
+**NOTE** Set `CONTAINER_PULL_HTTP_PROXY` and `CONTAINER_PULL_HTTPS_PROXY` to
+use a http proxy exclusively for the image puller. If the `CONTAINER_PULL`
+variables are not set, the default proxy handling from bazel (automatically
+picking up variables like `HTTP_PROXY`) applies.
+
 **NOTE:** `container_pull` is suspected to have thread safety issues. To
 ensure multiple `container_pull`(s) don't execute concurrently, please use the
 bazel startup flag `--loading_phase_threads=1` in your bazel invocation.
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1318,6 +1318,9 @@ variable. The caching feature hasn't been thoroughly tested and may be thread
 unsafe. If you notice flakiness after enabling it, see the warning below on how
 to workaround it.
 
+**NOTE** Set `CONTAINER_PULL_HTTP_PROXY` and `CONTAINER_PULL_HTTPS_PROXY` to use
```
gravypod, line 1321: Can we leave a note that the standard HTTP_PROXY and HTTPS_PROXY environment variables are still supported?

gravypod (changes requested):

README.md
rmohr: Done. How is it now?

rmohr:
Everything should be good now. What I do not understand, is how the puller is then actually updated once this change is in. Do I need to do something?

rmohr:
@gravypod let me know if I should change anything else.

rmohr:
I would still be interested in adding this. I still think it is a good additional feature for transparent proxies.

PR #1796, opened 2021-03-31
replace "tar A" with more portable command

ptarjan:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Tests fail on OSX with this message:
```
Usage:
  List:    tar -tf <archive-filename>
  Extract: tar -xf <archive-filename>
  Create:  tar -cf <archive-filename> [filenames...]
  Help:    tar --help
```

BSD `tar` doesn't seem to have `-A`. But `tar` archives are just concatenatable, so I think `cat` is enough here

Issue Number: N/A

## What is the new behavior?
```
bazel test tests/docker/package_managers/...
```
now passes on my laptop

- [ ] Yes
- [x] No

### DIFF
--- a/docker/toolchain_container/toolchain_container.bzl
+++ b/docker/toolchain_container/toolchain_container.bzl
@@ -145,8 +145,7 @@ def _language_tool_layer_impl(
         ctx.actions.run_shell(
             inputs = installables_tars,
             outputs = [final_installables_tar],
-            command = "tar cvf {output_tar} --files-from /dev/null && \
-        for i in {input_tars}; do tar A --file={output_tar} $i; done".format(
+            command = "cat {input_tars} > {output_tar}".format(
                 output_tar = final_installables_tar.path,
                 input_tars = " ".join(installables_tars_paths),
             ),


pcj:
Please rebase your last PR.

docker/toolchain_container/toolchain_container.bzl
```diff
@@ -145,8 +145,7 @@ def _language_tool_layer_impl(
         ctx.actions.run_shell(
             inputs = installables_tars,
             outputs = [final_installables_tar],
-            command = "tar cvf {output_tar} --files-from /dev/null && \
-        for i in {input_tars}; do tar A --file={output_tar} $i; done".format(
+            command = "cat {input_tars} > {output_tar}".format(
```
alexeagle, line 148: wow, I didn't know tars were a delimited record format. according to https://superuser.com/questions/941475/concatenate-multiple-tar-files-in-one-command/941552 it means that anyone untarr'ing this thing has to pass a flag. How do we tell if this is a breaking change? Also I'm not sure if there are tests in this repo that cover such a potential breakage, what happens if you intentionally break this spot (like put nothing in the output_tar) - does something go red?
Sorry to require such poking around :(

ptarjan: I think `-A` still makes archives that require you to use `-i` anyways. From https://www.manpagez.com/man/1/gnutar/

```
       -i, --ignore-zeros
              Ignore  zeroed  blocks in archive.  Normally two consecu-
              tive 512-blocks filled with zeroes mean EOF and tar stops
              reading  after  encountering them.  This option instructs
              it to read further and is useful  when  reading  archives
              created with the -A option.
```

There are many tests that do indeed fail if I screw this up (which is why I found it, since my laptop's `tar A` doesn't work).

A more compatible thing here would be to untar them all somewhere and then re-tar, but that would be more expensive. Want me to do that?

alexeagle: My only concern is with backwards compat. If we can tell a compelling story that users had to do the same thing before to use these tars as they'll have to do in the next release, we are all good

ptarjan: Well the docs lied....
```
$ echo foo > foo
$ echo bar > bar
$ tar cvf foo.tar foo
foo
$ tar cvf bar.tar bar
bar
$ tar cvf foobar.tar --files-from /dev/null
$ tar A --file=foobar.tar foo.tar
$ tar A --file=foobar.tar bar.tar
$ tar tvf foobar.tar
-rw-r--r-- root/root         4 2021-03-31 23:42 foo
-rw-r--r-- root/root         4 2021-03-31 23:42 bar
$ cat foo.tar bar.tar > foobar-cat.tar
$ tar tvf foobar-cat.tar
-rw-r--r-- root/root         4 2021-03-31 23:42 foo
```

alexeagle: Okay so that means we can't merge this until the next major release?

ptarjan:
I don't think I care enough to push this through. As it stands, you can't run this on BSD because it doesn't have `-A`

Issue #1814, opened 2021-04-24
Incorrect sha when fetching `com_github_google_go_containerregistry`


SrodriguezO:
# 🐞 bug report

### Affected Rule

The `com_github_google_go_containerregistry` external dep loaded by `container_deps()`

### Is this a regression?
This started happening out of the blue.

### Description
We started noticing these errors in our CI this evening:
```
ERROR: An error occurred during the fetch of repository 'com_github_google_go_containerregistry':
    Traceback (most recent call last):
File "/.../.cache/bazel/.../external/bazel_gazelle/internal/go_repository.bzl", line 31, column 33, in _go_repository_impl
 		ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://api.github.com/repos/google/go-containerregistry/tarball/8a2841911ffee4f6892ca0083e89752fb46c48dd] to ... Checksum was cadb09cb5bcbe00688c73d716d1c9e774d6e4959abec4c425a1b995faf33e964 but wanted 60b9a600affa5667bd444019a4e218b7752d8500cfa923c1ac54ce2f88f773e2
```

Downloading the artifact directly confirms the hash is now different from that expected in [go_repositories.bzl](https://github.com/bazelbuild/rules_docker/blob/master/repositories/go_repositories.bzl) 

I can't find any info on this artifact online, so I haven't been able to determine if the change is expected and legitimate. I suspect this will cause issues for other rules_docker users. 

## 🔬 Minimal Reproduction

Add rules_docker to the workspace as shown in the README and run a build that uses the rules.

## 🔥 Exception or Error

```
ERROR: An error occurred during the fetch of repository 'com_github_google_go_containerregistry':
    Traceback (most recent call last):
File "/.../.cache/bazel/.../external/bazel_gazelle/internal/go_repository.bzl", line 31, column 33, in _go_repository_impl
 		ctx.download_and_extract(
Error in download_and_extract: java.io.IOException: Error downloading [https://api.github.com/repos/google/go-containerregistry/tarball/8a2841911ffee4f6892ca0083e89752fb46c48dd] to ... Checksum was cadb09cb5bcbe00688c73d716d1c9e774d6e4959abec4c425a1b995faf33e964 but wanted 60b9a600affa5667bd444019a4e218b7752d8500cfa923c1ac54ce2f88f773e2
```

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu 20.04
  </code>
[...]

PR #1815, opened 2021-04-24
Update SHA for go-containerregistry

umeshkumhar:
update sha for com_github_google_go_containerregistry

## PR Checklist

Please check if your PR fulfils the following requirements:

- [x] Tests for the changes have been added (for bug fixes/features)

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Breaking http_archive or git_repository due to SHA change of @com_github_google_go_containerregistry

Issue Number: 1814

## What is the new behaviour?
Will pull tarball @com_github_google_go_containerregistry and verify SHA properly

- [ ] Yes
- [x] No

## Other information

https://github.com/bazelbuild/rules_docker/issues/1814


umeshkumhar:
@googlebot I signed it!

alexeagle (approved):

alexeagle:
Looks like it was broken by https://github.com/bazelbuild/continuous-integration/commit/15fbac414e5065bff3b46e2ace5e119c4704a72c
rules_docker asks for buildifier 0.22.0 which is very old, and its command line API is not compatible with how Bazel CI setup invokes it.
I'll update us to current.

repositories/go_repositories.bzl
```diff
@@ -39,7 +39,7 @@ def go_deps():
             name = "com_github_google_go_containerregistry",
             urls = ["https://api.github.com/repos/google/go-containerregistry/tarball/8a2841911ffee4f6892ca0083e89752fb46c48dd"],  # v0.1.4
             strip_prefix = "google-go-containerregistry-8a28419",
-            sha256 = "60b9a600affa5667bd444019a4e218b7752d8500cfa923c1ac54ce2f88f773e2",
+            sha256 = "cadb09cb5bcbe00688c73d716d1c9e774d6e4959abec4c425a1b995faf33e964",
```
SrodriguezO, line 42: I'm not opposed to updating the sha, but is there nay concern the artifact was tampered with?

I'm not sure if this is relevant, but neither sha matches the shas of the artifacts for that version in the regular non-api project: https://github.com/bazelbuild/rules_docker/issues/1814#issuecomment-826096584

cc @alexeagle

SrodriguezO: @alexeagle @umeshkumhar thoughts on this? Do we know why the sha suddenly changed? And why it differs from those in the repo's release page?

umeshkumhar: @SrodriguezO  Not 100% sure, but this issue is raised with this [commit ID](https://github.com/google/go-containerregistry/commit/0976a278d3ebefb47fdbf5a2ff2b3e8a405c9e89). In this PR, you can see ko version is changed.

There are some reported issues where ko on version change has changed the tar hashes, which impacted SHA checks..More info here:
https://github.com/kubernetes/kubernetes/issues/99376

One another possibility can be that the maintainer has re-published/ updated tarball, which caused SHA change, which I don't think have happened.

I have not explored to the root cause. Let me know if that helps

uhthomas: Hi,

I'm facing an issue where using rules_docker v0.16.0 results in:

```
Checksum was cadb09cb5bcbe00688c73d716d1c9e774d6e4959abec4c425a1b995faf33e964 but wanted 60b9a600affa5667bd444019a4e218b7752d8500cfa923c1ac54ce2f88f773e2
```

The root cause is explained in https://github.com/google/ko/issues/315 and https://github.com/kubernetes/kubernetes/issues/99376.

The TL;DR is that Kubernetes was using short hashes for versions, which may change in length over time. The hash is injected into the source through `git archive`.

This SHA256 change _should_ fix things for now, but will eventually break again. The real solution will be to update the dependency to a commit after https://github.com/kubernetes/kubernetes/commit/ab221b5d9f98833da2cbb5754fc2e7685bf80298.

uhthomas: Further, my solution to this problem, without updating from rules_docker v0.16.0, has been to explicitly pull in github.com/google/go-containerregistry v0.4.1 via rules_go.

See https://github.com/uhthomas/6f.io/compare/8548f56df648693c0e518454cfccad0792e48eb8..3915a86d6bd9c6a376198b1d1d79979a7d4c542f

PR #1829, opened 2021-04-26
Update go-containerregsitry dependency to 0.4.1

ImJasonH:
## PR Checklist

- [n] Tests for the changes have been added (for bug fixes / features)
- [n] Docs have been added / updated (for bug fixes / features)

## PR Type

- [x] Bugfix
- [ ] Feature
- [x] Build related changes
- [ ] CI related changes

Issue Number: #1814

## What is the new behavior?

This updates the go-containerregistry dependency from [0.1.4](https://github.com/google/go-containerregistry/releases/tag/v0.1.4) (October 2020) to [0.4.1](https://github.com/google/go-containerregistry/releases/tag/v0.4.1) (March 2021)

This removes a dependency on a non-deterministic package resulting in inconsistent archive SHAs generated by GitHub's archive server. That upstream issue was also fixed in https://github.com/kubernetes/kubernetes/issues/99376.

Other than removing this non-determinism, no major functional changes are expected. Some performance improvements and bugfixes have been added since 0.1.4, which might be beneficial.

- [ ] Yes
- [x] No

### DIFF
--- a/repositories/go_repositories.bzl
+++ b/repositories/go_repositories.bzl
@@ -37,11 +37,8 @@ def go_deps():
     if "com_github_google_go_containerregistry" not in excludes:
         go_repository(
             name = "com_github_google_go_containerregistry",
-            urls = ["https://api.github.com/repos/google/go-containerregistry/tarball/8a2841911ffee4f6892ca0083e89752fb46c48dd"],  # v0.1.4
-            strip_prefix = "google-go-containerregistry-8a28419",
-            sha256 = "cadb09cb5bcbe00688c73d716d1c9e774d6e4959abec4c425a1b995faf33e964",
+            commit = "efb2d62d93a7705315b841d0544cb5b13565ff2a",
             importpath = "github.com/google/go-containerregistry",
-            type = "tar.gz",
         )
     if "com_github_pkg_errors" not in excludes:
         go_repository(


steeve:
Thank you for being so reactive, but 👎 

Github branch/commit tarballs are dynamically generated. Don't depend on them having a stable SHA256. If the upstream rule is not distributing built artefacts, you'll need to use git_repository or go_repository or whatever rule that is stable.

In this case, use `go_repository`'s `commit` and `remote` to pin a specific commit.

ImJasonH:
I've updated this PR to use `commit` and `importpath` for go-containerregistry.

I haven't touched the other dependencies, but I can update those if you'd like.

steeve:
Not my call, but I feel this could be a good idea. Or let go_repository pull them as in module mode maybe?

alexeagle (approved):

gravypod (approved):

uhthomas:
The `go_repository` target is missing the `sum` attribute. The module downloaded through this rule will not be verified and poses a security risk.

steeve:
> I've updated this PR to use `commit` and `importpath` for go-containerregistry.
> 
> I haven't touched the other dependencies, but I can update those if you'd like.

It is not downloaded as a module, but via a git commit, which is safe.

steeve:
> The `go_repository` target is missing the `sum` attribute. The module downloaded through this rule will not be verified and poses a security risk.

It is not downloaded as a module, but via a git commit, which is safe.

uhthomas:
> > The `go_repository` target is missing the `sum` attribute. The module downloaded through this rule will not be verified and poses a security risk.
> 
> It is not downloaded as a module, but via a git commit, which is safe.

Downloading an archive just based on a git commit is _not_ safe. All content must be verified, whether it be through the `sum` or `sha256` attributes.

PR #1832, opened 2021-04-27
Create docs/ folder

alexeagle:
Move the existing starlark target there, and check in the generated output.
Uses generated_file_test since that prints the bazel run command to accept the new file,
making this workflow easy for novice contributors to follow.

Also don't make users install the stardoc repository anymore, now that they don't load() from this BUILD file they shouldn't need it.


UebelAndre:
It'd be great if the docs directory was a separate workspace which doubled as a visibility test.

README.md
```diff
@@ -1758,459 +1758,7 @@ A rule that assembles data into a tarball which can be use as in `layers` attr i
...
-  <tbody>
-    <tr>
-      <td><code>@io_bazel_rules_docker//toolchains/docker:toolchain_type</code></td>
-      <td>
-        See <a href="toolchains/docker/readme.md#how-to-use-the-docker-toolchain">How to use the Docker Toolchain</a> for details
-      </td>
-    </tr>
-  </tbody>
-</table>
+**MOVED**: See generated API documentation in docs/image.md
```
gravypod, line 1761: Can we make docs/image.md a link?

PR #1835, opened 2021-04-29
WIP: more containers stardoc stuff

zoidbergwill:
TODO: the difficult part of making sure macros / attrs have the accurate docs from the README and we're only exporting docs for the useful rules

Follow up of https://github.com/bazelbuild/rules_docker/pull/1832


alexeagle:
Let's keep doing just one at a time? Otherwise it's overwhelming to look through the diffs.

https://github.com/bazelbuild/rules_docker#container_bundle-1
has docs for `images` and has a rule docstring but
https://github.com/bazelbuild/rules_docker/blob/12e0f2a8fb32e702ad1dd99490179cca38c1eb96/docs/bundle.md
doesn't, for example.

alexeagle:
Hi, still have any time for this? Can I help get it unstuck? I'm keen to finish migrating the docs to stardoc (just approved a PR that makes manual readme changes which are out-of-date from the code :(

alexeagle:
I think I replaced this with 454981e

PR #1836, opened 2021-04-30
Propagate tags to app-layers in docker images for all langs

vsbus:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

We are using `py3_image()` with `layers` parameter in one of our projects. These images have `tags = ["manual"]` so we don't expect them to build with `bazel build //...`. However, we observing that `.tar` files with layers are still built. That happened because tags were not propagated to the layer's targets. This PR is trying to fix the issue.

Issue Number: N/A

## What is the new behavior?
`bazel build //...` doesn't build extra targets like layers when manual tag is used.

- [ ] Yes
- [x] No

### DIFF
--- a/cc/image.bzl
+++ b/cc/image.bzl
@@ -85,12 +85,12 @@ def cc_image(name, base = None, deps = [], layers = [], binary = None, **kwargs)
         fail("kwarg does nothing when binary is specified", "deps")
 
     base = base or DEFAULT_BASE
+    tags = kwargs.get("tags", None)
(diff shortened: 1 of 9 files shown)


vsbus:
Hey @gravypod , please take another look

gravypod (approved):
Looks good. Lets just get a second review from @alexeagle since this will be a user-visible change.

alexeagle (approved):
thanks!

PR #1851, opened 2021-05-12
Allow excluding external dependencies explicitly

vmax:
## PR Checklist

- [X] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Feature
- [x] Build related changes
- [ ] CI related changes

Only those external dependencies that were _already_ present can be excluded. Therefore, to load custom external repository, you'd have to put the definition **before** loading `rules_docker`

Issue Number: N/A

## What is the new behavior?

Allow excluding external dependencies explicitly

- [ ] Yes
- [X] No

Exclusion list is empty by default and combined with the old list, so no breaking changes are introduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,6 +152,8 @@ load(
     "@io_bazel_rules_docker//repositories:repositories.bzl",
     container_repositories = "repositories",
 )
+# OPTIONAL: pass a list of external repos you want excluded because they will be loaded later:
+# container_repositories(excludes = ["rules_pkg"])
 container_repositories()
 
 load("@io_bazel_rules_docker//repositories:deps.bzl", container_deps = "deps")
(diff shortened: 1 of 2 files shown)


alexeagle:
adding a second way to do it is overall more complicated, and i've seen most rulesets use this "first definition wins" semantics when a dependency appears twice in WORKSPACE
could you explain why you couldn't just specify the override first in your case?

vmax:
@alexeagle we're trying to maintain a consistent and predictable ordering in WORKSPACE for 10+ repos and only accidentally discovered that `@rules_pkg` was being loaded from `rules_docker` (as it's the first definition) instead of our own rules repo ([`@vaticle_bazel_distribution`](https://github.com/vaticle/bazel-distribution)). So main reason is adding explicitness into the definitions while maintaining the already-present ordering.

alexeagle:
IMO that's a real problem, but it's a design problem with Bazel's WORKSPACE semantics. I think adding this attribute to rules_docker isn't a good solution - not discoverable, different from what other rulesets do, gives multiple ways to do the same thing.
Feels to me like what you really needed in that scenario was for Bazel to print where rules_pkg was coming from. I think there's some activity on an upstream issue for that though I couldn't find it just now in a search.
@pcj any wisdom from your work in bzl of locating all repository definitions?

Issue #1802, opened 2021-04-07
entrypoint / launcher in nodejs_image isn't respected


martaver:
# 🐞 bug report

### Affected Rule

The problem is with `nodejs_image`.

### Is this a regression?

I'm not sure, I've only recently started with bazel.

### Description

I am trying to run a script prior to executing the nodejs binary, in order to initialise some system dependencies that I installed on a custom base image.

The 'dockerfile' way to do this is to use an ENTRYPOINT script described here:
https://stackoverflow.com/questions/41512237/how-to-execute-a-shell-command-before-the-entrypoint-via-the-dockerfile

...and then call node with CMD

However, `nodejs_image` seems to use CMD for `args`, and ENTRYPOINT for the `nodejs` binary.

Also when I provide it with an alternative `entrypoint` or a `launcher`, the resulting docker image ignores both, and the ENTRYPOINT remains stuck at: `"/app/path/to/myapp"`

The net result of this is that I can't configure 'launcher scripts' and use `nodejs_image` at the same time. 

## 🔬 Minimal Reproduction

Clone:
https://github.com/cleric-sh/cleric-bazel

Run:
```
bazelisk run //services/messenger:image
```
Then:
```
docker image inspect bazel/services/messenger:image
```

Not that the launcher defined in `images/chrome/BUILD`:`34` is left out.

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
OSX 11.2.3
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
Bazelisk version: v1.7.5
Build label: 4.0.0
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Jan 21 07:38:50 2021 (1611214730)
Build timestamp: 1611214730
Build timestamp as int: 1611214730
  </code>
</pre>

**Rules_docker version:**

<pre>
  <code>
v0.16.0
  </code>
</pre>

**Anything else relevant?**

PR #1854, opened 2021-05-18
Allow launcher in nodejs

martaver:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_docker/issues/1802

"entrypoint / launcher in nodejs_image isn't respected "

## What is the new behavior?

The `launcher` and `launcher_args` attributes are now accepted by `nodejs_image` and passed to `app_layer`.

- [ ] Yes
- [x] No

## Other information

This feature isn't really node specific. If maintainers agree, then I would update all the language images that use `app_layer` to accept `launcher` and `launcher_args`. Then I would also remove the documentation change from `nodejs_image` and instead make a general note that all the language images support the attributes.


run-bazel-in-docker.sh
```diff
@@ -0,0 +1,11 @@
+mkdir -p /tmp/build_output/
```
alexeagle, line 1: did you mean to check in this file? I guess it's for testing something. If so, maybe add some comments here about what it's used for and how to use it

alexeagle (commented):
This is awesome thanks!

You asked on slack about whether this should cover other languages, I don't really know...

Before we merge this, I have one question though. Would it be hard for a user to just skip using the convenient language rules like `nodejs_image` and just make their own `container_image` in order to get access to these attributes? maybe the use case is narrow enough that omitting them from the easy-to-use language wrappers is actually a feature to make things simpler for common users.

run-bazel-in-docker.sh
martaver: Added some docs... I created this script because the repo's tests fail on non-linux systems (I'm on MacOS):
```
tests/container/nodejs/nodejs_image_custom_binary_test.image: line 208: /usr/bin/docker: No such file or directory
```

martaver:
re: rolling your own `container_image`... I'd rather avoid that... `nodejs_image` forms a standard that I want to be able to lean on wherever possible. Deviating from it should be a last resort.

I would argue that launchers are common (roughly 50% of the images I deal with have them) and a minor enough implementation detail that it shouldn't warrant having to roll a custom `container_image`.

At the end of the day, there are basically only two ways to customise a `container_image`... change the base image, and the launch script / entry point. `nodejs_image` supports the former already, but not the latter.

This is also why I suspect it'd be good form to adder `launcher` and `launcher_args` to all the lang images.

alexeagle:
I tried to rebase this for you to land it, but I think you disabled the checkbox "allow edits from maintainers" - could you rebase yourself or check the box?

alexeagle (approved):

martaver:
Mmmmm done... looks like `.gitignore` had a conflict...

What did you think of exposing the launcher args for each language image btw?

alexeagle (approved):
If the same problem exists for other languages, then yeah we should fix it for all of them.

I'm okay with getting this fix in first, I don't think it's practical to make you do all of them at once.

PR #1856, opened 2021-05-20
Don't upload large files to remote-cache

### LINKED ISSUE #1855: container rules put high load on remote cache
I'm following up from a severe outage caused by a remote cache overload.

rules_docker tries to be careful to use layer tars as the primary outputs of rules, which is great. However there are lots of cases where an entire OCI image is the output of a rule, and this causes a ton of load on remote cache. In particular, we have some actions that run something fast (like `gzip`) to transform some large files into some other large files. These action results are not appropriate to send to the remote cache because under most conditions it's faster to perform the action again locally than to fetch from the remote cache.

As an illustration, take the example from the top of rules_docker README

`WORKSPACE`
```starlark
load("@io_bazel_rules_docker//container:container.bzl", "container_pull")

container_pull(
  name = "java_base",
  registry = "gcr.io",
  repository = "distroless/java",
)
```

Then I just build that image (no BUILD files required for this repro!)

By instrumenting the remote cache to reject uploads over 10MB (see https://github.com/buchgr/bazel-remote/issues/440) I can see that we attempt to upload *three* such blobs:

```
% bazel build @java_base//image --remote_cache=grpc://localhost:9092 --remote_retries=1 --execution_log_json_file=/tmp/execution_log.json
...
WARNING: Writing to Remote Cache:
Error while uploading artifact with digest '5d09c2db1d761a6cd292a453815df5fecfa98058e4f2c0185976d1f28dfc09ec/17786880'
WARNING: Writing to Remote Cache:
Error while uploading artifact with digest 'b04be07e97435bfe3590f9135809b12de5573155d9814e9c0bbaa3b56d7234ac/170127360'
WARNING: Writing to Remote Cache:
Error while uploading artifact with digest '1a3910ac3ab501fb4ed8210223ff77c5529b2d23aa737ce47161a812a2b185ad/201798144'
Target @java_base//image:image up-to-date:
  bazel-bin/external/java_base/image/image.tar
```
That's a 17MB object, a 162MB object, and a 192MB object being stored in the cache for a build that depends on the java_base image.

I think perhaps we should audit some of the actions that are expected to produce whole-image outputs, and mark these with `tags=['no-remote-cache']`

alexeagle:
Fixes #1855

### DIFF
--- a/skylib/zip.bzl
+++ b/skylib/zip.bzl
@@ -50,6 +50,12 @@ def _gzip(ctx, artifact, out, decompress, options, mnemonic):
             ) + (options or []),
             inputs = [artifact],
             outputs = [out],
+            execution_requirements = {
+                # This action produces large output files, but doesn't require much CPU to compute.
+                # It's not economical to send this to the remote-cache, instead local cache misses
+                # should just run gzip again.
+                "no-remote-cache": "1",
+            },
             mnemonic = mnemonic,
             tools = ctx.attr._zipper[DefaultInfo].default_runfiles.files,
         )
@@ -69,6 +75,13 @@ def _gzip(ctx, artifact, out, decompress, options, mnemonic):
             input_manifests = input_manifests,
             inputs = [artifact],
             outputs = [out],
+            execution_requirements = {
+                # See comment above.
+                # In this case it's not guaranteed that the user's gzip implementation
+                # is actually faster than a network fetch, but we guess that any typical
+                # one should be.
+                "no-remote-cache": "1",
+            },
             use_default_shell_env = True,
             mnemonic = mnemonic,
             tools = tools,
--- a/container/layer_tools.bzl
+++ b/container/layer_tools.bzl
@@ -183,6 +183,12 @@ def assemble(
         arguments = [args],
         tools = inputs,
         outputs = [output],
+        execution_requirements = {
+            # This action produces large output files, but doesn't require much CPU to compute.
+            # It's not economical to send this to the remote-cache, instead local cache misses
+            # should just run join_layers again.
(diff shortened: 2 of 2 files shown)


skylib/zip.bzl
```diff
@@ -50,6 +50,12 @@ def _gzip(ctx, artifact, out, decompress, options, mnemonic):
             ) + (options or []),
             inputs = [artifact],
             outputs = [out],
+            execution_requirements = {
```
gravypod, line 53: Should we add this same change to the `ctx.actions.run_shell` below?

gravypod (changes requested):
Looks good.

skylib/zip.bzl
alexeagle: yeah I was on the fence about it since it's harder to be sure that a user's gzip is actually behaving according to our "CPU load not high enough to be economical".
But in practice they are probably all gzip-equivalent programs

gravypod (approved):

sashamor:
Is there a way to disable this? We really do want to use remote-cache for our bazel Docker builds, our cache is fast and building locally each time takes way too long.

alexeagle:
Which action is slow? Join_layers? Maybe we should just revert that one.

sashamor:
Yes. I think reverting that one is a good idea. Our bazel cache is a giant bucket and we'd prefer not to rebuild anything if possible; can't see a way to force the join_layers to cache if this flag is set.

I also noticed that the apt rules also have remote-cache disabled, these are also slow and should be cached IMO.

Thanks for the prompt response.

sashamor:
Thank you! 😊

PR #1859, opened 2021-05-22
Migrate CI targets to buildkite

pcj:
This PR updates the presumit.yml off the legacy format and migrates most targets to buildkite.  

Targets not able to be run on buildkite are tagged as manual.

Macos CI will be re-enabled in a separate diff.


gravypod (approved):
Looks like a great state. Thanks!

PR #1869, opened 2021-05-23
Enable buildifier skylark-comment

pcj:


gravypod (approved):
Haha I like this lint rule.

PR #1871, opened 2021-05-24
More fixes for actions that shouldn't upload to remote-cache

alexeagle:
Similar to #1856. I found these by patching that PR and running through a big monorepo and searching execution log for more remote cache uploads over 10MB

### DIFF
--- a/container/layer.bzl
+++ b/container/layer.bzl
@@ -151,6 +151,11 @@ def build_layer(
         arguments = [args],
         tools = files + file_map.values() + tars + debs + [manifest_file],
         outputs = [layer],
+        execution_requirements = {
+            # This action produces large output files, so it's not
+            # economical to send this to the remote-cache
+            "no-remote-cache": "1",
+        },
         use_default_shell_env = True,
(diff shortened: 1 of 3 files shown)


gravypod (approved):
I think this is the right direction but I think we need to define more generally what *should* and *shouldn't* be cached by our rules. We might want to also run some tests against RBE/remote cache environments like what you're likely doing here.

alexeagle:
By the way for anyone who wants to figure out whether an execution log shows big artifacts from rules_docker, here's the command I was running. I'm excluding the mnemonics that are now `no-remote-cache` and see what else is over 10MB

`jq --raw-output 'select(.mnemonic != "RunAndCommitLayer" and .mnemonic != "ImageLayer" and .mnemonic != "GUNZIP" and .mnemonic != "GZIP" and .mnemonic != "JoinLayers") | .actualOutputs[] | select(.digest.sizeBytes | tonumber > 10485760) | [.path,.digest.sizeBytes] | @tsv' /Users/alex.eagle/Desktop/bazel_exec.oYGl.json`

PR #1892, opened 2021-06-15
Fix path handling in archive.py to always be posix and not native paths

luna-duclos:
## PR Checklist

No tests have been changed or added, but existing tests should validate the new behaviour still results in valid images

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

Currently, rules_docker when executed on windows will build a tar file with `\` characters in the tar, this works fine when opened with tools like 7z, however, when extracting using a typical `tar` tool (or when docker does it), the files end up mangled with <dirname>\<filename>` all jumbled up together in their file name

This is because archive.py makes a lot of assumptions about posix paths using `/` characters, but then bundles this up with also using `os.path` instead of `posixpath`

## What is the new behavior?

This fixes this problem by using `posixpath` in `archive.py`

- [ ] Yes
- [x] No

## Other information

I realize windows support is spotty and partial at best, but this gets us going :)

### DIFF
--- a/container/archive.py
+++ b/container/archive.py
@@ -19,6 +19,7 @@ import io
 import os
 import subprocess
 import tarfile
+import posixpath
 
 # Use a deterministic mtime that doesn't confuse other programs.
 # See: https://github.com/bazelbuild/bazel/issues/1299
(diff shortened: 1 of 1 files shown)


luna-duclos:
Would love some help figuring out whats failing on the mac

luna-duclos:
Aaand tests are green! Good to review :)

gravypod (approved):
Looks good. Thanks for doing this!

PR #1896, opened 2021-06-17
Polish output of package_manager rules

kaoet:
Set DEBIAN_FRONTEND to noninterative to suppress "unable to initialize frontend" warnings.

Use `docker create` + `docker start -a` to replace `docker run -d` + `docker attach`.
In the past, because the container is started before running `docker attach`, there will be a few lines at the beginning missing from logs. That will lead to hard to debug errors.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/docker/package_managers/install_pkgs.bzl
+++ b/docker/package_managers/install_pkgs.bzl
@@ -41,7 +41,8 @@ users will write something like:
 
 def _generate_install_commands(tar, installation_cleanup_commands):
     return """
-tar -xvf {tar}
+tar -xf {tar}
+export DEBIAN_FRONTEND=noninteractive
 dpkg -i --force-depends ./*.deb
 dpkg --configure -a
 apt-get install -f
--- a/docker/package_managers/run_install.sh.tpl
+++ b/docker/package_managers/run_install.sh.tpl
@@ -38,21 +38,22 @@ trap "rm -rf $tmpdir" EXIT
 mkdir -p $(dirname $tmpdir/%{installables_tar})
 cp -L $(pwd)/%{installables_tar} $tmpdir/%{installables_tar}
 cp -L $(pwd)/%{installer_script} $tmpdir/installer.sh
-# Temporarily create a container so we can mount the named volume
-# and copy files.  It's okay if /bin/true doesn't exist inside the
-# image; we are never going to run the image anyways.
-vid=$($DOCKER $DOCKER_FLAGS volume create)
-cid=$($DOCKER $DOCKER_FLAGS create -v $vid:/tmp/pkginstall $image_id /bin/true)
-for f in $tmpdir/*; do
-    $DOCKER $DOCKER_FLAGS cp $f $cid:/tmp/pkginstall
-done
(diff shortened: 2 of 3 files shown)


docker/package_managers/install_pkgs.bzl
```diff
@@ -41,7 +41,8 @@ users will write something like:
 
 def _generate_install_commands(tar, installation_cleanup_commands):
     return """
-tar -xvf {tar}
+tar -xf {tar}
+export DEBIAN_FRONTEND=noninteractive
```
gravypod, line 45: Would it be difficult to attempt to add a test that validates this is set? This seems like an important addition. I'm not sure how things worked before this.

docker/package_managers/run_install.sh.tpl
```diff
@@ -38,21 +38,22 @@ trap "rm -rf $tmpdir" EXIT
...
-
-$DOCKER $DOCKER_FLAGS attach $cid || true
-
-reset_cmd $image_id $cid %{output_image_name}
-$DOCKER $DOCKER_FLAGS save %{output_image_name} > %{output_file_name}
-$DOCKER $DOCKER_FLAGS rm $cid
-$DOCKER $DOCKER_FLAGS volume rm $vid
+
+# Subshells are used to have multiple traps.
+(
```
gravypod, line 43: I never knew about this. This is very interesting.

gravypod (changes requested):

docker/package_managers/install_pkgs.bzl
kaoet: There is no difference in the built image before and after the change. The only difference is the log message of "bazel build".

If `DEBIAN_FRONTEND` is not set, `dpkg` will try multiple frontends and eventually fallback to `Teletype`:

```
debconf: unable to initialize frontend: Dialog
debconf: (TERM is not set, so the dialog frontend is not usable.)
debconf: falling back to frontend: Readline
debconf: unable to initialize frontend: Readline
debconf: (Can't locate Term/ReadLine.pm in @INC (you may need to install the Term::ReadLine module) (@INC contains: /etc/perl /usr/local/lib/x86_64-linux-gnu/perl/5.28.1 /usr/local/share/perl/5.28.1 /usr/lib/x86_64-linux-gnu/perl5/5.28 /usr/share/perl5 /usr/lib/x86_64-linux-gnu/perl/5.28 /usr/share/perl/5.28 /usr/local/lib/site_perl /usr/lib/x86_64-linux-gnu/perl-base) at /usr/share/perl5/Debconf/FrontEnd/Readline.pm line 7.)
debconf: falling back to frontend: Teletype
```

After we explicitly set `DEBIAN_FRONTEND`, `dpkg` won't print those logs.

PS: You can see these logs by running
```shell
docker run --rm debian:buster-slim bash -c 'apt-get update && apt-get install -y locales'
```

PR #1899, opened 2021-06-20
WIP: Precompile binaries into repo

gravypod:
When loading rules_docker we have a sort of bootstrapping problem. Bazel
executes WORKSPACE commands (repository_rule) before it executes things
that we define in our BUILD files (rule). Things defined in WORKSPACE
are the only things that can do non-hermetic things in Bazel. This includes
downloading things from external storage. This leads to a bootstrapping
issue:

  1. You want to download http://example.com/some.exe
  2. You implement `curl` as a `cc_binary`
  3. You need to execute `curl` in your WORKSPACE file but it's isn't built...

To solve this problem Bazel has a built in http downloader which
understands how to fetch single files from well-behaved HTTP servers.
This works for most things but rules_docker has an edge case. We need to
download docker/oci images from remote registries. This functionality is
implemented in the go-containerregistry project and turned into a CLI tool
in our //container/go/puller/... and //container/go/loader/... implementations.
Since we need to support a protocol not standardized by Bazel we need to hack
around this. This commit does the following things:

1. Make a script that compiles puller & loader for each GOOS/GOARCH we support
2. Adds a go_test that checks to make sure the puller & loader match what was
   compiled on your system. This only checks the GOOS/GOARCH of *your* system
   but CI will catch it for all systems as we run runners for each GOOS/GOARCH
   that we support.

This is better in a few ways:

1. There is currently no way to verify that the binaries for puller/loader have
   not been tampered with in the gcs bucket they are currently stored.
2. We know when a change we makes necessitates updating the puller/loader. We
   will get a test failure automatically in the same PR.
3. We no longer need to run a gcs bucket for rules_docker.

This is worse in the following ways:

1. The .zip file we distribute gets bigger.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


gravypod:
I'm closing this PR since committing the binaries into the repo for each arch is not a good long term solution. Each arch is ~16MB of binary data and to support all of the platforms we want to work on we're talking >100MB just for binaries you're not going to be using.

During the maintainers meeting we discussed two possible solutions:

1. Precompile binaries on each release and attach them to Github releases as artifacts.
2. Compile the golang binary using `rules_go` during the execution of repository rules. This will require we have a dep on `rules_go` but will allow us to support any host arch.

PR #1917, opened 2021-07-23
Setting visibility of image_bzl to public so that it can be used when…

### LINKED ISSUE #1916: Can't use stardoc to document custom rules that depend on java_image
# 🐞 bug report

### Affected Rule

This concerns this load statement:

```
load("@io_bazel_rules_docker//java:image.bzl", "java_image")
```

### Is this a regression?

Unknown

### Description

We're using [stardoc](https://github.com/bazelbuild/stardoc) in order to document our custom rules. One of our rules uses this load statement:

```
load("@io_bazel_rules_docker//java:image.bzl", "java_image")
```

The normal way to solve this is to use a `bzl_library`, but the `bzl_library` defined in [image_bzl](https://github.com/bazelbuild/rules_docker/blob/master/java/BUILD#L36) is marked a private.

In contrast, some other rule we have uses [lang:image](https://github.com/bazelbuild/rules_docker/blob/master/lang/BUILD#L20) which is not marked as private and then stardoc works.

This should be easy to fix by making `image_bzl` public.

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
    Mac OS Catalina
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
4.1.0
  </code>
</pre>

**Rules_docker version:**

<pre>
  <code>
0.17.0
  </code>
</pre>

ngeor:
… documenting custom rules with stardoc (Fixes #1916)

## PR Checklist

## PR Type

- [ x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ x] No


ngeor:
@googlebot I signed it!

ngeor:
Good morning folks, can I please get a review? :pray: @alexeagle  @gravypod @pcj @smukherj1

alexeagle:
I'm surprised if the issue is only in Java - should we change this everywhere?
Note that you can just pass --nocheck_visibility to Bazel to bypass the checking when you build your stardoc target

ngeor:
Right, I only use Java so that's why I reported it only for Java :) From some colleague who is more experienced with Bazel, he mentioned `rules_docker` might be a bit inconsistent regarding visibility. Let me see if I can use the flag you mentioned.

ngeor:
I gave it a try, with a mix of sometimes referencing `bzl` files and sometimes `bazel_library` targets, I got it to work. But, I can't actually use the `--nocheck_visibility` flag because that means every single build command we run against our repo (e.g. `bazel build //...`) won't work anymore unless we specify that flag.

ngeor:
Ok, my colleague advised me to use patches. I am applying patches on top of `rules_docker`. I should've done this before I created the MR, apologies for this. This revealed two more issues:

- another file we need to be changed is `repositories/BUILD`. As you predicted, more files might be needed to be public for other people, but for me just these two suffice. I opted in my patch to just make a library with a wildcard:

```

bzl_library(
    name = "repositories",
    srcs = glob(["*.bzl"]),
)
```

Furthermore, I see I need to patch another repo, Bazel Gazelle, because you use it and they don't even provide a `bazel_library`. I patched that as well and then I was able to use stardoc in a more sane way.

ngeor:
Related PR for gazelle https://github.com/bazelbuild/bazel-gazelle/pull/1092

ngeor:
Thank you for merging @alexeagle , I took the latest git sha from rules_docker and gazelle and I was able to use stardoc to auto-generate the documentation for our Bazel rules/macros!

PR #1925, opened 2021-08-30
Introduce bazel_integration_test and examples as nested WORKSPACE

alexeagle:


alexeagle:
Not surprisingly, it fails the first time it needs to run a Go program

```
ERROR: /Users/alex.eagle/Projects/rules_docker/examples/basic/BUILD:18:16: every rule of type container_image_ implicitly depends upon the target '@io_bazel_rules_docker//container/go/cmd/digester:digester', but this target could not be found because of: no such package '@io_bazel_rules_docker//container/go/cmd/digester': BUILD file not found in directory 'container/go/cmd/digester' of external repository @io_bazel_rules_docker. Add a BUILD file to a directory to mark it as a package.
ERROR: Analysis of target '//:basic_alpine' failed; build aborted: Analysis failed
```

gravypod (commented):
Big fan of this. Gives us room to start managing the complexity.

alexeagle:
@jonjohnsonjr since most of this go/gazelle machinery is borrowed from go-containerregistry, I wonder if you have any opinions on whether we should
- continue to make users also install those deps
- publish pre-built go binaries for every platform * every utility in here (up to ten: https://github.com/bazelbuild/rules_docker/tree/master/container/go/cmd )
- use viper to publish a single Go program with multiple commands, so that we only publish one binary for every platform

jonjohnsonjr:
> publish a single Go program with multiple commands, so that we only publish one binary for every platform

This seems like it would have a few nice properties...

1. Having 1 binary instead of 10 would reduce the overall size of these dependencies, given that go binaries are pretty heavy.
2. Possibly consolidate some common configuration or parsing code.
3. Distribution is a lot simpler, so other projects could maybe interop with rules_docker by using this binary?

Do you consider these binaries to have a stable API? Or is this just an implementation detail of the bazel rules? I could imagine pushing most of these tools into an external dependency like [`crane`](https://github.com/google/go-containerregistry/tree/main/cmd/crane) if rules_docker can use a different internal representation for images.

alexeagle:
ooh @jonjohnsonjr Crane looks like just the thing! I could imagine all the go utilities in this repo are included in the scope of what crane can do, or else they should be.

So there's an interesting, if difficult, path forward where we
- Add a crane toolchain to rules_docker that fetches the binary for all platforms
- Ensure crane has feature parity with the various tools here
- remove tools one at a time, replacing with spawning crane actions
- eventually remove the Go dependency altogether

I can't commit to doing all that myself, but it would be interesting to do some exploratory work.

jonjohnsonjr:
> I could imagine all the go utilities in this repo are included in the scope of what crane can do, or else they should be.

I'd say yes -- anything interesting that you might want to do to an image would reasonably fit within the ethos of `crane`.

> I can't commit to doing all that myself, but it would be interesting to do some exploratory work.

I'm in the same boat -- the only real blocker I see is the [intermediate form](https://github.com/bazelbuild/rules_docker/blob/master/intermediate-form.md) for images, but it _should_ be straightforward to migrate from that to the standard [OCI Image Layout](https://github.com/opencontainers/image-spec/blob/main/image-layout.md).

alexeagle:
this should be replaced with bazel-contrib/rules_bazel_integration_test anyway

PR #1932, opened 2021-09-27
Feature/fix docker commit windows

tomasz-cybulski-oqton:
Fix Windows docker builds.


docker/util/run.bzl
```diff
@@ -479,4 +479,4 @@ commit_layer = struct(
 
 def _process_commands(command_list):
     # Use the $ to allow escape characters in string
-    return 'sh -c $\"{0}\"'.format(" && ".join(command_list))
```
tomasz-cybulski-oqton, line 482: @Irvenae Do you have suggestion how to use the correct shell based on the host operating system?

PR #1960, opened 2021-11-17
feat: add timeout attribute to container_pull

alexeagle:
The PULLER_TIMEOUT env variable is convenient, but one very large image shouldn't require that all of them have timeouts changed.


ptarjan:
Amazing! Thank you. Is there precident of which should win when both an attribute and environment variable are both specified?

alexeagle:
I don't know of a common pattern for this, it seems like the rule attribute should be more specific, but I could also see an argument that the environment is quicker to override than a code change so you'd want the environment variable to take precedence.

ptarjan:
Yeah I argued with myself that both ways had good reasons so was hoping
there was a precident. Alas.

gravypod (approved):
LGTM

Issue #1599, opened 2020-08-26
Add Support for Multi-Arch and Multi-Platform Images


jeremyje:
It's currently not possible to use `docker manifest` and `docker buildx` to build multi-arch and multi-platform images. This is necessary if you're targeting multiple architectures (amd64, arm64, etc.) as well as multiple operating systems, (linux, windows_2004, windows_1909, etc). Windows here is interesting in that there's [strict host compatibility requirements](https://docs.microsoft.com/en-us/virtualization/windowscontainers/deploy-containers/version-compatibility?tabs=windows-server-2004%2Cwindows-10-2004).

https://www.docker.com/blog/multi-arch-build-and-images-the-simple-way/

excavador:
Hello!

Any progress? I really want to see bazel way to build multi-arch and multi-platform images. It would simplify so much our kubernetes configuration

excavador:
Any updates?

excavador:
Any updates?

jonjohnsonjr:
Somewhat blocked on https://github.com/bazelbuild/bazel/issues/6519

Some discussion in https://github.com/bazelbuild/rules_docker/issues/300

I've got ideas in https://github.com/bazelbuild/rules_docker/issues/1695#issuecomment-748254384 but need some guidance re: the rules UX.

wenerme:
keep is open

jwillker:
I have this necessity today.

xuzhenglun:
any update?

lbcjbb:
The pull request to support fat manifest is ready since 3 months >>> https://github.com/bazelbuild/rules_docker/pull/2087

mehtamanan:
+1

farcop:
+1

loeffel-io:
+++++

friendly-pineapple:
+999

dlshle:
while (true) +

loeffel-io:
> while (true) +

😂

avbhandaru:
👀

ofey404:
+1

PR #1963, opened 2021-11-21
Transition container image target platform

uhthomas:
Transition to the target platform associated with the `architecture` and
`operating_system` attributes on container image rules.

This change allows for container image rules to build the correct binary for
the target platform, regardless of the host platform. Container image rules
would require the use of the `target_compatible_with` attribute to prevent
mismatching host and target platforms building dependencies incorrectly.
Additionally, hosts which did not match the target platform had to explicitly
specify the target platform with the `--platforms` command-line option.

This change fixes the aforementioned issues and
https://github.com/bazelbuild/rules_docker/issues/690.

Massive thank you to @joneshf for the initial source. It has been adapted to
automatically select the target platform associated with the container image,
as opposed to always using `@io_bazel_rules_go//go/toolchain:linux_amd64`.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_docker/issues/690

`container_image` rules do not transition the target platform.

## What is the new behavior?

`container_image` rules transition the target platform associated with the `architecture` and `operating_system` attributes.

- [ ] Yes
- [x] No


uhthomas:
I noticed that unlike [`jar_dep_layer`](https://github.com/bazelbuild/rules_docker/blob/8f6a2aa6e906f518603848a59917d46c972ce99e/java/image.bzl#L161); [`jar_app_layer`](https://github.com/bazelbuild/rules_docker/blob/8f6a2aa6e906f518603848a59917d46c972ce99e/java/image.bzl#L224), [`_war_app_layer`](https://github.com/bazelbuild/rules_docker/blob/8f6a2aa6e906f518603848a59917d46c972ce99e/java/image.bzl#L383) and [`_war_dep_layer`](https://github.com/bazelbuild/rules_docker/blob/8f6a2aa6e906f518603848a59917d46c972ce99e/java/image.bzl#L346) use [`_container.image`](https://github.com/bazelbuild/rules_docker/blob/8f6a2aa6e906f518603848a59917d46c972ce99e/container/container.bzl#L36-L38) instead of [`lang_image`](https://github.com/bazelbuild/rules_docker/blob/8f6a2aa6e906f518603848a59917d46c972ce99e/lang/image.bzl#L242). Is this on purpose? This difference means that at current, `java_image` rules will not inherit this change.

alexeagle:
@smukherj1 do you have any answer to that question about why Java is different in this regard?
@uhthomas are there scenarios where it matters? .jar files are supposedly "portable" - I guess JNI that depends on a C++ library (SWIG?) would require the transition?

uhthomas:
@alexeagle Thanks for taking a look! I'm not too familiar with the java image rules, java rules or cross-compilation of java. I'm not sure what effect changing the target platform would have, if any. A cursory Google shows that Java may sometimes depend on C as you suggested. As such, even if the JARs are portable, the C would still need to be built for the correct target platform.

lang/image.bzl
```diff
@@ -34,8 +34,8 @@ def _binary_name(ctx):
     # /app/foo/bar/baz/blah
     return "/".join([
         ctx.attr.directory,
-        ctx.attr.binary.label.package,
-        ctx.attr.binary.label.name,
+        ctx.attr.binary[0].label.package,
```
alexeagle, line 37: for this to be safe, it seems like the `binary` attribute would have to be declared with `allow_single_file=true`? how do you know the zeroth element should be used?

lang/image.bzl
```diff
@@ -250,7 +271,7 @@ image = struct(
         # the runfiles dir.
         "binary": attr.label(
             executable = True,
-            cfg = "target",
+            cfg = image_transition,
```
alexeagle, line 274: what would users do if they're using `container_image` rather than one of the convenience lang-specific rules? do they need to make their own transition, or just refer to this one?

alexeagle (commented):
This seems fantastic. Does it mean users no longer need to specify a `--platform` on the command line when building an image? I think some documentation update may be required too.

Maintainers here don't know much about transitions, we'll try to get someone from the bazel team to help verify.

lang/image.bzl
uhthomas: I also feel it's an odd design decision, but it is documented behaviour when using an [outgoing edge transition](https://docs.bazel.build/versions/main/skylark/config.html#accessing-attributes-with-transitions).

> When attaching a transition to an outgoing edge (regardless of whether the transition is a 1:1 or 1:2+ transition) ctx.attr is forced to be a list if it isn’t already. The order of elements in this list is unspecified.

It looks like we aren't the only ones who looked twice as per this comment in [rules_go](https://github.com/bazelbuild/rules_go/blob/aaf48cbd40dbc74a6199a1ff18db72090b11952b/go/private/rules/transition.bzl#L266).

I would imagine it's for [1:2+ transitions](https://docs.bazel.build/versions/main/skylark/config.html#defining-12-transitions):

> Outgoing edge transitions can map a single input configuration to two or more output configurations.

In such scenarios, there may be multiple binaries. However for our 1:1 transition, it will only ever have a single item.

lang/image.bzl
uhthomas: The `binary` and `dep` attributes are not defined on the `container_image` rule and my initial thoughts were that it might not be necessary to include this anywhere else.

I thought about it for a bit and I think you raise a good point. It might be necessary to include the transition for all label attributes to ensure everything is built for the correct target.

A quick look at which attributes may need to have an attached transition:
- [container_layer.files](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/container/layer.bzl#L342)
- [container_layer.tars](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/container/layer.bzl#L369)
- [container_image.base](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/container/image.bzl#L585)
- [container_image.label_files](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/container/image.bzl#L665)
- [container_image.launcher](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/container/image.bzl#L686)
- [container_image.layers](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/container/image.bzl#L700)
- [lang/image.base](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/lang/image.bzl#L245)
- [lang/image.binary](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/lang/image.bzl#L251) (done)
- [lang/image.data](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/lang/image.bzl#L258)
- [lang/image.dep](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/lang/image.bzl#L266) (done)


Another thought: I wonder if transitions affect `select`? If so, would this mean basically every attribute would need to be transitioned? Assuming this is the case, it might make sense to use an [incoming edge transition](https://docs.bazel.build/versions/main/skylark/config.html#incoming-edge-transitions) instead.

I'll test that incoming edge transition idea and get back to you.

uhthomas: Just had a quick play with incoming edge transitions and I don't think they're what we're looking for.

uhthomas: Incoming edge transitions seem to work as expected. I can update this PR to use them instead as that will then transition the entire target, rather than just individual attributes. Attributes with `cfg = "host"` will be respected.

uhthomas:
Thanks @alexeagle! You're absolutely correct, explicitly targeting platforms with the `--platforms` flag is no longer required. 

I think you're right that some documentation updates may be in order. I'll take a look tomorrow. 😄

lang/image.bzl
uhthomas: Assuming we're going to use an incoming edge transition instead, this may no longer be relevant.

lang/image.bzl
uhthomas: Done.

All rules which inherit [container/image](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/container/image.bzl#L802) now transition to the correct target platform. Dependencies of these rules also implicitly transition (i.e [container_layer](https://github.com/bazelbuild/rules_docker/blob/61e2df80a0f92c2c9e96c17390b1ecc64af65e40/docs/container.md#container_layer)).

ahmedkirasystems:
Hey @uhthomas I tried consuming rules from your repo.
```

http_archive(
    sha256 = "62b79a11bae076a8acd97749a06e5b8448c793ec5a730aea925e68c6f46dcaff",
    strip_prefix = "rules_docker-73964b225c38052bbfb1ea556bfb49d2bd2edc77",
    urls = ["https://github.com/uhthomas/rules_docker/archive/73964b225c38052bbfb1ea556bfb49d2bd2edc77.tar.gz"],
)
```

I'm using the same command which works when using `0.21.0` version.
```
bazel run //:app
```

I get error related to https://github.com/bazelbuild/bazel/issues/10134

```
INFO: Invocation ID: db25a6c0-c7e1-4a87-8071-d0548bc83ac1
ERROR: While resolving toolchains for target @io_bazel_rules_go//:cgo_context_data: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
ERROR: Analysis of target '//:app' failed; build aborted: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
INFO: Elapsed time: 1.225s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (72 packages loaded, 549 targets configured)
FAILED: Build did NOT complete successfully (72 packages loaded, 549 targets configured)
    Fetching @bazel_gazelle_go_repository_config; Restarting.
```

I'm using `container_image`, which from the conversations above, I get the feeling that it should work.

uhthomas:
Thank you for trying out these changes @ahmedkirasystems. You're completely right and I feel silly for not testing this on a more minimal repo which doesn't have additional CC toolchains.

I cannot reproduce this with a minimal `container_image`:

```starlark
container_image(name = "some_image")
```

but I can reproduce this with `go_image`:

```starlark
go_image(
    name = "some_image",
    embed = [":some_lib"],
)
```

`go_image` targets typically have their `goarch` and `goos` attributes configured which [implicitly disable cgo](https://github.com/bazelbuild/rules_go/blob/aaf48cbd40dbc74a6199a1ff18db72090b11952b/go/private/rules/transition.bzl#L146). This does not happen when using the `image` rule transition.

This behaviour can be observed with [rules_docker v0.21.0](https://github.com/bazelbuild/rules_docker/releases/tag/v0.21.0) and the following `go_image` targets. On a MacOS host, the `pure = "off"` build fails with the same error you observed:

```starlark
# fail
go_image(
    name = "some_image",
    embed = [":some_lib"],
    goarch = "amd64",
    goos = "linux",
    pure = "off",
)

# ok
go_image(
    name = "some_image",
    embed = [":some_lib"],
    goarch = "amd64",
    goos = "linux",
    pure = "on",
)

# ok 
go_image(
    name = "some_image",
    embed = [":some_lib"],
    goarch = "amd64",
    goos = "linux",
    pure = "auto",
)

# ok (implicit `pure = "auto"`)
go_image(
    name = "some_image",
    embed = [":some_lib"],
    goarch = "amd64",
    goos = "linux",
)
```

This seems like some weird behaviour on `rules_go`'s part honestly, but does mostly fall inline with the behaviour of the native Go toolchain. I think the only exception is that the native Go toolchain does not implicitly disable cgo when `goarch` and `goos` are the same as the host.

Side note: I spent a little bit trying to understand why both `pure = "auto"` and `pure = "on"` built okay, but the latter would fail to execute with `standard_init_linux.go:228: exec user process caused: no such file or directory`. Turns out I had `--@io_bazel_rules_go//go/config:linkmode=pie` in my `.bazelrc`. Not sure why this broke things, but thought it was worth noting.

I was going to suggest that this could be fixed by setting `pure = "on"` but `rules_go` still seems to want to use C? Explicitly configuring `goarch` and `goos` again properly disables C. I imagine this is because the `pure` attribute is a little bit misleading, such cgo is actually controlled [elsewhere](https://github.com/bazelbuild/rules_go/blob/aaf48cbd40dbc74a6199a1ff18db72090b11952b/go/toolchain/toolchains.bzl#L60-L72) (and [internal_cgo_off](https://github.com/bazelbuild/rules_go/blob/4cd45a2ac59bd00ba54d23ebbdb7e5e2aed69007/go/platform/list.bzl#L80)) via platforms.

So I think the correct fix is either transition `@io_bazel_rules_go//go/toolchain:cgo_constraint` directly, or use the native Go platform if the downstream rule is `go_image`.

---

Sorry, I understand this comment is quite long.

**TL;DR:** The `go_image` rule needs to replicate the original behaviour of force disabling cgo by default. I'll think about the best way to do this, any input would be really appreciated.

uhthomas:
I wonder if there's some sort of subtle bug somewhere in `rules_go`, or if the documentation is wrong. Per https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#go_binary:

> If auto, pure mode is enabled when no C/C++ toolchain is configured or when cross-compiling.

So, I'm not sure why pure mode is not enabled when there is no C/C++ toolchain.

alexeagle:
rules_docker is hard to maintain due to all the languages being supported in this repo! I don't know the answer for this Go transition question either. Does it make sense to cut scope of this PR to cover everything except Go? Or else maybe we need another maintainer like @pcj (or a rules_go maintainer like @achew22 or @steeve ) to advise why the auto-selection of `pure` mode isn't working under transitions.

uhthomas:
I was trialing some changes which would configure cgo on behalf of `rules_go`, but I think that's wrong and so I'm going to revert it. I think there's a bug in `rules_go` somewhere as `pure` has absolutely no effect at all when using transitions instead of `goarch` and `goos`. This behaviour can be observed by targeting a custom `linux/amd64` platform from MacOS:

```sh
bazel build --platforms=//:linux_x86_64 //:some_bin
```

```starlark
platform(
    name = "linux_x86_64",
    constraint_values = [
        "@platforms//cpu:x86_64",
        "@platforms//os:linux",
    ],
)

# fail
go_binary(
    name = "some_bin",
    embed = [":some_lib"],
)

# fail
go_binary(
    name = "some_bin",
    embed = [":some_lib"],
    pure = "off",
)

# fail
go_binary(
    name = "some_bin",
    embed = [":some_lib"],
    pure = "on",
)

# ok
go_binary(
    name = "some_bin",
    embed = [":some_lib"],
    goarch = "amd64",
    goos = "linux",
)

# ok
go_binary(
    name = "some_bin",
    embed = [":some_lib"],
    goarch = "amd64",
    goos = "linux",
    pure = "on",
)
```

joneshf:
> Does it make sense to cut scope of this PR to cover everything except Go?

I can't speak for anyone except myself. And I'm not really part of this PR (aside from being referenced for having coalesced someone else's work into a related issue comment) so take this comment for what it's worth. 

To me, having some way to build go images without having to specify the `--platforms` flag is the appealing part of this PR. In that regard, I'd be more than willing to use one of the workarounds @uhthomas posted above. I don't actually get the whole `pure` stuff myself, but as long as some incantation can build an image from macOS, I'd be happy.

I also can respect and understand that any forward progress is good progress, and that the people that might be able to help diagnose the issues might not be able to in a short amount of time (for whatever reasons).

I guess what I'm saying (as someone with no real weight behind their vote) is that cutting scope for this PR makes sense to me. Again, take it for what it's worth.

alexeagle (approved):
looks great, I can't wait to try it.

Let's try again for rules_go in a followup PR?
I think we also need a followup for docs change, or can you make a blurb to put in the next release notes about what changes users are expected to make (to their code or their behavior)?

alexeagle:
@achew22 this seems like a sad enough missing feature for go + docker that you might want to poke someone on rules_go who is familiar with this `pure` auto-selection.

pcj:
@uhthomas Can you add documentation on how this is intended to be used for end-users?  Under what conditions could this break existing users?

gravypod:
@alexeagle since we've tagged the last release is this good to be merged in?

uhthomas:
Hi all, apologies for the lack of activity the past couple of weeks. I've been thinking a lot about the right way to deal with `go_image` and how these changes may affect existing users. At current, the following may not behave as expected:

```starlark
go_image(
    name = "some_image",
    embed = [":go_default_library"],
)
```

> It may be important to note that without the changes in this PR, the `go_image` binary would be built for the host platform and consequently produce an invalid container image.

The above target will build correctly when the host and target platforms match but, assuming additional toolchains haven't been configured, will fail to resolve a valid C toolchain where the host and target platforms do not match. The suggested workaround would be to set `pure = "on"`, but as [discussed earlier](https://github.com/bazelbuild/rules_docker/pull/1963#issuecomment-980475834) this does not work without also configuring `goarch` and `goos`. I imagine this is a bug in rules_go. It may be possible to configure `@io_bazel_rules_go//go/toolchain:cgo_constraint` directly depending on the value of the `pure` attribute, but this seems hacky. Given these changes it seems `goarch` and `goos` may actually be redundant and should maybe be deprecated or removed. I can't think of a use-case where it would be useful for a `go_image` binary to differ from the container image.

I suppose this shouldn't actually break existing users as I imagine 99% will either already be using `goarch` and `goos` or a `linux x86_64` host. It should not have been possible to build a `go_image` target any other way.

Given this conclusion I think it may actually make sense to proceed with these changes and iterate based on feedback.

What are your thoughts? Are there any concerns?

I'd like to write some documentation, should this be part of a follow-up PR?

Thanks for all the support!

alexeagle:
Yeah merged. We should release again now so we have few commits to bisect in case it broke

uhthomas:
Hah! I made the previous comment just as the PR was merged. Thank you.

alexeagle:
no worries @uhthomas can you send a snippet you'd like to include in the release notes, for what users should change as they update? Are they forced to make changes as they upgrade or can they do them after? and notes for go_image users

uhthomas:
Raised https://github.com/bazelbuild/rules_go/issues/3033

Yep, no worries @alexeagle. I don't believe users should have to make any changes. If there are problems, the [`target_compatible_with`](https://docs.bazel.build/versions/main/be/common-definitions.html#common.target_compatible_with) attribute can be used to exclude the target from non-linux x86_64 builds.

```skylark
...
target_compatible_with = [
    "@platforms//cpu:x86_64",
    "@platforms//os:linux",
],
...
```

Users should now expect all image targets to build correctly regardless of the host platform. This should also pave a path forward for https://github.com/bazelbuild/rules_docker/issues/1599.

As for `go_image` users, the `goarch` and `goos` attributes should be dropped if there is an appropriately configured cc toolchain and [--incompatible_enable_cc_toolchain_resolution](https://github.com/bazelbuild/bazel/issues/7260) is flipped.

alexeagle:
TypeScript users will need to update the `tsconfig.json` to add more `rootDirs` settings to account for this change, as the TypeScript compiler has to be told where to look for outputs of previous actions.

`./bazel-out/k8-fastbuild-ST-4a519fd6d3e4/bin` is the entry to add if `tsconfig.json` is in the project root, otherwise `../../../bazel-out/k8-fastbuild-ST-4a519fd6d3e4/bin/path/to/folder` where `path/to/folder` is the path in the workspace to the folder containing tsconfig.json and the number of leading `..` segments matches the segments in that path.

See https://github.com/bazelbuild/rules_nodejs/issues/3284

dragonsinth:
Wanted to draw attention to https://github.com/bazelbuild/rules_docker/issues/2009 -- I can't seem to build at all on MacOS from this commit on.  I've tried all of:

```
bazel build go/src/fs/services/amber/main/amber:image_push
bazel build --incompatible_enable_cc_toolchain_resolution go/src/fs/services/amber/main/amber:image_push
bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 go/src/fs/services/amber/main/amber:image_push
bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 --incompatible_enable_cc_toolchain_resolution go/src/fs/services/amber/main/amber:image_push 
```

Note that the second formulation is what we've been using (in 0.22.0).

In all cases the error is the same:
```
ERROR: While resolving toolchains for target @io_bazel_rules_go//:cgo_context_data: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
ERROR: Analysis of target '//go/src/fs/services/amber/main/amber:image_push' failed; build aborted: No matching toolchains found for types @bazel_tools//tools/cpp:toolchain_type. Maybe --incompatible_use_cc_configure_from_rules_cc has been flipped and there is no default C++ toolchain added in the WORKSPACE file? See https://github.com/bazelbuild/bazel/issues/10134 for details and migration instructions.
```

I'm a little unclear on whether this is a bug, whether I need to be running a newer version of bazel itself, or if I need some other kind of update to how we're setup.

Thanks!
Scott

alexeagle:
@uhthomas any suggestion for that issue?

uhthomas:
Morning!

Apologies for the weirdness this change has caused. It seems mostly due to rules requiring a C toolchain despite not actually using it.

I really hope that https://github.com/bazelbuild/bazel/issues/14726 will help with this.

Until then, do you think it would be sensible to put this behind an attribute or command line flag?

@dragonsint Would you be able to provide a minimal reproduction? I don't see anything obvious, but can take a closer look this evening 😄

dragonsinth:
@uhthomas check out this repo on master and run literally anything on mac.  For example;

```
~/github/go/src/github.com/bazelbuild/rules_docker bazel test -k --test_env=HOME --define=ENV_KEY=my_key --define=ENV_VALUE=my_value --test_output=errors --verbose_failures //tests/container:all
DEBUG: /private/var/tmp/_bazel_scottb/c8f88c60da9e66091021ee89ff8fcae9/external/bazel_toolchains/rules/rbe_repo/version_check.bzl:68:14: 
Current running Bazel is ahead of bazel-toolchains repo. Please update your pin to bazel-toolchains repo in your WORKSPACE file.
DEBUG: /private/var/tmp/_bazel_scottb/c8f88c60da9e66091021ee89ff8fcae9/external/bazel_toolchains/rules/rbe_repo/checked_in.bzl:125:14: buildkite_config not using checked in configs; Bazel version 5.0.0 was picked/selected but no checked in config was found in map {"0.20.0": ["8.0.0"], "0.21.0": ["8.0.0"], "0.22.0": ["8.0.0", "9.0.0"], "0.23.0": ["8.0.0", "9.0.0"], "0.23.1": ["8.0.0", "9.0.0"], "0.23.2": ["9.0.0"], "0.24.0": ["9.0.0"], "0.24.1": ["9.0.0"], "0.25.0": ["9.0.0"], "0.25.1": ["9.0.0"], "0.25.2": ["9.0.0"], "0.26.0": ["9.0.0"], "0.26.1": ["9.0.0"], "0.27.0": ["9.0.0"], "0.27.1": ["9.0.0"], "0.28.0": ["9.0.0"], "0.28.1": ["9.0.0"], "0.29.0": ["9.0.0"], "0.29.1": ["9.0.0", "10.0.0"], "1.0.0": ["9.0.0", "10.0.0"], "1.0.1": ["10.0.0"], "1.1.0": ["10.0.0"], "1.2.0": ["10.0.0"], "1.2.1": ["10.0.0"], "2.0.0": ["10.0.0"], "2.1.0": ["10.0.0"], "2.1.1": ["10.0.0", "11.0.0"], "2.2.0": ["11.0.0"], "3.0.0": ["11.0.0"], "3.1.0": ["11.0.0"], "3.2.0": ["11.0.0"], "3.3.0": ["11.0.0"], "3.3.1": ["11.0.0"], "3.4.1": ["11.0.0"], "3.5.0": ["11.0.0"], "3.5.1": ["11.0.0"], "3.6.0": ["11.0.0"], "3.7.0": ["11.0.0"], "3.7.1": ["11.0.0"], "3.7.2": ["11.0.0"], "4.0.0": ["11.0.0"]}
ERROR: /Users/scottb/github/go/src/github.com/bazelbuild/rules_docker/tests/container/BUILD:817:16: While resolving toolchains for target //tests/container:alpine_arch_ppc64le: no matching toolchains found for types //toolchains/docker:toolchain_type
ERROR: /Users/scottb/github/go/src/github.com/bazelbuild/rules_docker/tests/container/BUILD:724:16: While resolving toolchains for target //tests/container:alpine_linux_amd64_image_oci: no matching toolchains found for types //toolchains/docker:toolchain_type
```

And many many more like this.

uhthomas:
I see.

Thank you for helping me reproduce the issue. I don't think this can be resolved until optional toolchains are implemented.

It looks like @katre is working hard on this, so let's hope we see this land in Bazel soon!

For the time being, @alexeagle, do you think it would be sensible to hide this transition behind a config flag (or rule attribute)? This allows users who have the correct CC toolchains to use transitions, but doesn't break existing users who may not have said CC toolchains.

Again, apologies for the inconvenience. It's a really unfortunate situation and I hope that one day everything will "just work" as it should 😄 We're getting there!

alexeagle:
@uhthomas I agree we'll need to add some way to opt-in or opt-out of the transitions feature. Making it default to off is breaking to users who already successfully upgraded past 0.23 and removed their `--platforms` flags on their build, so I think we'd make the flag an opt-out instead with transitions . It has the downside that every user making the upgrade has to discover the breakage and then find that the opt-out exists, so I also see an argument that we should make it opt-in with the intent to flip the default later. I think a custom flag is the way to go, since a typical repo may have many rules_docker callsites that would need to be aware of the value. Sth like `bazel build --@io_bazel_rules_docker//transitions:enable=false //my:container_target` ?

Even if optional toolchains landed in Bazel now, we'd need that cherry-picked to a 5.x release, and then there would still be lots of users on an older LTS release. So I think we're going to be in this situation for a while.

Our other choice is just to revert the change. We originally discussed cutting a rules_docker release containing only this change so that it would be easier to back out if we found problems. It's a bit harder since 0.23 ended up including more, but it's possible.

I'll leave it up to @gravypod how to do the release mechanics.

uri-canva:
https://github.com/bazelbuild/proposals/blob/main/designs/2022-01-21-optional-toolchains.md#abstract mentions there are dummy toolchains people are using to work around the lack of optional toolchains. Is that something that rules_docker can use?

uri-canva:
> This allows users who have the correct CC toolchains to use transitions

What are these correct CC toolchains? Should we set up cross compilation toolchains for host -> target transitions? Is there documentation we can refer to?

uhthomas:
> https://github.com/bazelbuild/proposals/blob/main/designs/2022-01-21-optional-toolchains.md#abstract mentions there are dummy toolchains people are using to work around the lack of optional toolchains. Is that something that rules_docker can use?

I also read this, but wasn't able to find much info. A [cursory search](https://cs.github.com/bazelbuild/bazel?q=dummy+CC) of `bazelbuild/bazel` for `dummy` and `CC` doesn't seem to show much either.

> > This allows users who have the correct CC toolchains to use transitions
> 
> What are these correct CC toolchains? Should we set up cross compilation toolchains for host -> target transitions? Is there documentation we can refer to?

Zig introduced a [CC cross-compiler](https://andrewkelley.me/post/zig-cc-powerful-drop-in-replacement-gcc-clang.html) relatively recently. @motiejus has written some Bazel rules, [bazel-zig-cc](https://sr.ht/~motiejus/bazel-zig-cc/), which have been working quite well the past few months.

This shouldn't be required unless cross compiling C, though.

katre:
This is my fault: I forgot history and was unclear.

In Bazel, there is no dummy CC toolchain (anymore). There is, however, a [dummy Android SDK](https://cs.opensource.google/bazel/bazel/+/master:tools/android/android_sdk_repository_template.bzl;drc=d9fffbae9be80a9396e1e283eaacf02319a81f80;l=323) that shows the same idea: a guaranteed-to-fail toolchain is created and registered, hopefully with the lowest possible priority, in order to allow for better error messaging if no actual toolchain is found. This is also a use for optional toolchains and I am looking forward to ripping it out.

(I'll go ahead and say that there is a dummy CC toolchain in Google's monorepo, and it's a huge PITA, and also needs to be ripped out).

Please let me know if you have further questions about toolchains, although for specifics of CC cross-compilation I'll have to refer you to other experts.

dragonsinth:
Can someone TLDR a workaround if there is one?  What does one need to do, specifically, to use `0.23.0` on Mac?  Is there some command line, .bazelrc, or WORKSPACE workaround?

uhthomas:
Although not ideal, `rules_docker` can be patched to disable transitions.

**//third_party:io_bazel_rules_docker.patch**

```diff
diff -urN a/container/image.bzl b/container/image.bzl
--- a/container/image.bzl       1969-12-31 19:00:00.000000000 -0500
+++ b/container/image.bzl       2000-01-01 00:00:00.000000000 -0000
@@ -795,7 +795,7 @@
     attrs = _attrs,
     outputs = _outputs,
     implementation = _impl,
-    cfg = _image_transition,
+    # cfg = _image_transition,
 )

 container_image_ = rule(
```

**WORKSPACE**

```starlark
http_archive(
    sha256 = "85ffff62a4c22a74dbd98d05da6cf40f497344b3dbf1e1ab0a37ab2a1a6ca014",
    strip_prefix = "rules_docker-0.23.0",
    urls = ["https://github.com/bazelbuild/rules_docker/releases/download/v0.23.0/rules_docker-v0.23.0.tar.gz"],
    patch_args = ["-p1"],
    patches = ["//third_party:io_bazel_rules_docker.patch"],
)
```

kush-work:
@uhthomas I tried the above workaround but faced this error
```
ERROR: An error occurred during the fetch of repository 'io_bazel_rules_docker':
	File "/private/var/tmp/_bazel_kush/ad19ce41cd089d3b231bf82dcc7ef7f6/external/bazel_tools/tools/build_defs/repo/http.bzl", line 121, column 10, in _http_archive_impl
		patch(ctx)
	File "/private/var/tmp/_bazel_kush/ad19ce41cd089d3b231bf82dcc7ef7f6/external/bazel_tools/tools/build_defs/repo/utils.bzl", line 135, column 22, in patch
		ctx.patch(patchfile, strip)
Error in patch: Unable to load package for //third_party:io_bazel_rules_docker.patch: BUILD file not found in any of the following directories. Add a BUILD file to a directory to mark it as a package.
```

Any help here? (Apologies if it's a noob question)

uhthomas:
Make sure there is a `BUILD.bazel` file in the `third_party` directory.

```sh
touch third_party/BUILD.bazel
```

It should look something like this:

```
third_party/
├── BUILD.bazel
└── io_bazel_rules_docker.patch
BUILD.bazel
WORKSPACE
```

mtsgrd:
I tried the workaround as well, but face a different error.. Any ideas what I'm doing wrong?

```
➜  monorepo git:(master) ✗ bazel version
Build label: 5.0.0-homebrew
➜  monorepo git:(master) ✗ bazel run //api:deployment
...
INFO: Repository 'io_bazel_rules_docker' used the following cache hits instead of downloading the corresponding file.
 * Hash '85ffff62a4c22a74dbd98d05da6cf40f497344b3dbf1e1ab0a37ab2a1a6ca014' for https://github.com/bazelbuild/rules_docker/releases/download/v0.23.0/rules_docker-v0.23.0.tar.gz
If the definition of 'io_bazel_rules_docker' was updated, verify that the hashes were also updated.
ERROR: An error occurred during the fetch of repository 'io_bazel_rules_docker':
	File "/private/var/tmp/_bazel_mattias/8b2d0da21301e8d11b52e94111a879cf/external/bazel_tools/tools/build_defs/repo/http.bzl", line 122, column 10, in _http_archive_impl
		patch(ctx, auth = auth)
	File "/private/var/tmp/_bazel_mattias/8b2d0da21301e8d11b52e94111a879cf/external/bazel_tools/tools/build_defs/repo/utils.bzl", line 167, column 22, in patch
		ctx.patch(patchfile, strip)
Error in patch: Error applying patch /Users/mattias/c/monorepo/third_party/io_bazel_rules_docker.patch: Cannot find file to patch (near line 2), old file name (a/container/image.bzl       1969-12-31 19:00:00.000000000 -0500) doesn't exist, new file name (b/container/image.bzl       2000-01-01 00:00:00.000000000 -0000) doesn't exist.
```

uhthomas:
@mtsgrd I imagine your first error may have been caused by not including `patch_args = ["-p1"]`?

Thanks for taking a proper look at the patch, it makes sense. Again, apologies for the inconvenience. I'll spend some time shortly to see if there's anything which can be done to make this better.

mtsgrd:
@uhthomas no I tried both -p1 and -p0, but no luck. How did you generate yours? I see the format is different.

purkhusid:
This also breaks rules_docker for anybody that is using custom platforms. E.g. we use `rules_dotnet` and the rules use platforms to select which .Net version to use so we have a custom platform like this:

```
platform(
    name = "custom_platform",
    constraint_values = [
        "@io_bazel_rules_docker//platforms:run_in_container",
        "@io_bazel_rules_dotnet//dotnet/toolchain:5.0.201",
    ],
    parents = [
        "@local_config_platform//:host",
    ],
)
```

After upgrading rules_docker we get:
```
While resolving toolchains for target //src/dotnet/fsharp_bootstrap/src/api:bin.exe: no matching toolchains found for types @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core
```
And if I add `--toolchain_resolution_debug=@io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core`to the command we get the following:
```
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:darwin_amd64_5.0.201_darwin_amd64_fsharp_toolchain-impl-fsharp; mismatching values: osx, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:darwin_aarch64_5.0.201_darwin_amd64_fsharp_toolchain-impl-fsharp; mismatching values: arm64, osx, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:linux_amd64_5.0.201_darwin_amd64_fsharp_toolchain-impl-fsharp; mismatching values: 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:windows_amd64_5.0.201_darwin_amd64_fsharp_toolchain-impl-fsharp; mismatching values: windows, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:darwin_amd64_5.0.201_darwin_aarch64_fsharp_toolchain-impl-fsharp; mismatching values: osx, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:darwin_aarch64_5.0.201_darwin_aarch64_fsharp_toolchain-impl-fsharp; mismatching values: arm64, osx, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:linux_amd64_5.0.201_darwin_aarch64_fsharp_toolchain-impl-fsharp; mismatching values: 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:windows_amd64_5.0.201_darwin_aarch64_fsharp_toolchain-impl-fsharp; mismatching values: windows, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:darwin_amd64_5.0.201_linux_amd64_fsharp_toolchain-impl-fsharp; mismatching values: osx, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:darwin_aarch64_5.0.201_linux_amd64_fsharp_toolchain-impl-fsharp; mismatching values: arm64, osx, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:linux_amd64_5.0.201_linux_amd64_fsharp_toolchain-impl-fsharp; mismatching values: 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:windows_amd64_5.0.201_linux_amd64_fsharp_toolchain-impl-fsharp; mismatching values: windows, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:darwin_amd64_5.0.201_windows_amd64_fsharp_toolchain-impl-fsharp; mismatching values: osx, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:darwin_aarch64_5.0.201_windows_amd64_fsharp_toolchain-impl-fsharp; mismatching values: arm64, osx, 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:linux_amd64_5.0.201_windows_amd64_fsharp_toolchain-impl-fsharp; mismatching values: 5.0.201
(13:20:38) INFO: ToolchainResolution:     Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: Rejected toolchain @io_bazel_rules_dotnet//dotnet/toolchain:windows_amd64_5.0.201_windows_amd64_fsharp_toolchain-impl-fsharp; mismatching values: windows, 5.0.201
(13:20:38) INFO: ToolchainResolution:   Type @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core: target platform @io_bazel_rules_docker//platforms:image_transition: No toolchains found.
(13:20:38) ERROR: /workspace/monorepo/src/dotnet/fsharp_bootstrap/src/api/BUILD.bazel:3:14: While resolving toolchains for target //src/dotnet/fsharp_bootstrap/src/api:bin.exe: no matching toolchains found for types @io_bazel_rules_dotnet//dotnet:toolchain_type_fsharp_core
(13:20:38) ERROR: Analysis of target '//src/dotnet/fsharp_bootstrap/deploy/api:image' failed; build aborted: 
```
It seems like this change breaks our setup because when building the image we transition to the platform that rules_docker uses for the transition but that one is missing the `@io_bazel_rules_dotnet//dotnet/toolchain:5.0.201` constraint

sashamor:
This PR breaks skaffold, which expects the output image to be in `bazel info bazel-bin`. We use skaffold locally for our go microservices and cannot update due to this patch.
See: https://github.com/GoogleContainerTools/skaffold/issues/7228

Is there a way to force the built images to be put into the `k8-fastbuild` directory so skaffold can find them?

uhthomas:
This sounds like a bug in skaffold. I don't believe any tools should rely on Bazel to output build artefacts to the same directory all the time. Instead, `bazel query` should be used.

I remember seeing a similar item for Gazelle or rules_go but can't find it for the life of me.

uri-canva:
> Instead, `bazel query` should be used.

What's the verb / option to output build artefacts paths?

uhthomas:
I was going to sugest `bazel query --output=location` but reading further into https://github.com/bazelbuild/bazel/issues/8739 it looks like it may be less trivial than that.

`bazel cquery` seems to be the cleanest solution https://github.com/bazelbuild/bazel/issues/8739#issuecomment-924663891.

lazcamus:
Is it possible to update the release notes for `0.23` to include "this release is not expected to work on Mac OSX" under the "Breaking Changes" section @gravypod ?

fmeum:
@uhthomas In the (probably very common) case of a Linux x86_64 image built as part of a build with target platform Linux x86_64, the transition to a synthetic platform will change the configuration and thus cause all dependencies of the image to be rebuilt. Do you think that it would be feasible to move the configuration change from an incoming to an outgoing dependency edge, perhaps by wrapping the image rules in a macro that applies a forwarding rule? This would make it possible to disable the transition if CPU and OS are already at the expected values.

uhthomas:
> @uhthomas what about the config option to disable transitions? [#1963 (comment)](https://github.com/bazelbuild/rules_docker/pull/1963#issuecomment-1035029299) `--@io_bazel_rules_docker//transitions:enable=false`

Sounds like a reasonable suggestion. I can cut a PR for that soon.

uhthomas:
> @uhthomas In the (probably very common) case of a Linux x86_64 image built as part of a build with target platform Linux x86_64, the transition to a synthetic platform will change the configuration and thus cause all dependencies of the image to be rebuilt. Do you think that it would be feasible to move the configuration change from an incoming to an outgoing dependency edge, perhaps by wrapping the image rules in a macro that applies a forwarding rule? This would make it possible to make the transition a no-op if CPU and OS are already at the expected values.

That's an interesting idea, I like it. Would you be able to help me understand how the transition would become no-op by switching to an outgoing dependency edge?

fmeum:
@uhthomas I've thought more about it and now see two ways to prevent rebuilds:

1. Let users specify a `string_dict` mapping of CPU/OS pairs to the label of the platform the user would like to use for that pair. Then, instead of using the synthetic platform, set `platforms` to the singleton lost containing that platform in the transition. This should result in the transition being a no-op when the platform doesn't change.

2. Apply the transition on the outgoing edges of every image rule. This allows the transition to access the values of attributes specified via select expressions, which currently isn't supported for transitions on incoming edges (https://github.com/bazelbuild/bazel/issues/15157).
Then add common `string` attributes `current_os` and current_cpu` to these rules and populate them in a wrapper macro using select on `@platforms//os:{cpu,os}` to translate the `constraint_value`s into string. In the transition, return an empty dict if the image's os and cpu match the current ones read from the attrs.

Happy to provide more details and think about further alternatives.

uri-canva:
The next LTS release is still half a year away isn't it?

thefallentree:
the suggestion ```--@io_bazel_rules_docker//transitions:enable=false ``` doesn't really work , bazel complains  ```ERROR: @io_bazel_rules_docker//transitions:enable :: Error loading option @io_bazel_rules_docker//transitions:enable: no such package '@io_bazel_rules_docker//transitions': BUILD file not found in directory 'transitions' of external repository @io_bazel_rules_docker. Add a BUILD file to a directory to mark it as a package.
```

**Comment by uhthomas** [comment]
@thefallentree Whst version of rules_docker are you using? (including sha256)

**Comment by thefallentree** [comment]
> @thefallentree which version of rules_docker are you using? (including sha256)

I've been using this.
```
     http_archive(
         sha256 = "85ffff62a4c22a74dbd98d05da6cf40f497344b3dbf1e1ab0a37ab2a1a6ca014",
         strip_prefix = "rules_docker-0.23.0",
         urls = ["https://github.com/bazelbuild/rules_docker/releases/download/v0.23.0/rules_docker-v0.23.0.tar.gz"],
     )
```
and just to clarify , 0.23.0 doesn't work for me because of  java_image rule is complaining about unresolved c++ toolchain, so I'm not sure how that's possible, do you have any more information about it?

thefallentree:
For anyone landed here: 

Now with  ```rules_docker-0.25.0``` new release

you can now use ```--@io_bazel_rules_docker//transitions:enable=false```

and it should work!

cfife-btig:
I don't know what other people's use cases are - it sounds like a lot of people wanted a disable transitions flag because of an error with go images in macOS.

For us, we are having some inconveniences because as I believe @fmeum has said, the build configurations for images are different which is causing us to have to reproduce all of the build steps anytime we build an image, even if I had already built all the prior steps.

Since we don't have any special build targets, I went ahead and tried using the new disable transitions flag, but I don't think this works at least for my use case. Images still have a different build configuration and makes me run my pipeline from scratch. Maybe optional toolchaining or something like what @fmeum suggested would be better. Or maybe this is an issue with the disable transitions flag.

In the meantime, my team has not felt like the latest releases have caused more slowness then help, so we've just been staying at 0.22 rules_docker since we don't have any special build targets.

PR #1978, opened 2021-12-10
fix: correct link to toolchain documentation

mgred:
## PR Type

The link on _How to use the Docker Toolchain_ in `docs/container.md` neither works in Github nor in the online documentation

## What is the new behavior?

The link now works and points to the right page and anchor.

- [ ] Yes
- [x] No


container/image.bzl
```diff
@@ -946,7 +946,7 @@ def container_image(**kwargs):
             layers inside the Docker registry.
 
     This rule references the `@io_bazel_rules_docker//toolchains/docker:toolchain_type`.
-    See [How to use the Docker Toolchain](toolchains/docker/readme.md#how-to-use-the-docker-toolchain) for details.
+    See [How to use the Docker Toolchain](https://github.com/bazelbuild/rules_docker/tree/master/toolchains/docker/readme.md#how-to-use-the-docker-toolchain) for details.
```
gravypod, line 949: This would break previewing at a specific PR. Could we change do a similar change except replace `https://github.com/bazelbuild/rules_docker/tree/master/` with `/`? That will allow GitHub to rewrite the urls for a specific branch/pr/commit. This is very useful for looking into the past for docs.

gravypod (changes requested):

container/image.bzl
mgred: @gravypod thanks for your review and suggestions! I made the change and tested it, perfectly works. This will still not work for the documentation published under `docs.aspect.dev. Is this something we should have in mind?

mgred: Sorry, I fixed and rebased already :upside_down_face:

gravypod (approved):

PR #2004, opened 2022-01-16
Add registry_file attribute to the container_pull rule

jrbeverly:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The rule `container_push` supports passing the repository and tag by either Label (`repository_file, tag_file`) or String (`repository, tag`). The attribute `registry` only supports String as an option.

Both the `registry` and `repository` are required attributes.

## What is the new behavior?

This change adds the attribute `registry_file` to the `container_push` rule. This allows all of the destination parameters (`registry, repository, tag`) to be supplied as either Label or String. The registry attribute pair follows the behaviour of `repository`, by requiring the String value to always be provided, even if overridden by the `repository_file`.

```python
container_push(
    name = "push_image",
    image = ":application_image",
    registry = "",
    registry_file = ":test.registry",
    repository = "docker/test",
    tag = "latest",
)
```

- [ ] Yes
- [x] No

## Other information

Passing the `registry` as a Label is intended for cases where the list of possible registries is not a fixed set. This is common in AWS (Amazon Web Services) where the registry for each account is based on the AWS Account ID and region. Publishing in this kind of infrastructure can mean that production/release artifacts are located within one "Artifacts" account, while developer sandbox artifacts can be spread across multiple sandbox accounts. Allowing for the `registry` to be a Label allows for the registry to be calculated based on the current working sandbox. 

The existing behaviour of requiring the String attribute to always be provided has been carried over from `repository`. When using `repository_file`, you must specify the `repository` value, as it is a mandatory field. This is typically done by providing an empty string, and then specifying the `repository_file`. This could be altered, but would see a breaking change with the macro interface for `container_push` no longer requiring the specification of `registry` and `repository`.

```python
container_push(
    name = "push_image",
    registry = "",
    registry_file = ":test.registry",
    repository = "docker/test",
    tags = ["manual"],
    # ...
)
```


container/push.bzl
```diff
@@ -47,6 +47,12 @@ def _impl(ctx):
 
     # Parse and get destination registry to be pushed to
     registry = ctx.expand_make_variables("registry", ctx.attr.registry, {})
+
+    # If a registry file is provided, override <registry> with file value
+    if ctx.file.registry_file:
```
gravypod, line 52: To make it clearer that this is overriding registry could we do something like:

```
if not ctx.file.registry_file:
  registry = ctx.expand_make_variables("registry", ctx.attr.registry, {})
else:
   registry = "$(cat {})"....
```

I wasn't immediately able to see that this was overriding this field (I had assumed `pusher_input.append` was adding an argument or something).

tests/container/BUILD
```diff
@@ -334,6 +334,44 @@ container_push(
     tags = ["manual"],
 )
 
+genrule(
```
gravypod, line 337: Is this the way you intend to use this in your repository (via `genrule`)? Could we add an example of how this might actually be used as a test? That would make providing support for this feature easier as we could link to the test for a demo of the approximate use case.

gravypod (changes requested):
I'm not completely sold on this feature. I know there is substantial prior art here (`tag_file`, etc) but I think this points to an unclean design of `rules_docker` which couples *building* with *uploading* artifacts in a strange way. I think this is something that could be investigated further (if only we had the time :cry:). 

I'm not opposed to this but if we could add a better example of how someone might use this I'd be really happy (either in discussion here or via tests).

container/push.bzl
jrbeverly: I'm not sure on shifting this one, as the pattern that the file has follows the "assignment, then if override" ([repository_file](https://github.com/bazelbuild/rules_docker/blob/74be3091ffc0d2fc4d4050d9524519e848d0868c/container/push.bzl#L53), [tag_file](https://github.com/bazelbuild/rules_docker/blob/74be3091ffc0d2fc4d4050d9524519e848d0868c/container/push.bzl#L60), [client_config](https://github.com/bazelbuild/rules_docker/blob/74be3091ffc0d2fc4d4050d9524519e848d0868c/container/push.bzl#L108)) pattern.

I am fine with changing it, but it would break with the pattern in the method.

tests/container/BUILD
jrbeverly: I think we would likely leverage a custom `rule` (or `repository_rule`) for creating the inputs for `registry_file/repository_file`, as I'd expect those to be conditional on 1) Where are we working in? (sandbox/staging/etc), and 2) Which entity/user does this repository belong to? (e.g. `myuser/mycontainer`). 

The `genrule` is a stand-in for the command execution that would be responsible for determining the registry based on those conditions (1 & 2).

jrbeverly (commented):
I think your review comment summarizes the issue we are encountering, which is that to use `container_push` in our multi-registry AWS ECR model, we are needing to figure out ways to encode those registries into the build process, when I suspect we'd want to provide these values at "runtime" (e.g. specifying `--dst` in a `bazel run`)

I'll provide a more descriptive use-case/context below:

In AWS (Amazon Web Services), containers can be pushed to ECR (Elastic Container Registry) on a per-account, per-region basis. What this looks like is that the registry for given account region looks like so `123412341234.dkr.ecr.us-east-1.amazonaws.com` (`{account}.dkr.ecr.{region}.amazonaws.com`). While working on projects, developers will sometimes wish to publish the images to a repository for use in an account region they are working in. As these repositories do not always exist, a `bazel run //:publish.create` exists that can create the repository for them on-demand. The published container can then be used in the account region for validation, testing, perftest, scanning, etc.

When finished, a pull request (changeset) can be proposed, and those artifacts will be published to a staging repository (for tests or provisioning a staging version of the service). When reviewed & merged, those artifacts will be published to a production repository, that will mirror the artifacts to multiple regions (`us-east-1`, `ca-central-1`, `eu-central-1`, etc)

As artifacts aren't always published in the exact same region as deployment (e.g. `eu-west-3` might use repositories located in `eu-central-1`), it isn't a perfect 1 to 1 mapping. The decision isn't so much a "don't publish in this region", as a "Is it needed to publish in this region, or is an existing region sufficient?"

The rule responsible for building the "repository create script" would source the registry & repository name from a common `publish.registry`. This is the same rule that would provide the `registry_file` & `repository_file` to the `container_push` rule. An outline of what it kind of would look like is as such (but not the same implementation):

```python
# container_image(name = docker_app)

# Command responsible for retrieving the intended ECR (could be a repository_rule)
genrule(
    name = "publish.registry",
    srcs = [],
    outs = ["img.registry", "img.repository"],
    cmd = """awsecr-on-demand repository $@""",
    output_to_bindir = 1,
)

# Runnable command to create the ECR
awsecr_repository_create(
    name = "publish.create",
    registry_file = ":img.registry",
    repository_file = ":img.repository"
)

# Runnable command to publish image to the ECR
container_push(
    name = "publish.push",
    format = "Docker",
    image = ":docker_app",
    registry = "",
    registry_file = ":img.registry",
    repository = "",
    repository_file = ":img.repository"
    tag = "test", # likely to be versioned by stamp variables
    tags = ["manual"],
)
```

PR #2032, opened 2022-02-26
Specifying client config.json with Bazel label

linzhp:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

The `client_config` attribute of `docker_toolchain_configure` only accepts an absolute path to a directory. This is not portable, because absolute paths are machine dependent.

Issue Number: N/A

## What is the new behavior?
Allow the `client_config` attribute of `docker_toolchain_configure` to accept a Bazel label to the `config.json` file, so we can check the file into a repository and check out the repository anywhere.

- [ ] Yes
- [x] No

## Other information

Ideally, we should declare `docker_client_config` attribute of `container_pull` as a label instead of a string, so the changes in the content of client config will invalidate the cache and trigger a rebuild. However, this will be a backward-incompatible change, as absolute paths will no longer work. What do you think?


gravypod:
This looks good. Is there any specific use case you're attempting to satisfy here @linzhp?

gravypod (approved):

linzhp:
Yes, we want to configure the `credHelpers` in `config.json`, so we can access our internal docker registry with a custom authentication, but we don't want require all dev and CI machines to check out our repository in exactly the same location. So we hope to have something like:

```python
    docker_toolchain_configure(
        name = "docker_config",
        client_config = "@//third_party/docker/registry/config:config.json",
    )
```

gravypod:
Thanks @linzhp. Just wanted to get a "reason" on the issue so that when someone Googles "how do I X rules_docker" they find this page :)

PR #2041, opened 2022-03-14
toolchain_configure converts Bazel label of config.json to absolute path

linzhp:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

Issue Number: #2040

## What is the new behavior?
`toolchain_configure` will build a `docker_toolchain` target with an absolute path to the config.json even given a Bazel label, so `container_push` will always get an absolute path. We cannot pass the Bazel label to `container_push` because it is a regular Bazel rule, which executes in Bazel sandbox and cannot get an absolute path given a label. The `container_pull` rule is able to do that because it is a repository rule.

Revert the doc changes for `docker_toolchain` in #2032 because it's `client_config` cannot take a Bazel label.

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

Issue #2040, opened 2022-03-14
Allowing to specify Bazel label for docker config.json in container_push


linzhp:
# 🚀 feature request

### Relevant Rules

* `container_push`

### Description

Currently, if people wants to use a custom config.json for `container_push`, they have to specify the absolute path to the directory, which is machine-dependent. #2032 allows `container_pull` to use a Bazel label to config.json. We need to make `container_push` to have that too.

### Describe the solution you'd like

Technically, it's wrong for any Bazel rule to take an absolute path as input. Since rules_docker is not v1.0 yet, I wish we could introduce a breaking change to make all `client_config` and `docker_client_config` attributes to be Bazel label only.

### Describe alternatives you've considered

We can continue to accept both absolute paths and Bazel labels.

linzhp:
@gravypod @alexeagle Can you review #2041 and #2042, and advice which is the right way to go?

PR #2042, opened 2022-03-15
Changing docker client config attributes to Bazel labels

linzhp:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #2040

## What is the new behavior?

`toolchain_configure` only takes a Bazel label

- [x] Yes
- [ ] No

Existing applications will have to change the `client_config` of their `toolchain_configure` rule to be a Bazel label.

## Other information
This is more correct than #2032 and #2041, but it introduces a breaking change.


toolchains/docker/toolchain.bzl
```diff
@@ -68,8 +68,8 @@ docker_toolchain = rule(
         ),
         "client_config": attr.string(
             default = "",
-            doc = "A custom directory or a Bazel label for the docker client config.json. " +
-                  "If DOCKER_CONFIG is not specified, the value of the " +
+            doc = "A custom directory for the docker client config.json. If " +
```
uhthomas, line 71: Would you agree the documentation for this attribute reads strangely? I don't really understand what it's trying to say. Might be good to update this?

uhthomas (commented):
The PR and logic itself seems sound to me, I don't see any use-cases where an absolute path would be desirable. Despite this, it should still possible for the toolchain to use the client config specified in the home directory, or with the `DOCKER_CONFIG` environment variable.|

Pending response to the comments I left, I'm happy to approve this.

toolchains/docker/toolchain.bzl
linzhp: How about now?

uhthomas: Much better, though the second to last sentence is missing an "If" at the beginning I believe.

uhthomas (approved):
Thank you for making those changes! LGTM

PR #2043, opened 2022-03-16
perf: remove no-remote-cache execution requirements

alexeagle:
These were introduced to reduce load on a remote-cache instance to avoid network saturation.
A month later, a feature was added in one remote-cache implementatation which provides a different fix:
https://github.com/buchgr/bazel-remote/issues/440 rejects large input files on upload.

In practice, while these action do often produce huge outputs, they are also slow to re-execute.
In many cases it's worth it to use a remote-cache for RunAndCommitLayer in particular to avoid a local rebuild
even though it's a large network fetch.
Currently users can't configure this because we've hardcoded the values. If they do want to keep the
no-remote-cache execution requirement, they can do this via a tag (provided they opt-in to
experimental_allow_tags_propagation, see https://github.com/bazelbuild/bazel/issues/8830)

https://github.com/bazelbuild/rules_docker/pull/1856#issuecomment-1060240919 is an example of a user
asking for these to be removed.

### DIFF
--- a/container/layer.bzl
+++ b/container/layer.bzl
@@ -166,11 +166,6 @@ def build_layer(
         input_manifests = xz_input_manifests,
         tools = files + file_map.values() + tars + debs + [manifest_file] + xz_tools,
         outputs = [layer],
-        execution_requirements = {
-            # This action produces large output files, so it's not
-            # economical to send this to the remote-cache
-            "no-remote-cache": "1",
-        },
         use_default_shell_env = True,
         mnemonic = "ImageLayer",
     )
--- a/container/layer_tools.bzl
+++ b/container/layer_tools.bzl
@@ -187,12 +187,6 @@ def assemble(
         arguments = [args],
         tools = inputs,
         outputs = [output],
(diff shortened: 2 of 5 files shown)


gravypod (approved):

sluongng:
yes, using `--modify_execution_info` you can override the execution requirement per mnemonic. So you should be able to  add/remove the `no-remote-cache` for each mnemonic that was modified in this PR.

alexeagle:
I'll begin by pointing out that rules_docker has some actions which are doing The Wrong Thing by reading very large inputs and writing very large outputs. rules_oci is a better alternative because we only run actions that only ever operate on a single layer at a time, no "flattening" operations which cause this problem. For example, there should never be an action that takes a base layer as an input, it should just have its digest shuttled around between metadata files.

@nkoroste the relevant mnemonics for rules_docker actions that often have huge inputs/outputs are:

```
GZIP
GUNZIP
RunAndCommitLayer
ImageLayer
JoinLayers
```

Issue #2046, opened 2022-03-19
Mac 12.3 removed python 2.7, resulting in a regression


sbussard:
# 🐞 bug report

### Affected Rule

rules_docker

### Is this a regression?

Yes, kind of (see below)

### Description

MacOS 12.3 removed python 2, so rules_docker doesn't work anymore.

https://www.macrumors.com/2022/01/28/apple-removing-python-2-in-macos-12-3/#:~:text=Apple%20will%20no%20longer%20bundle,security%20patches%2C%20or%20other%20changes.

## 🔬 Minimal Reproduction

WORKSPACE
```bazel
http_archive(
    name = "io_bazel_rules_docker",
    sha256 = "59536e6ae64359b716ba9c46c39183403b01eabfbd57578e84398b4829ca499a",
    strip_prefix = "rules_docker-0.22.0",
    urls = ["https://github.com/bazelbuild/rules_docker/releases/download/v0.22.0/rules_docker-v0.22.0.tar.gz"],
)

load(
    "@io_bazel_rules_docker//repositories:repositories.bzl",
    container_repositories = "repositories",
)

container_repositories()

load(
    "@io_bazel_rules_docker//java:image.bzl",
    _java_image_repos = "repositories",
)

_java_image_repos()
```

BUILD
```bazel
load("@io_bazel_rules_docker//java:image.bzl", "java_image")

java_image(
    name = "image",
    srcs = glob(["src/*.java"]),
    main_class = "App",
)
```

## 🔥 Exception or Error

<pre><code>
env: python: No such file or directory
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
MacOS 12.3
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
Bazelisk version: v1.11.0
Build label: 4.2.2
Build target: bazel-out/darwin-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Thu Dec 2 18:28:52 2021 (1638469732)
Build timestamp: 1638469732
Build timestamp as int: 1638469732
  </code>
</pre>

**Rules_docker version:**

<pre>
  <code>
0.22.0
  </code>
</pre>

**Anything else relevant?**

sbussard:
This fixes it for me

```zsh
brew install python3
brew link python@3.9
ln -s -f /usr/local/bin/python3.9 /usr/local/bin/python
```

[...]

PR #2051, opened 2022-03-29
Replace python with python3 in incremental_load.sh.tpl

dennajort:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

On MacOS 12.3, Apple removed the default `python` executable.
Building images doesn't work and throw this error:
`env: python: No such file or directory`

Issue Number: #2046 2046

## What is the new behavior?
Build on a system without `python` executable (as MacOS 12.3+) works as expected.

- [ ] Yes
- [x] No

### DIFF
--- a/container/incremental_load.sh.tpl
+++ b/container/incremental_load.sh.tpl
@@ -115,7 +115,7 @@ function find_diffbase() {
     NEW_DIFF_IDS+=("${diff_id}")
   done
 
-  TOTAL_DIFF_IDS=($(cat "${name}" | python -mjson.tool | \
+  TOTAL_DIFF_IDS=($(cat "${name}" | python3 -mjson.tool | \
       grep -E '^ +"sha256:' | cut -d'"' -f 2 | cut -d':' -f 2))
 
   LEGACY_COUNT=$((${#TOTAL_DIFF_IDS[@]} - ${#NEW_DIFF_IDS[@]}))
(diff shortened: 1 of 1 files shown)


linzhp (commented):
Can you sign CLA?

dennajort:
I think I did sign it @linzhp

linzhp (approved):

container/incremental_load.sh.tpl
```diff
@@ -115,7 +115,7 @@ function find_diffbase() {
     NEW_DIFF_IDS+=("${diff_id}")
   done
 
-  TOTAL_DIFF_IDS=($(cat "${name}" | python -mjson.tool | \
+  TOTAL_DIFF_IDS=($(cat "${name}" | python3 -mjson.tool | \
```
linzhp, line 118: For backward-compatibility of old systems that don't have python3, can you do this instead?

```suggestion

  export PYTHON="python"
  if command -v python3 &>/dev/null; then
      PYTHON="python3"
  fi

  TOTAL_DIFF_IDS=($(cat "${name}" | $PYTHON -mjson.tool | \
```

linzhp (changes requested):

container/incremental_load.sh.tpl
dennajort: Sounds good to me !
At first I was thinking if there is a way to pull this info directly from bazel, but I'm not very skilled in it and didn't find a way.

linzhp: rules_docker doesn't manage Python toolchain on its own and depends on system Python instead. So this is the only way.

linzhp (approved):

PR #2053, opened 2022-03-29
Run script allows empty docker_args

illicitonion:
Currently when using `bazel run` on a docker image, I get:
```
/[redacted].executable: line 281: docker_args[@]: unbound variable
```

Alternatively, we could just `set +u` before execing docker.

### DIFF
--- a/container/incremental_load.sh.tpl
+++ b/container/incremental_load.sh.tpl
@@ -277,6 +277,17 @@ if [[ "%{run}" == "True" ]]; then
   # We can clean up before we exec docker, since the exit handler will no longer run.
   cleanup
 
+  # Bash treats empty arrays as unset variables for the purposes of `set -u`, so we only
+  # conditionally add these arrays to our args.
+  args=(%{run_statement})
+  if [[ ${#docker_args[@]} -gt 0 ]]; then
+    args+=("${docker_args[@]}")
+  fi
+  args+=("%{run_tag}")
+  if [[ ${#container_args[@]} -gt 0 ]]; then
+    args+=("${container_args[@]}")
+  fi
+
   # This generated and injected by docker_*.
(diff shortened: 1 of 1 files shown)


container/incremental_load.sh.tpl
```diff
@@ -277,6 +277,18 @@ if [[ "%{run}" == "True" ]]; then
   # We can clean up before we exec docker, since the exit handler will no longer run.
   cleanup
 
+  # Bash treats empty arrays as unset variables for the purposes of `set -u` - work around this.
```
mindaugasrukas, line 280: nit: to be clear, `set -u` is not an issue but a kinda runtime linter. So this is not a workaround but a bug fix. Could you rephrase the comment to look not like a workaround?

container/incremental_load.sh.tpl
```diff
@@ -277,6 +277,18 @@ if [[ "%{run}" == "True" ]]; then
...
+  args=(%{run_statement})
+  if [[ ${#docker_args[@]} -gt 0 ]]; then
+    args+=("${docker_args[@]}")
+  fi
+  args+=("%{run_tag}")
+  if [[ ${#container_args[@]} -gt 0 ]]; then
+    args+=("${container_args[@]}")
+  else
+    args+=("-")
+  fi
```
mindaugasrukas, line 290: Please use the original form of `"${container_args[@]:-}"` or remove the `else` part. As if it's not set, it is empty but not `-`:
For more details, please see https://tldp.org/LDP/abs/html/parameter-substitution.html

mindaugasrukas (changes requested):

illicitonion:
Thanks for the review @mindaugasrukas - updated, PTAL!

mindaugasrukas (approved):
LGTM

linzhp (approved):

PR #2056, opened 2022-04-11
We have found some maintainers!

gravypod:
We have some maintainers.


uhthomas (approved):

alexeagle (approved):
Awesome, I'll update https://github.com/bazelbuild/rules_docker/discussions/2038 also

PR #2060, opened 2022-04-15
container_push: add doc about no transaction

sluongng:
When use container_push with skip_unchanged_digest, 
user often rely on the push process to (1) check for existence 
of the image in the registry and (2) push the image if (1) is false.

This leave a time window between (1) and (2) where the image
could have been pushed by an external process such as another CI
job running in parallel. In such situation, depends on the container
registry configuration, the push arrive later can either fail for attempting
to override an already pushed image, or it will override the previously
pushed image.

There is no transactional guarantee that can help coordinate (1) and (2) in
the current container registry spec. So let's document this edge case and
advise users to work around by using some external distributed lock for
running container_push target.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [ ] No


linzhp (approved):

linzhp:
Can you run buildifier to format the file?

linzhp (approved):

linzhp:
Can you run buildifier to pass the CI check?

sluongng:
CI now passed but it seems like we run into another issue caused by https://github.com/bazelbuild/bazel/issues/15558

A separate update to gpg key inside WORKSPACE is needed

Issue #1876, opened 2021-05-26
Keep package-installed files listing for Debian packages installed in a distroless image


pombredanne:
# 🚀 feature request

### Relevant Rules
When a package is installed, only metadata are kept and the list of installed files is lost/not saved with the package metadata.

I have a concern with what happens here: https://github.com/bazelbuild/rules_docker/blob/d18033b7eb3429a55dc4a579b5c19af57ab25e5f/container/build_tar.py#L224

### Description

In a distroless container image, the as-installed .deb packages are not saved with their files/md5sums file lists in what would be in `/var/lib/dpkg/info` on a regular Debian install. As a result, it is not possible to relate an installed package in a distroless image/layer to the set of files that were installed with this package.

This data can be important for software composition analysis and its security and license compliance tracking applications.

### Describe the solution you'd like

Each installed package should include some installed file listing possibly added in some per package file in the  `status.d/` directory. This is a Debian standard in `/var/lib/dpkg/info/<package name>`

This would make distroless images more readily introspectable, otherwise there is no intrisic way to relate a package (in status.d) to the set of its installed files.

@tejal29 you committed this originally with @dlorenc ... any insight to share there?

### Describe alternatives you've considered

I cannot fathom an in-container alternative to keep a tab of each packaged-installed file. Tracking outside would mean maintaining some external database which does not seem practical.

fedemengo:
@pombredanne I would love to help, do you already have something I start from?

pombredanne:
@fedemengo sorry for a late reply. I do not have anything done yet, but I would likely either:
- continue using a "non-like-debian" `/var/lib/dpkg/status.d/<package name>` and add `/var/lib/dpkg/status.d/<package name>.md5sums` (originally under `/var/lib/dpkg/info/<package name>.md5sums`) and/or create a `/var/lib/dpkg/status.d/<package name>.list` file for installed files  also originally under `/var/lib/dpkg/info/<package name>.list`
- OR just keep and copy over the `/var/lib/dpkg/info/` directory as is which will contain all the original parts of the packages.

pombredanne:
@fedemengo I do not think anything is "lost" .... but rather that in https://github.com/bazelbuild/rules_docker/blob/6ea707babdcd54514e0884278ac624fb8bda19c1/container/build_tar.py#L224
[...]

PR #2065, opened 2022-04-23
Add md5sums file list to distroless container

pombredanne:
This ensures that a "distroless" container layer tarball built from
Debian packages contains not only the control file of each package, but
also the md5sums file that lists original files included in a package.
The md5sums file is extracted from a .deb package and saved side-by-side
with the control under this path:
  var/lib/dpkg/status.d/<package-name>.md5sums

Reference: https://github.com/bazelbuild/rules_docker/issues/1876
Signed-off-by: Philippe Ombredanne <pombredanne@nexb.com>

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_docker/issues/1876

The current behaviour is to skip including the md5sum listing the list of package files when installing a Debian package in a distroless container image.  

## What is the new behavior?

The new behaviour is to included the md5sum listing the list of package files when installing a Debian package in a distroless container image, side-by-side with the control file.

- [ ] Yes
- [x] No

## Other information

I used only conservative Python syntax and kept the same code style as before.
There was no specific documentation entry, so I added a docstring


pombredanne:
For reference, I started this:
- [x] with https://github.com/GoogleContainerTools/distroless/issues/741
- [x] then here with https://github.com/bazelbuild/rules_docker/issues/1876

pombredanne:
@loosebazooka  gentle ping. You asked for a patch here in https://github.com/GoogleContainerTools/distroless/issues/741#issuecomment-918272365
> Oh sorry, yeah I mean I don't know why this form of metadata was chosen. Anyway, it seems like the correct place to inject the metadata is in rules_docker. Please provide a patch there.

Do you mind to have a look before it goes stale?
This is fairly straight forward and is tested.

pombredanne:
@smukherj1 @alexeagle @pcj @gravypod gentle ping too. Some kind of review (and approval) would be very nice!

container/build_tar.py
```diff
@@ -222,18 +223,34 @@ def write_temp_file(self, data, suffix='tar', mode='wb'):
...
       with tarfile.open(metadata_tar) as tar:
+        tar_members = tar.getmembers()
         # Metadata is expected to be in a file.
-        control_file_member = list(filter(lambda f: os.path.basename(f.name) == TarFile.PKG_METADATA_FILE, tar.getmembers()))
+        control_file_member = list(filter(lambda f: os.path.basename(f.name) == TarFile.PKG_METADATA_FILE, tar_members))
         if not control_file_member:
-           raise self.DebError(deb + ' does not Metadata File!')
+           raise self.DebError(deb + ' does not contain a control Metadata File!')
         control_file = tar.extractfile(control_file_member[0])
         metadata = b''.join(control_file.readlines())
```
aiuto, line 239: It feels like this could be more readably expressed as
```
metadata = controlfile.read().decode("utf-8")
pkg_name = TarFile.parse_pkg_name(metadata, deb)
```

aiuto (commented):
This looks good to me, but I don't maintain this repo.
Again, repeating the friendly ping.

container/build_tar.py
pombredanne: @aiuto good point. Done in the latest push. I also rebased. Thank you for the review!

pombredanne:
Joshua Katz asked on Slack:

> Might I ask: what is the use case here? What do these files do?


Here is the skinny:
- the background is explained in this original issue https://github.com/GoogleContainerTools/distroless/issues/741
- a debian package comes with an actual payload of code, data, docs and metadata. The metadata is essentially a control file as name/value pairs like the name and version of a package, its description, etc. And a list of the files installed in the payload tarball. This is an md5sums file with (path, md5) for each installed file.
- This listing of files paths help introspects an installed Debian-based OS/VM/Container to find which files were exactly installed from this package archive. Short of this you cannot guess this information.
- Debian packages install these files as part of the regular install. Tools that analyze Debian installation such as VMs or Docker images (like https://github.com/nexB/scancode.io/ that I co-maintain) rely on this to find which files come from which package.
- the bazel docker rules ignore this md5sums file when building distroless containers that are built from Debian packages.
- Therefore it is impossble or hard to introspect and observer which files come from which original Debian package in a distroless container built with these rules

This introspection is important to security and license compliance.
Without this, Distroless containers make this process brittle and complicated to automate and creates extra busywork for no good reason.

loosebazooka:
Are scanners going to be aware of the distroless specific custom directory structure: `/var/lib/dpkg/status.d`? Why not just write this information into the dpkg standard `.../info` (and keep whatever is currently in `../status.d` for continuity)?

pombredanne:
@loosebazooka re:

> Are scanners going to be aware of the distroless specific custom directory structure: /var/lib/dpkg/status.d? Why not just write this information into the dpkg standard .../info (and keep whatever is currently in ../status.d for continuity)?

All the open source scanners that can handle distroless images have long been aware of  the `/var/lib/dpkg/status.d` "oddity" and have long  been able to handle this. :innocent: :grin: 

This is true for scancode.io and scancode-toolkit but also all the other open source tools capable of scanning Distroless images that I know of including tern and anchore.

In scancode-toolkit (and scancode.io), this is handled there https://github.com/nexB/scancode-toolkit/blob/aba31126dcb3ab57f2b885090f7145f69b67351a/src/packagedcode/debian.py#L345 (this has recently been deeply refactored but this was there long before).

All these tools open source (or not) __will__ benefit directly from this small patch.

pombredanne:
@loosebazooka 

> Why not just write this information into the dpkg standard .../info (and keep whatever is currently in ../status.d for continuity)?

I am all for it.... I guess I have no idea why a /status.d was used in the first, except that it does allow independent installation without having to worry of merging control files in one?

Since this is already in odd territory that's no longer exactly Debian, I just kept on piling on this.

loosebazooka:
Sure, I don't have a strong preference either way, I was just curious. I was hoping to reduce the burden on new scanners, but I barely know what that would even mean.

LGTM from the distroless side.

gravypod:
This was reviewed by the SWEs on Distroless and it looks good. I'm going to approve this PR but we can't merge until the tests go green. Some of these tests are flakey so they might just need to be rerun.

gravypod (approved):
Lets get the tests pass then we can submit this. Thanks for putting this together!

pombredanne:
@gravypod the issue (or the fix) is that the published gpg checksum has just been updated https://github.com/bazelbuild/rules_docker/blob/900572d5185c84e846b66d90cff1ca5636a04cf2/WORKSPACE#L125 by @sluongng and seems to be because of a deper global issue of key expiration https://github.com/bazelbuild/bazel/issues/15558

I rebased and forced pushed and this should clean this up hopefully.

pombredanne:
@gravypod we are all green now :green_apple: :tada: 
@loosebazooka would you have an idea of how long this will take to propagate to actual public distroless images (as used in k8s) once merged?

PR #2035, opened 2022-03-01
Upgrade rules_go and Gazelle

linzhp:
## PR Checklist

Please check if your PR fulfills the following requirements:

- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [x] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?
rules_go and Gazelle are pretty old

Issue Number: #2036

## What is the new behavior?
* Upgrade rules_go to 0.30
* Upgrade Gazelle to 0.24
* Upgrade Bazel to 5.0, because rules_go requires Bazel 4.2.1

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

## Other information
New rules_go no longer allows to call `go_register_toolchains` multiple times, so we have to move it out of `go_deps`, because it will be called another time in `python/image.bzl`.

PR #2080, opened 2022-05-04
Update bazel-gazelle version

eine-doodka:
## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes
- [x ] Other... Please describe: Dependency update

Issue Number: N/A

- [ ] Yes
- [ x] No

## Other information

There is a long-running problem with go_repository from `Bazel-gazelle` throttling on `bazel sync` over clean project.
It's well-described [here](https://github.com/bazelbuild/bazel-gazelle/issues/1175) and fixed [here](https://github.com/bazelbuild/bazel-gazelle/pull/1206). There is actually a whole release (0.25.0) that includes these changes. I think we should merge them into the upstream.

UPD. Apparently it's not that simple. Looks like we have to update the version of Go itself here. Should we?

### DIFF
--- a/repositories/go_repositories.bzl
+++ b/repositories/go_repositories.bzl
@@ -37,7 +37,7 @@ def go_deps(go_repository_default_config = "@//:WORKSPACE"):
         go_repository_default_config (str, optional): A file used to determine the root of the workspace.
     """
     go_rules_dependencies()
-    go_register_toolchains()
+    go_register_toolchains(version = "1.18")
     gazelle_dependencies(go_repository_default_config = go_repository_default_config)
     excludes = native.existing_rules().keys()
     if "com_github_google_go_containerregistry" not in excludes:
--- a/repositories/repositories.bzl
+++ b/repositories/repositories.bzl
@@ -180,8 +180,8 @@ def repositories():
     if "bazel_gazelle" not in excludes:
         http_archive(
             name = "bazel_gazelle",
-            sha256 = "cdb02a887a7187ea4d5a27452311a75ed8637379a1287d8eeb952138ea485f7d",
-            urls = ["https://github.com/bazelbuild/bazel-gazelle/releases/download/v0.21.1/bazel-gazelle-v0.21.1.tar.gz"],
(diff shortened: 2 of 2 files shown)


eine-doodka:
Again, here is another problem. After update for the Go version has been made, we start failing because of multiple initialisation. If you override version at repositories/go_repositories.bzl:40, the build will fail, since the same line would be called at WORKSPACE:38 and after that - at WORKSPACE:232.
The whole problem burns down to "`rules_go` doesn't allow us to instantiate multiple instances of `go_sdk`,  even if they have the same version". Looks like an expected behaviour for me. Should we go to `rules_go` repo and fix it there?

linzhp:
Why do you need multiple instances of `go_sdk`?

eine-doodka:
@linzhp I don't, actually. But it looks like`rules_docker` instantiate them twice:
> If you override version at repositories/go_repositories.bzl:40, the build will fail, since the same line would be called at WORKSPACE:38 and after that - at WORKSPACE:232.
That only happens if you set the version manually to, say 1.18.
I'm trying to understand whether this double init should be that way.

repositories/go_repositories.bzl
```diff
@@ -37,7 +37,7 @@ def go_deps(go_repository_default_config = "@//:WORKSPACE"):
         go_repository_default_config (str, optional): A file used to determine the root of the workspace.
     """
     go_rules_dependencies()
-    go_register_toolchains()
+    go_register_toolchains(version = "1.18")
```
uhthomas, line 40: Oh, this is weird. I don't know what the behaviour of calling `go_register_toolchains` twice is, especially with an explicit version.

Maybe this should change? I'm afraid of how it may conflict with an existing SDK.

linzhp:
> If you override version at repositories/go_repositories.bzl:40, the build will fail, since the same line would be called at WORKSPACE:38 and after that - at WORKSPACE:232.

#2035 may fix that, but I didn't have time to fix all issues with that PR

Issue #1989, opened 2021-12-22
Add retry option to container_push rule


jumbosushi:
# 🚀 feature request

### Relevant Rules

[`container_push`](https://github.com/bazelbuild/rules_docker/blob/4c6d61c356aed293ea18c30d7dc50cfa609a3cf6/contrib/push-all.bzl#L96) in `contrib/push-all.bzl` 

### Description

It would be great if `container_push` rule supports a retry option. (this is an idea briefly explored in #673). We recently ran into a transient networking issue which was resolved by retrying manually:

```
# note: gcr info is deducted
2021/12/19 19:16:45 Unable to publish images: unable to push image gcr.io/xyz:abc: Head "https://gcr.io/v2/xyz/blobs/sha256:abc": read tcp 10.248.250.64:53422->142.250.148.82:443: read: connection reset by peer
```

### Describe the solution you'd like

- Add a new `retry` attribute to `container_push` rule like the following. 
```starlark
container_push = rule(
    attrs = {
        ...
        "retry": attr.bool(
            default = False,
            doc = "If true, retry push images on transient errors.",
        ),
    }
    ...
)
```
- Update `container/go/cmd/pusher/pusher.go` to support `--retry` boolean flag to enable retry logic

### Describe alternatives you've considered

We could add a retry logic to our `bazel run` invocations as a workaround, but we run it at various places within our CI system :/ 

---

I would be interested in working on the implementation! Let me know if the feature would be useful

uhthomas:
This would be great. Transient network failures when pushing dozens of images using `container_bundle` is really inconvenient.

PR #2101, opened 2022-05-30
Add retry-count to container_push

itmecho:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, if an image fails to push, it will immediately fail. While this makes sense for single image pushes, it's not ideal when pushing multiple containers as often this means running the entire build step again.

Issue Number: #1989 (related)

## What is the new behavior?
Adds a `retry_count` attribute to `container_push`. If the push fails, it will be retried until the `retry_count` is reached. On each retry, a log line will be printed to show the error that occurred as well as which attempt it was.

- [ ] Yes
- [x] No

## Other information
I've updated the docs to include the new attribute but I'm not sure if there are tests that cover it. I'm happy to add the tests but I might need some guidance as to where and how to implement them!


itmecho:
I've found the container_push tests in `tests/container/BUILD` but I'm not sure how to test it properly as it would require a push to fail one or more times and then succeed

container/go/cmd/pusher/pusher.go
```diff
@@ -126,8 +127,17 @@ func main() {
...
+		err := push(stamped, img)
+		if err == nil {
+			break
+		}
+
+		if *retryCount > 0 && retry < *retryCount {
+			log.Printf("Failed to push image on attempt %d: %v", retry+1, err)
+		} else {
+			log.Fatalf("Error pushing image to %s: %v", stamped, err)
+		}
```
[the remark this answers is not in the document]
itmecho: Thanks for taking a look! The reason I implemented it this way is that the actual issue we have is that sometimes it fails to load the auth credentials for GCR for 1 or 2 pushes out of > 30. This way it covers that scenario as well and means we don't have to rerun the entire build.

container/go/cmd/pusher/pusher.go
```diff
@@ -44,6 +44,7 @@ var (
 	format              = flag.String("format", "", "The format of the uploaded image (Docker or OCI).")
 	clientConfigDir     = flag.String("client-config-dir", "", "The path to the directory where the client configuration files are located. Overiddes the value from DOCKER_CONFIG.")
 	skipUnchangedDigest = flag.Bool("skip-unchanged-digest", false, "If set to true, will only push images where the digest has changed.")
+	retryCount          = flag.Int("retry-count", 0, "Amount of times the push will be retried.")
```
uhthomas, line 47: Should we validate this value? What happens when the retry count is `-1`? Some users may assume this to mean infinite retries, when in fact the container image would not even attempt to be pushed once.

container/go/cmd/pusher/pusher.go
```diff
@@ -126,8 +127,17 @@ func main() {
...
-	if err := push(stamped, img); err != nil {
-		log.Fatalf("Error pushing image to %s: %v", stamped, err)
+	for retry := 0; retry < *retryCount+1; retry++ {
+		err := push(stamped, img)
+		if err == nil {
+			break
+		}
+
+		if *retryCount > 0 && retry < *retryCount {
+			log.Printf("Failed to push image on attempt %d: %v", retry+1, err)
```
uhthomas, line 137: This error is missing the `stamped` variable. It will make it tricky to know what actually failed without this. Maybe reuse the original format?

```suggestion
			log.Printf("Error pushing image to %s (attempt %d): %v", stamped, retry, err)
```

container/go/cmd/pusher/pusher.go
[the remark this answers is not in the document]
uhthomas: I think this is reasonable. It would be cleaner to use @lbcjbb's suggestion, but I also empathise with flakiness beyond just pushing to container registries.

uhthomas (changes requested):

container/go/cmd/pusher/pusher.go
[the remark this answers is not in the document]
lbcjbb: After verification `WithRetry*` options are not available in `v0.5.1` of the library, so we can't use them for the moment.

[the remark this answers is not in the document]
itmecho: Yea, that definitely feels like a nicer solution if it was only network issues that caused the problems.

container/go/cmd/pusher/pusher.go
itmecho: Good point. Do you think it's better to log using `log.Fatalln` if it's negative or set it to `0` and log a warning instead?

container/go/cmd/pusher/pusher.go
itmecho: Makes sense! Pushed that change.

container/push.bzl
```diff
@@ -90,6 +90,9 @@ def _impl(ctx):
         tag = tag,
     ))
 
+    if ctx.attr.retry_count > 0:
+        pusher_args.append("-retry-count={}".format(ctx.attr.retry_count))
```
uhthomas, line 94: Maybe there should be some validation here? Might be confusing if a user provides -1 and nothing happens.

I think we can merge given this change.

uhthomas (changes requested):

EdSchouten:
@itmecho Friendly ping :-)

itmecho:
> @itmecho Friendly ping :-)

Oh sorry! I thought this commit covers that validation?

https://github.com/bazelbuild/rules_docker/commit/4e540514fef7ddd56121f81c481993ab9d3843f6

Is that no good or is there something else I've missed?

com6056:
@itmecho do you have time to wrap this up? If not, happy to pick this up!

itmecho:
> @itmecho do you have time to wrap this up? If not, happy to pick this up!

@com6056 Sorry for the delay! I've pushed some validation and a doc update, hopefully that covers it!

The build looks like it's failing on something to do with docker/debian9? Not sure if that's related to my changes or not...

com6056:
> > @itmecho do you have time to wrap this up? If not, happy to pick this up!
> 
> @com6056 Sorry for the delay! I've pushed some validation and a doc update, hopefully that covers it!
> 
> The build looks like it's failing on something to do with docker/debian9? Not sure if that's related to my changes or not...

@uhthomas any idea why that build is failing or who we can ask about it?

PR #2103, opened 2022-06-03
Add config flag to further disable transition effects caused by bazel pre-5.0 support

bduffany:
(DRAFT -- looking for feedback)

This is an attempt to address an issue that even when setting --@io_bazel_rules_docker//transitions:enable=no, we still get some config differences across targets built as a docker image dep, and targets not built as a docker image dep. This appears to be caused by this support for Bazel pre-5.0.

Looking for some initial feedback as to whether this approach seems reasonable -- if so, I can add tests and open for review.

Context: https://bazelbuild.slack.com/archives/CA3NW13MH/p1654287739299629

Even more context: we build our app twice (once for pushing to GCS, and once just to get the SHA256 digest of the app so that we know where to find it in GCS), and these path differences due to the config transition are causing the app SHA to mismatch, since we have some path references in some of the app's contents.

---

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/container/image.bzl
+++ b/container/image.bzl
@@ -768,6 +768,9 @@ _outputs["build_script"] = "%{name}.executable"
 
 def _image_transition_impl(settings, attr):
     if not settings["@io_bazel_rules_docker//transitions:enable"]:
+        if not settings["@io_bazel_rules_docker//transitions:support_bazel_pre5.0"]:
+            return {}
+
         # Once bazel < 5.0 is not supported we can return an empty dict here
         return {
             "//command_line_option:platforms": settings["//command_line_option:platforms"],
@@ -790,6 +793,7 @@ _image_transition = transition(
     implementation = _image_transition_impl,
     inputs = [
         "@io_bazel_rules_docker//transitions:enable",
+        "@io_bazel_rules_docker//transitions:support_bazel_pre5.0",
         "//command_line_option:platforms",
     ],
     outputs = [
--- a/transitions/BUILD
+++ b/transitions/BUILD
@@ -32,3 +32,8 @@ config_setting(
     name = "disabled",
     flag_values = {"@io_bazel_rules_docker//transitions:enable": "false"},
 )
+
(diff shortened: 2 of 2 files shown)


bduffany:
This is an attempt to address an issue that even when setting `--@io_bazel_rules_docker//transitions:enable=no`, we still get some config differences across targets built as a docker image dep, and targets not built as a docker image dep. This appears to be caused by this support for Bazel pre-5.0.

Looking for some initial feedback as to whether this approach seems reasonable -- if so, I can add tests.

bduffany:
@laurynaslubys @uhthomas Wondering if either of you have any thoughts on this PR? (The description includes some context)

laurynaslubys (changes requested):
I'm not really comfortable with this suggestion. There are probably more things that could be streamlined if bazel < 5 was dropped. You could just have a small patch for this in your workspace I think.

uhthomas:
There is a `native.bazel_version` which could be used instead.

Really though, I would recommend upgrading to the latest version of Bazel. Is there anything preventing you from doing so?

bduffany:
@laurynaslubys that's understandable. Are there any plans for rules_docker to drop pre-5.0 support in the near-term future?

If not, I'm open to other ideas -- it would be nice to find some workaround that doesn't require a patch. For other folks using this flag, it'd be nice if the transition effects could be fully disabled so that builds both as a docker dep or not can be built in the same configuration (which has implications for caching, checksums, etc.)

@uhthomas It looks as though `native.bazel_version` only works in a repository context. (I tried using it and got an error.) Also, I am on the latest version of Bazel -- the problem is that this code which is intended to support earlier versions of Bazel is causing a configuration difference (the directory becomes `bazel-out-ST-{HASH}` instead of `bazel-out`).

Issue #2078, opened 2022-05-04
`nodejs_image` node repository settings don't match current master of `rules_nodejs`


psigen:
In the latest release and current master of `rules_nodejs`, there is not a `bin/node_repo_args.sh`, and the default name for the repository is suffixed with a platform, e.g. `nodejs_linux_amd64`:
* https://github.com/bazelbuild/rules_nodejs/blob/e4842c13c0ea16af1eb620c069d7e9116f566a1a/nodejs/repositories.bzl#L304
* https://github.com/bazelbuild/rules_docker/issues/2016#issuecomment-1034747672

This seems to contradict the default arguments in the `rules_docker` `nodejs_image` rule currently:
https://github.com/bazelbuild/rules_docker/blob/8f729088a71d2a243a657c10e0618a0ee3fea367/nodejs/image.bzl#L122-L123

Which means that we seem to need the following two options on every `nodejs_image` target to get them to compile:
```python
nodejs_image(
    [...]
    include_node_repo_args = False,
    node_repository_name = "nodejs_linux_amd64",
)
```

Otherwise I see one of the following two errors:
```
ERROR: [...]/BUILD.bazel:30:13: no such package '@nodejs//': The repository '@nodejs' could not be resolved: Repository '@nodejs' is not defined and referenced by '//my_node_target:image.2'
ERROR: Analysis of target '//my_node_target:image' failed; build aborted: 
```

```
ERROR: [...]/BUILD.bazel:30:13: no such target '@nodejs_linux_amd64//:bin/node_repo_args.sh': target 'bin/node_repo_args.sh' not declared in package '' defined by [...]/.cache/bazel/_bazel_pras/2e5a3dfbf28c2d99ea1eb601cd4191f0/external/nodejs_linux_amd64/BUILD.bazel and referenced by '//my_node_target:image.2'
```

---

Versions I am testing with:
- `rules_docker`: https://github.com/bazelbuild/rules_docker/releases/tag/v0.24.0
- `rules_nodejs`: https://github.com/bazelbuild/rules_nodejs/releases/tag/5.4.2

sashamor:
This issue was introduced in #1982.

PR #2128, opened 2022-07-17
Update `nodejs_image` repository settings to match current latest version of `rules_nodejs`

sashamor:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #2078 

## What is the new behavior?

See issue for more details.

- [X] Yes
- [ ] No

Removes the `include_node_repo_args` argument, which will break rules that use it. But this is probably only used for rules that disable it, so it'll be good for those rules to remove it when they update.

Remove the `include_node_repo_args = False` flag, which most repos include to fix this bug, and is no longer needed now that the bug is actually fixed.

### DIFF
--- a/nodejs/image.bzl
+++ b/nodejs/image.bzl
@@ -120,8 +120,7 @@ def nodejs_image(
         binary = None,
         launcher = None,
         launcher_args = None,
-        node_repository_name = "nodejs",
-        include_node_repo_args = True,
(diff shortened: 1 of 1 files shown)


samueljim (approved):
lgtm

uhthomas (approved):

Issue #2117, opened 2022-06-29
NodeJS images don't work in readonly filesystems


sashamor:
# 🐞 bug report

### Affected Rule

The issue is caused by the rule:  `nodejs_image` and other `nodejs` rules.

### Is this a regression?

No.

### Description

`nodejs_image` requires a writeable filesystem. Most container environments will not run containers with writeable filesystems. The kubernetes settings `readOnlyRootFilesystem: true` and `runAsNonRoot: true` are standard requirements for production-ready workloads.

Running without a writeable filesystem gives the error:
```
[link_node_modules.js] An error has been reported: [Error: EROFS: read-only file system, mkdir 'node_modules'] {
  errno: -30,
  code: 'EROFS',
  syscall: 'mkdir',
  path: 'node_modules'
} Error: EROFS: read-only file system, mkdir 'node_modules'
```
as `nodejs` runs the linker and tries to create a `node_modules` folder with symlinks: https://github.com/bazelbuild/rules_nodejs/blob/stable/internal/node/launcher.sh#L231

Without this, Bazel nodejs images are not secure enough to run in production systems.

## 🔬 Minimal Reproduction

Create a `nodejs_image` rule, e.g.:

```starlark
nodejs_image(
    name = "image",
    binary = ":app",
    # These two parameters are needed until bazelbuild/rules_docker#2078 is fixed.
    include_node_repo_args = False,
    node_repository_name = "nodejs_linux_amd64",
)
```

then run:
```
docker run --read-only bazel/path/to/nodejs/image
```
And you should see the error.
## 🔥 Exception or Error

```
[link_node_modules.js] An error has been reported: [Error: EROFS: read-only file system, mkdir 'node_modules'] {
  errno: -30,
  code: 'EROFS',
  syscall: 'mkdir',
  path: 'node_modules'
} Error: EROFS: read-only file system, mkdir 'node_modules'
```

Disabling the linker does fix the issue:

```starlark
nodejs_image(
    name = "image",
    binary = ":app",
    # These two parameters are needed until bazelbuild/rules_docker#2078 is fixed.
    include_node_repo_args = False,
    node_repository_name = "nodejs_linux_amd64",
    templated_args = [
        "--nobazel_run_linker",
        "--nobazel_node_patches",
    ],
)
```

But this breaks imports of local dependencies, since the node linker is what creates the symlinks in the locations needed for `require()` to discover them. So produces errors like:

```
[...]

PR #2129, opened 2022-07-17
Add node_modules linker to `nodejs_image`

sashamor:
## PR Checklist

## PR Type

- [X] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Currently, `nodejs` images can't run on readonly filesystems. This is a blocking issue for using Bazel nodejs images in production workloads.

This is because `nodejs` binary rules run a "linker" script on startup (see the readme for more details - https://github.com/bazelbuild/rules_nodejs/tree/stable/internal/linker).

To skip this step, `--nobazel_run_linker` can be passed to `templated_args` in `nodejs_image`, e.g.:

```starlark
load("@io_bazel_rules_docker//nodejs:image.bzl", "nodejs_image")

nodejs_image(
    name = "img",
    templated_args = [
        "--nobazel_run_linker",
    ],
)
```

However, this causes another issue - local dependencies (e.g. `require("@myrepo/foo")`) can't be resolved, since they aren't symlinked inside the `node_modules` folder as part of this linker step.

Issue Number: #2117

## What is the new behavior?

To fix this, we effectively turn the `.js` linker script into a `.bzl` script, reading the manifest file contents and linking them. This allows a `nodejs_image` to be created that will run on a readonly filesystem. With this fix, the following image is readonly-fs compatible:

```starlark
load("@io_bazel_rules_docker//nodejs:image.bzl", "nodejs_image")

nodejs_image(
    name = "img",
    data = [":js_lib"],
    entry_point = ":index.js",
    # These two parameters are needed until bazelbuild/rules_docker#2078 is fixed.
    include_node_repo_args = False,
    node_repository_name = "nodejs_linux_amd64",
    templated_args = [
        # This is needed to skip the linker, which creates the node_modules directory.
        "--nobazel_run_linker",
        # This is needed to skip writing over the node binary.
        "--nobazel_node_patches",
        # We need this so Bazel uses the node_modules resolver as though the linker ran (which it did, in
        # a previous step).
        "--nobazel_patch_module_resolver",
    ],
)
```

This creates a `node_modules` folder in the resulting image, in the same place its created when the binary runs - but instead, its created as part of the build process, rather than relying on the `launcher.sh` script to create it at runtime.

This will have no impact on images that run without the `--nobazel_run_linker` flag, since they will overwrite these files anyway. It's very likely that images currently being built with `--nobazel_run_linker` contain this bug but either:

a. Don't know they have this issue, since they don't import any local dependencies; or
b. Have used their own custom hacks (such as a custom base image) to work around this issue.

So, either way, it's better for the files to exist there :) We could even update the `nodejs` flags to default to `--nobazel_run_linker` and `--nobazel_patch_module_resolver`, since the linker "step" of `launcher.sh` has now already run. Although, there's no harm in running them twice.

See the issue for more details.

- [ ] Yes
- [X] No

## Other information

Enabling nodejs images to run with readonly filesystems is crucial in allowing them to be used in production workloads. In most security scanning systems for kubernetes, the first critical risk issue raised is that apps are run with writeable filesystems. Without this fix, these built containers can't be deployed to these systems, rendering the `nodejs` built images effectively useless for large-scale systems.

### DIFF
--- a/nodejs/image.bzl
+++ b/nodejs/image.bzl
@@ -99,8 +99,27 @@ def _npm_deps_runfiles(dep):
             depsets.append(pkg[NpmPackageInfo].sources)
     return depset(transitive = depsets)
 
+def _npm_deps_symlinks(reference_dir, binary):
+    # Make a node_modules folder with symlinks of a mix of local and remote
+    # dependencies, performing the job of the rules_nodejs linker.
+    node_modules = "/".join([reference_dir, "node_modules"])
+
+    # Get the action that writes the module mappings and parse it.
+    filewrite_actions = [a for a in binary.actions if a.mnemonic == "FileWrite"]
+    symlinks = {}
+    if len(filewrite_actions) == 1 and filewrite_actions[0].outputs.to_list()[0].basename.endswith(".module_mappings.json"):
+        root_module_sets = json.decode(filewrite_actions[0].content)["module_sets"][""]
+
+        # Replace the bazel-out/k8-.../bin prefix with the binary directory.
+        symlinks.update({
+            node_modules + "/" + module_name: "/".join([reference_dir] + root_module_sets[module_name].split("/")[3:])
+            for module_name in root_module_sets
+        })
+
+    return symlinks
+
 def _npm_deps_layer_impl(ctx):
-    return lang_image.implementation(ctx, runfiles = _npm_deps_runfiles, emptyfiles = _emptyfiles)
+    return lang_image.implementation(ctx, runfiles = _npm_deps_runfiles, emptyfiles = _emptyfiles, symlinks = _npm_deps_symlinks)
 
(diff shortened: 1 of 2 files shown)


nodejs/image.bzl
```diff
@@ -99,8 +99,27 @@ def _npm_deps_runfiles(dep):
             depsets.append(pkg[NpmPackageInfo].sources)
     return depset(transitive = depsets)
 
+def _npm_deps_symlinks(reference_dir, binary):
```
sashamor, line 102: I'm open to feedback of the implementation of this. I'm not sure if this is the best place for it, and I don't like the search logic to find a "FileWrite" command that writes a ".module_mappings.json" file. But I'm new to Starlark, and this works, so very open to feedback on the best way to write it and how to add tests :)

PR #2098, opened 2022-05-28
Make golden image digests work with Python 3.9

bpcreech:
This fixes #2097 wherein `image_test` fails with Python 3.9+

For future reference this commit was generated with:

```
bazel test tests/container:image_test
grep sha256: .../testlogs/tests/container/image_test/test.log  | grep Assertion | sed -r 's@AssertionError: .sha256:(.*). != .sha256:(.*).@s/\2/\1/@' > /tmp/seds
sed -f /tmp/seds -i tests/container/image_test.py

# then repeat the above a couple times because some of the digest errors are only exposed after you fix the first ones.
```

## PR Checklist

Please check if your PR fulfills the following requirements:

- [X] Tests for the changes have been added (for bug fixes / features)
- [ ] (n/a) Docs have been added / updated (for bug fixes / features)

## PR Type

What kind of change does this PR introduce?

- [X] Bugfix
- [ ] Feature
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [X] Build related changes
- [ ] CI related changes
- [ ] Documentation content changes
- [ ] Other... Please describe:

## What is the current behavior?

This test fails with Python 3.9 and it looks like the best fix is to change the golden image digests to match the new reality.

Issue Number: #2097

## What is the new behavior?

Just new digests in the test to match CPython 3.9 `tarfile` behavior.

## Does this PR introduce a breaking change?

- [ ] Yes
- [X] No

Technically Python 3.9 creates a breaking changes in that the image digests generated by `rules_docker` aren't reproducible across the upgrade from Python <=3.8, but this PR didn't create and doesn't attempt to change that situation.

PR #2148, opened 2022-08-22
make it easier to update image_test hashes

zevdg:
We had to do this recently with #2098.  This is also useful for organisations that link against different zlib implementations (namely zlib-ng). This change makes it possible to easily swap out the hashes without modifying the test code.

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Feature
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

test hashes are hardcoded into test, and must be updated manually

Issue Number: N/A

## What is the new behavior?
test hashes are now encoded in testdata/goldens and can be regenerated by the test when needed. 

- [ ] Yes
- [x] No


zevdg:
Ok, this is ready for reviewer now.

testdata/BUILD
```diff
@@ -1102,3 +1102,8 @@ bzl_library(
     srcs = ["utils.bzl"],
     visibility = ["//visibility:private"],
 )
+
+filegroup(
```
gravypod, line 1106: Would it be possible to create an image hash json file? This way we could have `testdata/image-hash.json` which could have:

```
{
  "test_base_with_entrypoint": "..."
}
```

This might be a little easier to use outside of google3 but I'm not sure. WDYT?

zevdg: Can you explain how that would make it easier? These files are only meant to be used in tests, and the test-helper I added in this CL is designed to be generic.  It's essentially just a simple python port of [a generic test helper like this](https://pkg.go.dev/github.com/xorcare/golden#section-readme). I don't see how making it less generic makes it better.

Issue #2124, opened 2022-07-07
java_image args broken in 0.23.0+


joca-bt:
# 🐞 bug report

### Affected Rule

`java_image`

(And possibly all other language rules.)

### Is this a regression?

Yes.

0.22.0: worked fine
0.23.0: broken

### Description

`java_image` `args` parameter broken in 0.23.0 (probably by https://github.com/bazelbuild/rules_docker/pull/1957).

Previously, adding `args` to `java_image` would result in those args being propagated to the underlying `java_binary` rule and hence to the JVM when starting the application.

This behaviour was broken in 0.23.0 and these args are now being passed to docker.

0.22.0:
```
> bazel run :image
hello there!
[--stuff, --stuff]
```

0.23.0:
```
> bazel run :image
unknown flag: --stuff
See 'docker run --help'.
```

I am guessing this happening with other language rules as well.

## 🔬 Minimal Reproduction

`BUILD`:
```
load("@io_bazel_rules_docker//java:image.bzl", "java_image")
java_library(name = "lib", srcs = [ "Main.java" ])
java_binary(name = "bin", main_class = "demo.Main", runtime_deps = [ ":lib" ])
java_image(name = "image", base = "@java//image", main_class = "demo.Main", args = [ "--stuff" ], runtime_deps = [ ":lib" ])
```

`Main.java`:
```
package demo;
import java.util.Arrays;
public class Main {
    public static void main(String[] args) {
        System.out.println("hello there!");
        System.out.println(Arrays.toString(args));
    }
}
```

`WORKSPACE`:
```
load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
http_archive(
    name = "io_bazel_rules_docker",
    # sha256 = "59536e6ae64359b716ba9c46c39183403b01eabfbd57578e84398b4829ca499a",
    # strip_prefix = "rules_docker-0.22.0",
    # url = "https://github.com/bazelbuild/rules_docker/releases/download/v0.22.0/rules_docker-v0.22.0.tar.gz",
    sha256 = "85ffff62a4c22a74dbd98d05da6cf40f497344b3dbf1e1ab0a37ab2a1a6ca014",
    strip_prefix = "rules_docker-0.23.0",
    url = "https://github.com/bazelbuild/rules_docker/releases/download/v0.23.0/rules_docker-v0.23.0.tar.gz",
)
load("@io_bazel_rules_docker//repositories:repositories.bzl", "repositories")
repositories()
[...]

PR #2184, opened 2022-11-06
De-duplicate run args passed to bazel image targets.

psigen:
When 'bazel run' is called, it expects that the args specified in the BUILD file are not known to the run command and will prepend them to all other arguments.

However, we have already baked those commands into the image, because users expect that 'args' in the case of an image target will be included in the image.  This templated script de-duplicates the args specified in the BUILD file by passing itself the number of args already included in the image.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_docker/issues/2124

In the case of an image target being passed arguments, the arguments will actually be passed to the image
target by bazel twice, once at build time and another at runtime.  This interferes with the expected order of
arguments (because command arguments are inserted before docker run arguments)

## What is the new behavior?

The new behavior should match the documentation.

- [ ] Yes
- [x] No

### DIFF
--- a/container/incremental_load.sh.tpl
+++ b/container/incremental_load.sh.tpl
@@ -249,6 +249,16 @@ function read_variables() {
 # This is not executed if the single argument --norun is passed or
 # no run_statements are generated (in which case, 'run' is 'False').
 if [[ "%{run}" == "True" ]]; then
+  # When 'bazel run' is called, it expects that the args specified in the BUILD
+  # file are not known to the run command and will prepend them to all other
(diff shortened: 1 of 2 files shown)


psigen:
Please, if anyone has a good way to add unit tests for this, please help me add them

psigen:
Hmm, I will try to rebase this to bring it up to date

Issue #2200, opened 2022-12-29
Issues with `install_pkgs`'s parameter `output_image_name` in multiple image builds


lukokr-aarch64:
# 🐞 bug report

### Affected Rule

 The issue is caused by the rule:  `install_pkgs`

### Is this a regression?
No.

### Description
We have found a potenital race condition in a large repository when `output_image_name` is not unique. 
When running in parallel the following is prone to race conditions:
```sh
reset_cmd $image_id $cid %{output_image_name} # another target overwrites %{output_image_name} 
"$DOCKER" $DOCKER_FLAGS save %{output_image_name} > %{output_file_name} # wrong target is saved
```

I understand that the user is expected to know that `output_image_name` is the unqiue name of the image as loaded into the docker runtime, however this is not stated clearly and as such in large codebases with many images it is possible that duplication will occur. 

I would propose that `output_image_name` should be made optional, and when not specified an image name based on the label path will be created similiar to the intermediate container used for `container_test`. Furthermore, if the image is not expected to be used, would you consider removing this from the API and deleteing the image to reduce bloat in the docker runtime?

I have a WIP patch I am happy to put up if you agree with the above. Please let me know 👍 

## 🔬 Minimal Reproduction

This is a bit tricky to reproduce as it's timing related. I have uploaded a patch which has reproduced this issue with a hack to exacerbate the issue.
[repro.patch](https://github.com/bazelbuild/rules_docker/files/10320062/repro.patch)

I have added a random sleep to highlight the race condition:
```sh
reset_cmd $image_id $cid %{output_image_name}
sleep $(( ($RANDOM % 4 * 4) + 1 ))s
"$DOCKER" $DOCKER_FLAGS save %{output_image_name} > %{output_file_name}
```

This of course only happens when `output_image_name` is not unique. 

## 🔥 Exception or Error
When the race bug happens, the wrong image is saved, for the above patch it results in errors such as:
<pre><code>
==================================
====== Test file: wget.yaml ======
==================================
=== RUN: Command Test: wget
--- FAIL
duration: 1.169700321s
Error: Error creating container: API error (400): failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "wget": executable file not found in $PATH: unknown
</code></pre>

## 🌍  Your Environment

**Operating System:**

<pre>
  <code>
Ubuntu 20.04.5 LTS
  </code>
</pre>

**Output of `bazel version`:**

<pre>
  <code>
[...]

PR #2202, opened 2023-01-03
Fixes `output_image_name` in install_pkgs to be optional

lukokr-aarch64:
Previously `output_image_name` needed to be unique across the entire build tree. When building many containers in parallel this could result in clashes between the containers without warning or error.

This patches changes `output_image_name` to be optional, with the default value being set to the current target full label making it unique across large builds.

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: #2200

## What is the new behavior?

When `output_image_name` is not provided by the user, it will default to full label path similiar to `container_test`.

- [ ] Yes
- [x] No

### DIFF
--- a/docker/package_managers/install_pkgs.bzl
+++ b/docker/package_managers/install_pkgs.bzl
@@ -66,13 +66,17 @@ def _impl(ctx, image_tar = None, installables_tar = None, installation_cleanup_c
       image_tar: File, overrides ctx.file.image_tar
       installables_tar: File, overrides ctx.file.installables_tar
       installation_cleanup_commands: str, overrides ctx.attr.installation_cleanup_commands
-      output_image_name: str, overrides ctx.attr.output_image_name
+      output_image_name: str, overrides the image name
       output_tar: File, overrides ctx.outputs.out
     """
+
+    # Construct the name of the image if not given based on the full label path with :output tag.
+    default_output_image_name = (ctx.label.package + "_" + ctx.label.name).replace(":", "_").replace("@", "_").replace("/", "_") + ":output"
+
     image_tar = image_tar or ctx.file.image_tar
     installables_tar = installables_tar or ctx.file.installables_tar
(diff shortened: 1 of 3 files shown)


docker/package_managers/install_pkgs.bzl
```diff
@@ -167,8 +171,7 @@ _attrs = {
         default = "",
     ),
     "output_image_name": attr.string(
-        doc = ("Name of container_image produced with the packages installed."),
-        mandatory = True,
+        doc = ("Name of container_image produced with the packages installed. If not specified, resolves to the current target target label. *Must be unique* across the entire build tree."),
```
uhthomas, line 174: nit: not your fault, but why is this string wrapped in `()`?

uhthomas:
This seems like a reasonable change, though I just don't have enough context to know if this could break things for existing users. Do you have any insight @gravypod?

lukokr-aarch64:
Thanks for the review @uhthomas!

> This seems like a reasonable change, though I just don't have enough context to know if this could break things for existing users. Do you have any insight @gravypod?

For existing users the string was mandatory before this change, so the values they have been using up until this point will remain unchanged. 

Part of the original issue was trying to figure out with the maintainers if we can ditch the parameter all together, but this change is intended to be non-breaking.

docker/package_managers/install_pkgs.bzl
lukokr-aarch64: Ah this is actually my bad, looking at other rules it's to allow breaking the help text over multiple lines. Let me amend real quick

lukokr-aarch64: Fixed the formatting to match the other rules. LMK if that's okay to resolve now 👍

uhthomas:
Makes sense, thank you for your hard work.

uhthomas (approved):

PR #2203, opened 2023-01-04
Use 'cfg=target' for pusher binary

anguslees:
In container_push rule, `_pusher` executable is copied into the output
runfiles, but not directly invoked _by_ the container_push rule.  This
binary could potentially be cross-compiled by other remote executors,
and should use the 'target' config in this action.

### DIFF
--- a/container/push.bzl
+++ b/container/push.bzl
@@ -205,7 +205,7 @@ container_push_ = rule(
         ),
         "_pusher": attr.label(
             default = "//container/go/cmd/pusher",
-            cfg = "exec",
+            cfg = "target",
             executable = True,
             allow_files = True,
         ),
--- a/contrib/push-all.bzl
+++ b/contrib/push-all.bzl
@@ -126,7 +126,7 @@ container_push = rule(
         ),
         "_pusher": attr.label(
             default = Label("//container/go/cmd/pusher"),
-            cfg = "exec",
+            cfg = "target",
             executable = True,
             allow_files = True,
         ),


uhthomas:
Maybe I'm misunderstanding but shouldn't this be `exec`?

anguslees:
> Maybe I'm misunderstanding but shouldn't this be exec?

It's very possible that I'm misunderstanding too - so please double-check 😅 

As an executable rule, I think [the `container_push` rule](https://github.com/bazelbuild/rules_docker/blob/48ad6d6df43d1e4b9feeec961995aef01dd72080/container/push.bzl#L38-L141) [expands a template](https://github.com/bazelbuild/rules_docker/blob/48ad6d6df43d1e4b9feeec961995aef01dd72080/container/push.bzl#L118-L126) to produce a shell script - which is usually later executed by `bazel run`.  This template/shell script [_references_](https://github.com/bazelbuild/rules_docker/blob/48ad6d6df43d1e4b9feeec961995aef01dd72080/container/push.bzl#L123) the _pusher executable, but does not actually execute (`ctx.actions.run`) _pusher itself - [unlike _digester](https://github.com/bazelbuild/rules_docker/blob/48ad6d6df43d1e4b9feeec961995aef01dd72080/container/push.bzl#L98-L105), for comparison.

This means that (hypothetically), the `container_push` rule can run on one executor, and the _pusher binary can be built on another executor.  The only thing that is important, is that the _pusher binary works on the _target_ platform of the `container_push` rule.

uhthomas:
Wow.

uhthomas (approved):
Thank you :)

PR #2205, opened 2023-01-05
image.bzl: implement `tag_name`

motiejus:
rules_docker decides the name of the resulting docker image:

    bazel/{target}:{name}

The repository (`bazel/` by default) and `{target}` are fine. However, we want to customize the tag part -- the thing after the colon. This patch allows customizing the last part.

Why?
====

In our case, we are always building two images for two architectures using transitions. Here is an extract of a macro that builds two containers:

Then we build two containers using outgoing transitions:

    container_image(
        name = "{name}-post_transition".format(name = name),
        ...,
    )

    multiplatform_image(
        name = "{name}-amd64".format(name),
        src_image = ":{name}-post_transition".format(name = name),
        ...,
    )

    multiplatform_image(
        name = "{name}-arm64".format(name = name),
        src_image = ":{name}-post_transition".format(name = name),
        ...,
    )

Both images (`-amd64` and `-arm64`) build correctly, but the resulting image name, according to Docker, happens to be
`bazel/{target}:{name}-post_transition`. So the internal implementation details of the transitions leak to the resulting images.

This patch makes it possible to rename the post-transition image name to something we like more, e.g. the architecture name, or remove the suffix altogether.

### DIFF
--- a/container/image.bzl
+++ b/container/image.bzl
@@ -288,7 +288,8 @@ def _impl(
         workdir = None,
         user = None,
         null_cmd = None,
-        null_entrypoint = None):
+        null_entrypoint = None,
+        tag_name = None):
     """Implementation for the container_image rule.
 
     You can write a customized container_image rule by writing something like:
@@ -346,6 +347,7 @@ def _impl(
         user: str, overrides ctx.attr.user
         null_cmd: bool, overrides ctx.attr.null_cmd
         null_entrypoint: bool, overrides ctx.attr.null_entrypoint
+        tag_name: str, overrides ctx.attr.tag_name
     """
     name = name or ctx.label.name
     base = base or ctx.attr.base
@@ -367,6 +369,7 @@ def _impl(
     build_script = ctx.outputs.build_script
     null_cmd = null_cmd or ctx.attr.null_cmd
     null_entrypoint = null_entrypoint or ctx.attr.null_entrypoint
+    tag_name = tag_name or ctx.attr.tag_name
 
     # If this target specifies docker_run_flags, they are always used.
     # Fall back to the base image's run flags if present, otherwise use the default value.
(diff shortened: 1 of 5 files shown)


container/image.bzl
```diff
@@ -744,6 +749,12 @@ _attrs = dicts.add(_layer.attrs, {
 
         This field supports stamp variables.""",
     ),
+    "override_tag": attr.string(
+        doc = """Override final tag name.
+
+        By default the final docker image name is bazel/{target}:{name}.
+        This flag overrides the string after the colon.""",
+    ),
```
linzhp, line 757: Can we just call it `tag` or `tag_name` and document that if not specified, it would be {name}?

container/image.bzl
```diff
@@ -459,8 +462,10 @@ def _impl(
             null_cmd = null_cmd,
         )
 
-    # Construct a temporary name based on the build target.
-    tag_name = "{}:{}".format(_repository_name(ctx), name)
+    # Construct a temporary name based on the build target. This is the name
+    # of the docker container.
+    final_tag = tag_name if tag_name else name
+    tag_name = "{}:{}".format(_repository_name(ctx), final_tag)
```
linzhp, line 468:
```suggestion
    if not tag_name:
        tage_name = "{}:{}".format(_repository_name(ctx), name)
```

Can we do this to allow users to specific the whole tag, not just part of it?

motiejus: I don't think it's a good idea:
1. It already provides `repository`, to override the default `bazel/`. We can't delete that.
2. It is non-trivial to generate the string between `bazel/` and `:` if, say, the user wants to change only the tag (our case).

linzhp, line 468:
```suggestion
    # Construct the name of the docker container
    container_name = "{}:{}".format(_repository_name(ctx), tag_name if tag_name else name)
```

How about calling the variable `container_name`? From the [syntax](https://ktomk.github.io/pipelines/doc/DOCKER-NAME-TAG.html), "tag_name" is just the part after `:`. The full string is "container".

container/image.bzl
```diff
@@ -744,6 +749,9 @@ _attrs = dicts.add(_layer.attrs, {
 
         This field supports stamp variables.""",
     ),
+    "tag_name": attr.string(
+        doc = """Override final tag name. If unspecified, is set to name.""",
```
linzhp, line 753:
```suggestion
        doc = """Tag name of the container, default to name of the target.""",
```

linzhp (approved):

PR #2209, opened 2023-01-11
run.bzl: allow passing extra_deps to all the container_run rules

manuelnaranjo:
It's very convenient to be able to mount generated files into the build context so those can be consumed without impacting the size of the resulting image (for instance to execute commands on some intermediate artifacts)

Example:

```
container_run_and_commit_layer(
    name = "foo",
    image = "some-image",
    docker_run_flags = [
        # when mounting from the runfiles we may have a directory with symlinks, so we need to mount
        # the actual file
        "-v ${PWD}/$(location :bar/bar.tar):/data/bar.tar"
    ],
    extra_deps = [
        ":bar"
    ],
    commands = [
        "ls /data/bar.tar",
    ],
)
```

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Only `container_run_and_extract` allows to pass `extra_deps` which allows for instance to mount a directory into the container that's being built

Issue Number: 2156

## What is the new behavior?
Now `container_run_and_commit` and `container_run_and_commit_layer` also allows for this behaviour

- [ ] Yes
- [x] No

### DIFF
--- a/docker/util/run.bzl
+++ b/docker/util/run.bzl
@@ -34,6 +34,23 @@ load(
     "docker_path",
 )
 
+def _get_docker_run_flags(ctx, docker_run_flags = None):
+    """Helper method to expand docker_run_flags for mounting
+
+    When passing -v to docker_run_flags and some of those are $(location
+    <foo>) then the user will expect that we expand them properly
+
+    The challenge here is that ctx.expand_location expects Labels and
+    ctx.actions.run.inputs expects Files so it's not fully compatible
+    with the struct approach we use to expose the rule
+    """
+    docker_run_flags = docker_run_flags or ctx.attr.docker_run_flags
+
+    if ctx.attr.extra_deps:
+        docker_run_flags = [ctx.expand_location(arg, targets = ctx.attr.extra_deps) for arg in docker_run_flags]
+
+    return docker_run_flags
+
 def _extract_impl(
         ctx,
         name = "",
@@ -67,7 +84,7 @@ def _extract_impl(
     name = name or ctx.label.name
     image = image or ctx.file.image
     commands = commands or ctx.attr.commands
-    docker_run_flags = docker_run_flags or ctx.attr.docker_run_flags
+    docker_run_flags = _get_docker_run_flags(ctx, docker_run_flags)
     extract_file = extract_file or ctx.attr.extract_file
     output_file = output_file or ctx.outputs.out
     script = script_file or ctx.outputs.script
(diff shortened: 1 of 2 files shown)


uhthomas:
Sure, this change looks good to me. Is there any way we can test this behaviour?

manuelnaranjo:
> Sure, this change looks good to me. Is there any way we can test this behaviour?

I'll add more tests to the battery at `tests/docker/util/BUILD` didn't realize tests where there

manuelnaranjo:
@uhthomas there's a problem with testing this in CI, turns out where' doing docker in docker, which means if we're trying to mount something into a new image the path we pass to docker has to be a path available in the host. I could create a python_binary that parses /proc/mounts for the overlay layer and find the merged path out of the workdir or the diff layer.
```
I have no name!@43db76cdad66:/$ cat /proc/mounts  | grep overlay | awk '{ print $4 }'
rw,relatime,lowerdir=/var/lib/docker/overlay2/l/JA3ACMWQ362U36GUCG3H4NEMWA:/var/lib/docker/overlay2/l/YEQGBSNRFDAUC7Q5FCPT6ZCEHY:/var/lib/docker/overlay2/l/ZT7D7AJYKC2NWFTVZPJQ3GTH3C:/var/lib/docker/overlay2/l/KTSL4EIQJ2TK3KHOGCONRRZWGG:/var/lib/docker/overlay2/l/G3PMNKGOUJLSHJJX7TRMVLK77H:/var/lib/docker/overlay2/l/WGIKEK2IJN7IOWWWYDNDX6BLEF:/var/lib/docker/overlay2/l/GDB7KVGJRRLVVTM4LHMLU5B5O7:/var/lib/docker/overlay2/l/OCJ6KCDTLG3BI6754UVW7XTHGC:/var/lib/docker/overlay2/l/YKU2Y4SQYJG45YE62EYWHY7BE2,upperdir=/var/lib/docker/overlay2/15241b658ad6735f23f0b2d7e466a5ed0c8ee77d9ebdafcdcaa5a29c7c103f77/diff,workdir=/var/lib/docker/overlay2/15241b658ad6735f23f0b2d7e466a5ed0c8ee77d9ebdafcdcaa5a29c7c103f77/work
```

in the host:
```
$ sudo ls /var/lib/docker/overlay2/15241b658ad6735f23f0b2d7e466a5ed0c8ee77d9ebdafcdcaa5a29c7c103f77/merged
bin  boot  dev  etc  home  lib  lib32  lib64  libx32  media  mnt  opt  proc  root  run  sbin  srv  sys  tmp  usr  var
```

But it sounds like a very awful hack, and this may not work on RBE setups.

I guess that if we want to test is that extra_deps works just generating a binary which we can call through `docker_run_flags` where the output is empty would be enough. What do you think?

PR #2211, opened 2023-01-12
Allow set `container_image` labels by json file

bozaro:
## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

## What is the new behavior?
Allow set docker labels by .json file.

- [ ] Yes
- [x] No


uhthomas:
Is there an issue open where this feature has been discussed? Could you explain why this may be useful?

bozaro:
> Is there an issue open where this feature has been discussed?
No.

> Could you explain why this may be useful?
We use file for labels in two places:
 - set Docker labels for image garbage collector policy (we have differ policy for master and feature-branches. Branch type passed by stable stamping variables);
 - set marker for "upload if not uploaded yet" logic.

Little bit about "upload if not uploaded yet" workflow:
 - we check image in docker registry with upload marker (some uniuqe for build value) and return passed upload marker if image is not uploaded or image marker from uploaded image;
 - upload marker passed to labels;
 - rule makes push to registry.

As a result, we get the following effect:
 - initially image will be upload to registry;
 - if image already in registry, we got CACHE HIT (same image and upload marker) and push step will be skipped without downloading all layers to worker;
 - if image was remove from registry by GC, we got CACHE MISS (same image but new upload marker) and it will be upload.

bozaro:
Changes:
 - rebased to `master`;
 - use `labels_file` attribute instead of reusing `label_files` attribute.

Issue #309, opened 2018-02-07
How to expose ports for go_image?


chaitanya9186:
#305

mattmoor:
You can compose `go_image` with `container_image`, e.g.

```python
go_image(
    name = "foo",
    ...
)

container_image(
    name = "bar",
    base = ":foo",
    ports = [...],
)
```

chaitanya9186:
But  bazel run :bar target not running the container.

mattmoor:
This was harder to make the default for `container_image` (really: `docker_build`) which has been around for years.  You can enable that for that rule with: `legacy_run_behavior = False`.

You can also invert the ordering to get this:
```python
load("@io_bazel_rules_docker//go:image.bzl", GO_DEFAULT_BASE="DEFAULT_BASE")

container_image(
    name = "bar",
    base = GO_DEFAULT_BASE,
    ports = [...],
)

go_image(
    name = "foo",
    base = ":bar",
    ...
)
```

chaitanya9186:
I have tried the invertion approach too. If I run `docker run -d -p 8888:8888 <image>` I able to query with localhost:8888, but if I run `bazel run :foo` I am unable to query localhost:8888.

I referred the generated config file, it has the `ExposedPorts: {"8888:8888/tcp": {}}` filed. 

And I also checked `docker ps -a` list,  with manual `docker run ` I see `40ef854fdb77        <image>      "/app/server/serve..."   20 seconds ago      Up 19 seconds       0.0.0.0:8888->8888/tcp`.

with `bazel run :foo` I see `8fcc1ca061f0        <image>      "/app/server/serve..."   6 seconds ago       Up 5 seconds        8888/tcp`

I am not sure, what's happening, but I get a feeling that I miss some other setting.

mattmoor:
We don't expose network on the Docker daemon by default when a `lang_image` is run (except `war_image`).

I feel like this is a mistake, particularly since you can't override it.  We should be doing what we do for `war_image` for all `lang_image`.  I'll just [fix this](https://github.com/bazelbuild/rules_docker/pull/312).

Another thing to beware of is that ctrl+C of the container run by bazel won't kill it.  This is a longstanding Bazel bug, that I'll dig up and ping.

mattmoor:
here's the thread: https://github.com/bazelbuild/bazel/issues/3519

wkiser:
[...]

PR #2242, opened 2023-04-03
Add configurable entrypoint to py3_image

dbruggner:
This makes entrypoint for py3_image configurable so that a user can provide path to a hermetic python environment.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

Currently the Python binary is fixed as `/usr/bin/python`.

## What is the new behavior?

The Python binary can be choses by overrideing the `entrypoint` for example to point to a hermetic Python.

- [ ] Yes
- [x] No

### DIFF
--- a/python3/image.bzl
+++ b/python3/image.bzl
@@ -73,7 +73,7 @@ DEFAULT_BASE = select({
     "//conditions:default": "@py3_image_base//image",
 })
 
-def py3_image(name, base = None, deps = [], layers = [], env = {}, **kwargs):
+def py3_image(name, base = None, deps = [], layers = [], env = {}, entrypoint = ["/usr/bin/python"], **kwargs):
     """Constructs a container image wrapping a py_binary target.
 
   Args:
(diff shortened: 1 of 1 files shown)


DerekTBrown:
Wow, I came here to propose more-or-less the same thing. +1.

x-mass:
> It seems like the current direction is to override the entrypoint with an additional layer: [#309 (comment)](https://github.com/bazelbuild/rules_docker/issues/309#issuecomment-477593393)

In my opinion, overridden entrypoint is not the best approach, because it displaces the [python binary target](https://github.com/bazelbuild/rules_docker/blob/8e70c6bcb584a15a8fd061ea489b933c0ff344ca/lang/image.bzl#L208) to be executed from the final entrypoint command. This way you should also add `/app/path/to/target/image.binary` to entrypoint.

I see symlinking `/bin/usr/python` to actual python you want to use (as mentioned [here](https://github.com/bazelbuild/rules_docker/issues/1247#issuecomment-786393507) and [here](https://github.com/bazelbuild/rules_docker/pull/1248#issuecomment-548506405)) as a preferable solution. It could be done once to generate a base python image to prevent all further images from adding extra `container_image` layer. After current PR gets merged, it could be done in more straightforward way, declaring base image with entrypoint.

Your suggestion to add `container_options ` to be passed to underlying `container_image` seems good, but we should consider more use cases, imho

PR #2250, opened 2023-04-21
Support runfiles root_symlinks in images

wolfd:
Fixes symlink support and adds root_symlink support.

## PR Checklist

^ I'm not aware of docs that mention what specifically happens to runfiles.

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

Previously I was seeing that `runfiles.symlinks` (regardless of whether the target file existed in `runfiles.file`), did not point to the correct location in the image (they were broken).

Additionally, `root_symlinks` were entirely ignored

Issue Number: N/A

## What is the new behavior?

The overall method we take here is to create real files where the [root_]symlink is declared if the target `File` is not seen in `file_map`.

We now use `_final_file_path` instead of `layer_file_path`. This seems to work, but I am not sure if I'm missing a use-case here.

This PR also adds some extra test functionality to check that symlinks point at the expected location.

- [ ] Yes
- [x] No

I don't think that the previous behavior is considered working, and I'm having trouble imagining a case where that broken behavior would be desired.


lang/image.bzl
```diff
@@ -186,11 +192,24 @@ def _app_layer_impl(ctx, runfiles = None, emptyfiles = None):
     # app layer, we can already create symlinks to the runfiles path.
     if ctx.attr.binary:
         # Include any symlinks from the runfiles of the target for which we are synthesizing the layer.
-        symlinks.update({
-            (_reference_dir(ctx) + "/" + s.path): layer_file_path(ctx, s.target_file)
-            for s in _default_symlinks(dep).to_list()
-            if hasattr(s, "path")  # "path" and "target_file" are exposed to starlark since bazel 0.21.0.
```
wolfd, line 192: I figure that 0.21.0 is not worth worrying about anymore?

PR #2285, opened 2025-04-17
.bazelversion: upgrade to 8.2.0

sluongng:
Also provide some simple cleanup to make sure things are compatible


sluongng:
Just FYI: i don't have any plan to fix the existing WORKSPACE CI setup.

The changes should be small and obvious enough that they are safe to merge without CI.
I think we can resume CI once bzlmod support landed.

linzhp (approved):

